Issue #13, opened 2018-11-08
Support named path extraction


aturon:
The router supports both named (`{foo}`) and anonymous (`{}`) wildcards, but Tide only has an extractor (`Path`) for the latter.

However, extractors are given `RouteMatch` structures which contain named match information.

All that's needed is to add an additional extractor type for extracting path components by name. Here's a sketch:

```rust
trait NamedComponent: Send + 'static + std::str::FromStr {
    const NAME: &'static str;
}

struct Named<T: NamedComponent>(pub T);

impl<T: NamedComponent, S: 'static> Extract<S> for Named<T> { ... }
```

mythmon:
With #32 merged, should this issue be closed, or is there more work to do?

tirr-c:
I think this is good to close.

PR #32, opened 2018-11-09
Named path extractors

Stinners:
Added an extractor type based on the sketch in #13 as well as an example of implementing the traits to make it work


aturon:
Beautiful, thank you! :100:

PR #33, opened 2018-11-09
Adding String and Vec Body Types

### LINKED ISSUE #7: Add more body types
Currently the only `Body` extractor/constructor we offer is [`Json`](https://github.com/rust-net-web/tide/blob/master/src/body.rs#L115-L149).

We should at the very least offer types for working with the body as a raw `Vec<u8>` or `String`, but perhaps add some additional serde-based types as well.

tzilist:
Apologies in advance - I am extremely new to rust. 

Hopefully this PR closes #7 and also adds some examples on how to extract different body types

Let me know if you'd like any changes!


examples/body_types.rs
```diff
@@ -0,0 +1,36 @@
...
+use tide::body;
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+struct Message {
+  author: Option<String>,
+  contents: String,
+}
+
+async fn echo_string(msg: String) -> String {
+  println!("String: {}", msg);
```
pickfire: I think we shouldn't do `println!()` in examples, rather I think it would be nice to show an example with logging.

src/body.rs
```diff
@@ -146,4 +146,43 @@ impl<T: 'static + Send + serde::Serialize> IntoResponse for Json<T> {
             .body(Body::from(serde_json::to_vec(&self.0).unwrap()))
             .unwrap()
     }
-}
\ No newline at end of file
+}
+
+impl <S: 'static> Extract<S> for String {
```
pickfire: @tzilist I think it would be nice to `rustfmt` the file here, it uses `impl<S: 'static> ...` rather than `impl <S: 'static>`.

examples/body_types.rs
tzilist: Agreed! Logging middleware hasn't been implemented yet. Should I implement one in the example? Or should we wait until that is done to do that here?

pickfire: Yeah, I just saw that on https://github.com/rust-net-web/tide/issues/8. I would also like to try out experimenting with logging if I can. Maybe can just leave this be for now.

src/body.rs
```diff
@@ -146,4 +146,43 @@ impl<T: 'static + Send + serde::Serialize> IntoResponse for Json<T> {
...
+
+  fn extract(
+    data: &mut S,
+    req: &mut Request,
+    params: &RouteMatch<'_>,
+  ) -> Self::Fut {
+    let mut body = std::mem::replace(req.body_mut(), Body::empty());
+
+    FutureObj::new(Box::new(async move {
+      fn mk_err<T>(_: T) -> Response { StatusCode::BAD_REQUEST.into_response() }
```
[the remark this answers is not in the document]
tzilist: Was wondering that myself. Curious if we should wait for the config to be solidified (where we can state the default failure response) before worrying too much about this?

src/body.rs
tzilist: done! :)

aturon:
Thanks for this PR! 

One suggestion: I think that tying these extractors directly to the `String` and `Vec` types could be a bit confusing, because for the endpoint the argument just looks like... a `String`.

I think it'd be better to do something similar to `Json<T>`:

```rust
pub struct BodyString(pub String);
pub struct BodyBytes(pub Vec<u8>);
```

(feel free to improve the names)

tzilist:
@aturon I have taken your suggestion and added several types `Str`, `StrLossy` and `Bytes`. I am hoping that that is clear when setting the types for the actual endpoints, i.e.
```rust
fn endpoint(req_body: body::Bytes) 
```

or 

```rust
fn endpoint(req_body: body::StrLossy)
```

Let me know what you think!

aturon:
@tzilist Looks great! Once the rustfmt stuff is fixed, should be good to go!

tzilist:
@aturon Ok should be good to go now, thanks for your patience :)

aturon:
thanks @tzilist, but it looks like there's still a fair amount of reformatting to 2-space indents.

tzilist:
Updated this PR to reflect upstream changes.

tzilist:
And formatting check now works :)

aturon:
Thank you, merged!

PR #35, opened 2018-11-09
Strip whitespace

cramertj:


aturon:
will merge after rebase

cramertj:
@aturon I think this should be good to go post-revert of the other change.

PR #39, opened 2018-11-10
Add simple GraphQL example

tirr-c:
This PR adds a simple yet functional GraphQL server example powered by Juniper. The server has an accumulator that can add values.

This is somewhat boring, but I think this is a good place to start.


aturon:
Very cool! I'm OK to merge as-is, but would be even better with a few comments for folks who might not be familiar with Juniper.

tirr-c:
@aturon Done, please check!

aturon:
Excellent! Merging now!

PR #42, opened 2018-11-12
Feature: Default Headers Middleware

tzilist:
PR is to add the ability to create default headers for every request. This is more or less taken from actix-web!

I am also trying to add the method `headers` which lets you specify a Vec of headers, e.g.

```rust

app.middleware(
  DefaultHeaders::new()
    .headers(vec!(
      ["X-Version", "1.0.0"],
      ["X-Server", "Tide"]
     ))
);
```

As opposed to:

```rust

app.middleware(
  DefaultHeaders::new()
    .header("X-Version", "1.0.0")
    .header("X-Server", "Tide")
);
```
Not sure which one is more clear? Currently the latter is what is implemented. If it's decided that the other way is better, I can work on implementing it.

### DIFF
--- a/src/middleware/default_headers.rs
+++ b/src/middleware/default_headers.rs
@@ -54,10 +54,7 @@ impl<Data> Middleware<Data> for DefaultHeaders {
     ) -> FutureObj<'static, Response> {
         let headers = resp.headers_mut();
         for (key, value) in self.headers.iter() {
-            // if !resp.headers().contains_key(key) {
-            //     resp.headers_mut().insert(key, value.clone());
-            // }
-            headers.entry(key).unwrap().or_insert(value.clone());
+            headers.entry(key).unwrap().or_insert_with(|| value.clone());
         }
         FutureObj::new(Box::new(async { resp }))
     }
--- a/src/middleware/default_headers.rs
+++ b/src/middleware/default_headers.rs
@@ -28,6 +28,7 @@ impl DefaultHeaders {
     }
 
     #[inline]
+    #[cfg_attr(feature = "cargo-clippy", allow(clippy::match_wild_err_arm))]
     pub fn header<K, V>(mut self, key: K, value: V) -> Self
     where
         K: IntoHeaderName,
@@ -45,7 +46,7 @@ impl DefaultHeaders {
 
     // #[inline]
     // pub fn headers<K, V>(self, key_values_pairs: Vec<KeyValues>) -> Self {
-    //     for [key, value] in key_values_pairs.iter() {
+    //     for [key, value] in key_values_pairs {
     //         self.header(key, value);
     //     }
     //     self
--- a/src/middleware/default_headers.rs
+++ b/src/middleware/default_headers.rs
@@ -12,8 +12,6 @@ pub struct DefaultHeaders {
     headers: HeaderMap,
 }
 
-// type KeyValues = [&'static str; 2];
-
 impl DefaultHeaders {
     pub fn new() -> DefaultHeaders {
         DefaultHeaders::default()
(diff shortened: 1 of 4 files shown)


src/middleware/default_headers.rs
```diff
@@ -0,0 +1,70 @@
...
+use crate::{head::Head, Middleware, Response};
+
+#[derive(Clone)]
+pub struct DefaultHeaders {
+    headers: HeaderMap,
+}
+
+// type KeyValues = [&'static str; 2];
+
+impl Default for DefaultHeaders {
```
csmoe: `Default` can be derived for `DefaultHeaders` since `Default` is already impl'd for `HeaderMap` [here](https://docs.rs/http/0.1.13/src/http/header/map.rs.html#1842-1846).

src/middleware/default_headers.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub fn header<K, V>(mut self, key: K, value: V) -> Self
+    where
+        K: IntoHeaderName,
+        HeaderValue: HttpTryFrom<V>,
+    {
+        match HeaderValue::try_from(value) {
+            Ok(value) => {
+                self.headers.append(key, value);
+            }
+            Err(_) => panic!("Cannot create header key value pair"),
```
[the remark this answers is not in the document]
tzilist: I'm usually with ya here, I think this should be caught while users are developing though? I think crashing would be the right move here but I could be wrong :P Does that make sense to you?

src/middleware/default_headers.rs
tzilist: Cool! Didn't know you could do that - I'll change that.

src/middleware/default_headers.rs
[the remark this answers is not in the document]
tirr-c: Okay, that makes sense; my concern was, what if this is fed with untrusted user input? But this middleware is for default header values (which is likely static or from config files), so panicking on failure might be ok.

src/middleware/default_headers.rs
```diff
@@ -0,0 +1,70 @@
...
+
+impl<Data> Middleware<Data> for DefaultHeaders {
+    fn response(
+        &self,
+        data: &mut Data,
+        head: &Head,
+        mut resp: Response,
+    ) -> FutureObj<'static, Response> {
+        for (key, value) in self.headers.iter() {
+            if !resp.headers().contains_key(key) {
```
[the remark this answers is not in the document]
tirr-c: I intended `.or_insert_with(|| value.clone())` because cloning is considered expensive. But clippy doesn't complain about this, so it's good to go!

src/middleware/default_headers.rs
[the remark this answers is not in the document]
tzilist: Yeah I would imagine this would be from a config. It's possible it might come from an untrusted source but I think even then, failing and getting an alert would be preferable to silently failing and not knowing until later, but that could just be me :)

src/middleware/default_headers.rs
```diff
@@ -0,0 +1,53 @@
...
+    where
+        K: IntoHeaderName,
+        HeaderValue: HttpTryFrom<V>,
+    {
+        match HeaderValue::try_from(value) {
+            Ok(value) => {
+                self.headers.append(key, value);
+            }
+            Err(_) => {
+                panic!("Cannot create default header. Please check your default header values.")
```
tirr-c: How about using `.expect()` on `try_from`? We don't need clippy attr then :)

tzilist: When I try to do that, I get this error:
```
the method `expect` exists but the following trait bounds were not satisfied:
      `<http::header::value::HeaderValue as http::convert::HttpTryFrom<V>>::Error : std::fmt::Debug`
```

any thoughts on how to fix that? Should I use `map_err`?

tzilist: Ok, I used `map_err`, not sure if that is the correct way to handle this or not :P

src/middleware/default_headers.rs
```diff
@@ -18,20 +18,16 @@ impl DefaultHeaders {
...
-        match HeaderValue::try_from(value) {
-            Ok(value) => {
-                self.headers.append(key, value);
-            }
-            Err(_) => {
-                panic!("Cannot create default header. Please check your default header values.")
-            }
-        }
+        let value = HeaderValue::try_from(value)
+            .map_err(|_| ())
```
tirr-c: The error type is not `Debug` but `Into<http::Error>`, so let's convert it!

tzilist: to do this should i just put `map_err(Into::into)`? Sorry, not entirely sure how to do this :(

tzilist: Ok that seemed to have worked actually.

Issue #4, opened 2018-11-08
Support "mounting" subrouters and route-sensitive middleware


aturon:
Right now, all routes and middleware must be set up at the top-level `App`. However, it's often useful to be able to nest an existing router (or potentially `App` at a given route inside another `App`. Similarly, it should be possible to set up middleware at specific nestings within a router, so that it only applies to paths with a specific prefix.

To sketch, this might look something like:

```rust
let mut app = App::new(my_data);
app.at("/foo").nest(|router| {
    router.middleware(some_middlware); // only applies to routes within this nesting
    router.at("bar").get(bar_endpoint); // ultimately routes `/foo/bar` to `bar_endpoint`
})
```

Let's nail down the API design and then work out a good implementation approach.

tirr-c:
I came up with an idea, will try when I have enough time (maybe this weekend.)

Basic idea is,
* Make `Router::at` public
* Move `middleware` into `Router`
* `App::router(&mut self)` will return top level router
* `Router::nest<F: FnOnce(&mut Router)>(&mut self, builder: F)` will build a new `Router` and mount it into `self`

J-F-Liu:
I think the middleware design of [KOA](https://koajs.com/) framework is quite good, can we do something similar in Rust?
```javascript
app.use(async (ctx, next) => {
  const start = Date.now();
  await next();
  const ms = Date.now() - start;
  ctx.set('X-Response-Time', `${ms}ms`);
});
```

tirr-c:
@J-F-Liu Seems like #54 is the thing.

PR #43, opened 2018-11-15
(WIP) Nested routers

tirr-c:
An implementation of #4.

Module `url_table` has `Router` trait that performs routing and resource setup, and `GenericRouter` which is a trivial implementation of `Router`.

`router::Router` is an implementation of `url_table::Router`. It is like `GenericRouter` but it also has middlewares.

- [ ] Write documents
- [ ] Add more examples

### DIFF
--- a/examples/simple_nested_router.rs
+++ b/examples/simple_nested_router.rs
@@ -1,0 +1,37 @@
+// A `Router`-nesting version of the example `named_path`.
+
+#![feature(async_await, futures_api)]
+
+use tide::{
+    head::{Named, NamedComponent},
+    Router,
+};
+
+struct Number(i32);
+
+impl NamedComponent for Number {
+    const NAME: &'static str = "num";
+}
+
+impl std::str::FromStr for Number {
+    type Err = std::num::ParseIntError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        s.parse().map(|num| Number(num))
+    }
+}
+
+async fn add_two(Named(number): Named<Number>) -> String {
+    let Number(num) = number;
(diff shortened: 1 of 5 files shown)


aturon:
Thanks for this PR! This is a really clever approach, and definitely seems plausible. However, it's a bit of a shame to have to entangle the router and the URL table aspects so much -- though you've done a good job of at least separating them via a trait.

Thinking through this PR led me to another idea. I'm not sure if this will work. But:

- I think we should aim for a "final" representation of routing where there's no notion of nesting. That is, the result of nesting should be a "flattened" table.
  - One place this matters: in the current PR, the middleware along a path is *dynamically* inserted into a vector as you proceed. But if we flatten after nesting, we can *statically* set up those vectors, so they're ready to go when we actually route.

- Concretely, we could:
  - Store an `Arc<Vec<Box<dyn Middleware>>>` *per endpoint*, as well as per `Router`
  - Have a `Router` have an optional, private "parent" which is another `Router` and a location
  - When you do `nest`, a fresh `Router` is created *on the stack*, with the current router as a parent. You then get `&mut` access to this fresh router.
  - After the call to `nest` returns, we tell that fresh router to inline itself into the parent. That means it injects the contents of its URL table, adjusting the paths by adding the parent prefix, and adjusting the middleware vectors for its endpoints by pre-pending the parent router's middleware.

Does that make sense to you?

aturon:
Follow-up: we don't actually need to *store* the parent information in a router. But rather, we'd have something like:

```rust
impl Router {
    fn move_into_parent(self, parent: &mut Router, at: &str) { ... }
```

tirr-c:
I'll close this PR, and open a new one.

Issue #22, opened 2018-11-08
Add form extractors


aturon:
Definitely take a look at [Rocket](https://rocket.rs/v0.4/guide/requests/#forms), [Actix](https://actix.rs/docs/extractors/), and other frameworks for opportunities for reuse!

PR #44, opened 2018-11-15
Add form data extraction from body via serde_qs

jnicklas:
An attempt at resolving #22, this seemed almost too easy, so maybe I'm misunderstanding something here 😅

This makes it possible extract `x-www-form-urlencoded` data from the request body via [serde_qs](https://crates.io/crates/serde_qs). 

This means that even nested fields can be extracted via the same `foo[bar]` syntax used in node's qs and Ruby on Rails. This is somewhat of an opinionated decision, but the web ecosystem seems to be converging on this and it is a huge boon when submitting complex form data.

The alternative is only allowing a flat map of key->values but this is very limiting.

Another decision here is the naming of the extractor. Since there are two different form encodings, `application/x-www-form-urlencoded` and `multipart/form-data`<sup>1</sup> it might seem reasonable to name this extractor `body::UrlencodedForm` or something, but I think that choosing the simpler `body::Form` name is the better choice here. The reason being that urlencoded forms are far more common than multipart, and urlencoded is the standard when no encoding is specified. In other words, multipart forms really are a special case, and having both `Form` and `MultipartForm` would make the distinction clear. actix-web makes the same choice here, opting for `Form` and `Multipart` as names.

<sup>1</sup>well technically there is also `text/plain`, but it seems to be mostly useless

### DIFF
--- a/src/body.rs
+++ b/src/body.rs
@@ -150,6 +150,42 @@ impl<T: 'static + Send + serde::Serialize> IntoResponse for Json<T> {
     }
 }
 
+/// A wrapper for form encoded (application/x-www-form-urlencoded) (de)serialization of bodies.
+///
+/// This type is usable both as an extractor (argument to an endpoint) and as a response
+/// (return value from an endpoint), though returning a response with form data is uncommon
+/// and probably not good practice.
+pub struct Form<T>(pub T);
+
+impl<T: Send + serde::de::DeserializeOwned + 'static, S: 'static> Extract<S> for Form<T> {
+    // Note: cannot use `existential type` here due to ICE
+    type Fut = FutureObj<'static, Result<Self, Response>>;
+
+    fn extract(data: &mut S, req: &mut Request, params: &RouteMatch<'_>) -> Self::Fut {
+        let mut body = std::mem::replace(req.body_mut(), Body::empty());
+        FutureObj::new(Box::new(
+            async move {
+                let body = await!(body.read_to_vec()).map_err(mk_err)?;
+                let data: T = serde_qs::from_bytes(&body).map_err(mk_err)?;
+                Ok(Form(data))
+            },
+        ))
+    }
+}
+
+impl<T: 'static + Send + serde::Serialize> IntoResponse for Form<T> {
+    fn into_response(self) -> Response {
+        // TODO: think about how to handle errors
+        http::Response::builder()
+            .status(http::status::StatusCode::OK)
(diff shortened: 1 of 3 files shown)


src/body.rs
```diff
@@ -150,6 +150,42 @@ impl<T: 'static + Send + serde::Serialize> IntoResponse for Json<T> {
...
+        ))
+    }
+}
+
+impl<T: 'static + Send + serde::Serialize> IntoResponse for Form<T> {
+    fn into_response(self) -> Response {
+        // TODO: think about how to handle errors
+        http::Response::builder()
+            .status(http::status::StatusCode::OK)
+            .header("Content-Type", "Application/x-www-form-urlencoded")
```
tirr-c: It would be nice if the mime type is written in lowercase.

```suggestion
            .header("Content-Type", "application/x-www-form-urlencoded")
```

jnicklas: Agree on this, it's [written this way](https://github.com/rust-net-web/tide/blob/73dfd4dc260eafcde5252eab736470302bf1a11c/src/body.rs#L147) for `json`, I didn't know why and figured there might be a reason for it, so went with the same capitalization.

jnicklas:
Yes, urlencoded is more common, it was a typo.

src/body.rs
```diff
@@ -150,6 +150,42 @@ impl<T: 'static + Send + serde::Serialize> IntoResponse for Json<T> {
...
+            async move {
+                let body = await!(body.read_to_vec()).map_err(mk_err)?;
+                let data: T = serde_qs::from_bytes(&body).map_err(mk_err)?;
+                Ok(Form(data))
+            },
+        ))
+    }
+}
+
+impl<T: 'static + Send + serde::Serialize> IntoResponse for Form<T> {
```
tzilist: I know there is another one in this file, but would it make sense to move this to the [Response.rs](https://github.com/rust-net-web/tide/blob/master/src/response.rs) file?

aturon: I waffled on that and ended up keeping all code that was purely body-related here.

src/body.rs
aturon: I think that was in turn copied from Rocket :joy:

Lower-casing is good, feel free to fix json  while you're at it.

aturon:
This LGTM modulo a couple tiny nits!

src/body.rs
tzilist: Fair enough!

tirr-c:
Sorry, I kinda forgot this. Merging now!

PR #47, opened 2018-11-18
Allow handle file upload via multipart form

liufuyang:
Hi there, not at all an expert here but I would like to try help a bit if I can.

Do you think either one of the two methods I provided here could be a good way of handling multipart form file-upload? 

(I used some idea from here: https://github.com/rust-net-web/tide/pull/44)

The differences between the two methods I mentioned here are: (see the example first)
* `form_auto` use a new body type created in `body::MultipartFile` to allow user of tide easier get a `Vec<u8>` of the file uploaded.
* `form_manual` use a new head type `head::NamedHeader` to allow user get header "content-type" and raw body as `Vec<u8>` then manually use the `multipart` crate method to get what he want. 

Not sure whether I am heading the correct direction here but I would like to hear your opinion. I am kind of new to Rust, BTW 😆

### DIFF
--- a/examples/multipart-form/main.rs
+++ b/examples/multipart-form/main.rs
@@ -1,0 +1,74 @@
+#![feature(async_await, futures_api)]
+
+#[macro_use]
+extern crate serde_derive;
+
+use http::status::StatusCode;
+use std::io::Read;
+use tide::{body, App};
+
+#[derive(Serialize, Deserialize, Clone)]
+struct Message {
+    key1: Option<String>,
+    key2: Option<String>,
+    file: Option<String>,
+}
+
+async fn upload_file(
+    multipart_form: body::MultipartForm,
+) -> Result<body::Json<Message>, StatusCode> {
+    // https://stackoverflow.com/questions/43424982/how-to-parse-multipart-forms-using-abonander-multipart-with-rocket
+    let mut message = Message {
+        key1: None,
+        key2: None,
+        file: None,
+    };
+
+    let mut multipart = multipart_form.0;
+    multipart
+        .foreach_entry(|mut entry| match entry.headers.name.as_str() {
+            "file" => {
+                let mut vec = Vec::new();
+                entry.data.read_to_end(&mut vec).expect("can't read");
+                message.file = String::from_utf8(vec).ok();
+                println!("key file got");
+            }
+
+            "key1" => {
+                let mut vec = Vec::new();
+                entry.data.read_to_end(&mut vec).expect("can't read");
+                message.key1 = String::from_utf8(vec).ok();
+                println!("key1 got");
+            }
+
+            "key2" => {
+                let mut vec = Vec::new();
+                entry.data.read_to_end(&mut vec).expect("can't read");
+                message.key2 = String::from_utf8(vec).ok();
+                println!("key2 got");
+            }
+
+            _ => {
+                // as multipart has a bug https://github.com/abonander/multipart/issues/114
+                // we manually do read_to_end here
+                let mut _vec = Vec::new();
+                entry.data.read_to_end(&mut _vec).expect("can't read");
+                println!("key neglected");
+            }
+        })
+        .expect("Unable to iterate multipart?");
+
+    Ok(body::Json(message))
+}
+
+fn main() {
+    let mut app = App::new(());
+
+    app.at("/upload_file").post(upload_file);
+
+    app.serve("127.0.0.1:7878");
+}
+
+// Test with:
+// curl -X POST http://localhost:7878/upload_file -H 'content-type: multipart/form-data' -F file=@examples/multipart-form/test.txt
+// curl -X POST http://localhost:7878/upload_file -H 'content-type: multipart/form-data' -F key1=v1, -F key2=v2
--- a/src/body.rs
+++ b/src/body.rs
@@ -5,7 +5,9 @@
 
 use futures::{compat::Compat01As03, future::FutureObj, prelude::*, stream::StreamObj};
 use http::status::StatusCode;
+use multipart::server::Multipart;
 use pin_utils::pin_mut;
+use std::io::Cursor;
 
 use crate::{Extract, IntoResponse, Request, Response, RouteMatch};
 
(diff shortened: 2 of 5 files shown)


src/body.rs
```diff
@@ -117,6 +120,48 @@ fn mk_err<T>(_: T) -> Response {
...
+impl<S: 'static> Extract<S> for MultipartFile {
+    // Note: cannot use `existential type` here due to ICE
+    type Fut = FutureObj<'static, Result<Self, Response>>;
+
+    fn extract(data: &mut S, req: &mut Request, params: &RouteMatch<'_>) -> Self::Fut {
+        // https://stackoverflow.com/questions/43424982/how-to-parse-multipart-forms-using-abonander-multipart-with-rocket
+        let ct = req
+            .headers()
+            .get("content-type")
+            .expect("no header content-type");
```
aturon: Note that `expect` will panic on  failure, which is probably not  what we want here -- I suspect we want to return an error response instead?

src/body.rs
```diff
@@ -117,6 +120,48 @@ fn mk_err<T>(_: T) -> Response {
...
+                let mut file = None;
+                mp.foreach_entry(|mut entry| match entry.headers.name.as_str() {
+                    "file" => {
+                        let mut vec = Vec::new();
+                        entry.data.read_to_end(&mut vec).expect("can't read");
+                        file = Some(vec);
+                    }
+                    other => panic!("No known key {}", other),
+                })
+                .expect("Unable to iterate");
```
aturon: Same issues with `expect` here

src/body.rs
```diff
@@ -117,6 +120,48 @@ fn mk_err<T>(_: T) -> Response {
...
+
+        let mut body = std::mem::replace(req.body_mut(), Body::empty());
+
+        FutureObj::new(Box::new(
+            async move {
+                let body = await!(body.read_to_vec()).map_err(mk_err)?;
+
+                let mut mp = Multipart::with_body(Cursor::new(body), boundary);
+                let mut file = None;
+                mp.foreach_entry(|mut entry| match entry.headers.name.as_str() {
```
aturon: I'm a little confused about the intent here -- this looks like it's tied to only a very specific header for multipart uploads?  Probably we'll want to be more general.

tzilist: It might be worth just returning the multipart instead of `Vec<u8>` here and allow the users to loop through the response to find the file they are looking for.

examples/multipart-form/main.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    app.at("/form_auto").post(form_auto);
+    app.at("/form_manual").post(form_manual);
+
+    app.serve("127.0.0.1:7878");
+}
+
+// Test with:
+// curl -X POST   http://localhost:7878/form_manual -H 'Cache-Control: no-cache'   -H 'Postman-Token: e26f7bce-51b4-4d12-aba8-439ff8549bb9'   -H 'content-type: multipart/form-data'   -F file=@examples/multipart-form/test.txt
+// curl -X POST   http://localhost:7878/form_auto   -H 'Cache-Control: no-cache'   -H 'Postman-Token: e26f7bce-51b4-4d12-aba8-439ff8549bb9'   -H 'content-type: multipart/form-data'   -F file=@examples/multipart-form/test.txt
```
tzilist: Great idea adding these curl requests! One nit - I think we can remove the `postman-token` header :)

src/body.rs
liufuyang: Indeed, this is a highly WIP PR :) I would like to know whether I am pointing at the correct direction then later try fix all the details that can be fixed 😆

src/body.rs
liufuyang: Yep, that is what I thought as well. I actually tired to just returning the multipart yesterday but I couldn't win the fight with the compiler. Perhaps I can try again and it would be nice if some of you could help me out :)

liufuyang:
Thank you for the good input. I will try with just return the Multipart type back and "allow the users to loop through the response to find the file they are looking for." Will get back to you soon :)

src/body.rs
```diff
@@ -117,6 +119,42 @@ fn mk_err<T>(_: T) -> Response {
...
+            Some(h) => h,
+            None => {
+                // return FutureObj::new(Box::new(Err(
+                //     http::status::StatusCode::BAD_REQUEST.into_response()
+                // )));
+                panic!("Headers have no key as content-type")
+                // TODO: How to return early here? The above return is not working, saying:
+                //   134 |                 return FutureObj::new(Box::new(Err(
+                //       |                        ^^^^^^^^^^^^^^ the trait `core::future::future::Future` is not implemented for `std::result::Result<_, http::response::Response<body::Body>>
+            }
```
liufuyang: Question 1: How to return early here when the header has no `content-type` key?
I tried to do this but not working, error see in the comments in code above.
```rust
return FutureObj::new(Box::new(Err(
                http::status::StatusCode::BAD_REQUEST.into_response()
                )));
``` 🤔

File: src/body.rs (hunk 1 above)
Comment (liufuyang) [reply src/body.rs]: hmmm... I am not sure how to deal with this option and return early here... also see my new commit and comments below. (Still working on it...)

Comment (liufuyang) [reply src/body.rs]: pushed one more commit 995bba8, but not sure if my way of early returning FutureObj with async is the best way to deal with it though 🤔

File: src/body.rs (hunk 5 above)
Comment (liufuyang) [reply src/body.rs]: Looks like I need to use a `async { ... }`. But not sure if the current way of doing is good enough. Please give some feedback :)

Comment (aturon) [reply src/body.rs]: So, one thing you can do instead is the following:

```rust
const BOUNDARY = "boundary=";
let boundary = req.headers().get("content-type")
    .map(|ct| {
        let ct = ct.to_str().ok()?;
        let idx = ct.find(BOUNDARY)?;
        Some(ct[idx + "BOUNDARY".len()..].to_string())
    });
```

Then, inside the bottom `async` block, you can say:

```rust
let boundary = boundary.ok_or_else(mk_err)?;
```

liufuyang: Thank you very much, this looks much better and never thought one could do this. Learned a lot :) However I need to do some minor changes on it, as the `map` call returns me an option of option, and the `ok_or_else` takes in a function that takes no arguments. 

```rust
const BOUNDARY: &str = "boundary=";
        let boundary = req.headers().get("content-type").map_or(None, |ct| {
            let ct = ct.to_str().ok()?;
            let idx = ct.find(BOUNDARY)?;
            Some(ct[idx + BOUNDARY.len()..].to_string())
        });
...

let boundary = boundary.ok_or_else(|| {
                    StatusCode::BAD_REQUEST.into_response()
                })?;
```

Pushed another commit, perhaps you can take a look again?

examples/multipart-form/main.rs
```diff
@@ -0,0 +1,73 @@
...
+
+            "key2" => {
+                let mut vec = Vec::new();
+                entry.data.read_to_end(&mut vec).expect("can't read");
+                message.key2 = String::from_utf8(vec).ok();
+                println!("key2 got");
+            }
+
+            _ => {
+                println!("key neglected");
```
tirr-c: Reading `entry.data` to end might solve the problem.

examples/multipart-form/main.rs
```diff
@@ -0,0 +1,73 @@
...
+            }
+        })
+        .expect("Unable to iterate multipart?"); // TODO: why an unspecified key will break the foreach_entry() call??? See test curl example below.
```
tirr-c: `foreach_entry` will always try to read multipart boundary. If we don't read the data, the cursor will remain same, so it fails to read boundary. Seems like this is a bug of `multipart` crate.

liufuyang: 🤔 Perhaps I can go ask around on there repo?

liufuyang: It turns out to be a bug. https://github.com/abonander/multipart/issues/114

Reading entry.data again in our example to make the example work now.

aturon:
Thanks for the revisions, looks great!

Issue #45, opened 2018-11-17
Strip the result body when creating default HEAD responses


aturon:
If you don't specify a `HEAD` behavior but you do provide `GET`, Tide will automatically use the `GET` behavior. But it retains the body, which should be stripped.

tzilist:
@aturon did a little diggin into this, not sure if I am reading this right either, but it seems like hyper handles this for us. In the current code on master here are some example responses from curl
I'm running the `named_path` example:
```bash
url --get  --url http://127.0.0.1:8000/add_two/2   --header 'content-type: application/json' -v
```
Returns
```
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 8000 (#0)
> GET /add_two/2 HTTP/1.1
> Host: 127.0.0.1:8000
> User-Agent: Mozilla/5.0 (compatible; MSIE 9.0; Windows NT 6.1; Trident/5.0)
> Accept: */*
> Referer:
> content-type: application/json
>
< HTTP/1.1 200 OK
< content-type: text/plain
< content-length: 15
< date: Mon, 19 Nov 2018 04:53:55 GMT
<
* Connection #0 to host 127.0.0.1 left intact
2 plus two is 4
```

and the head request:

```bash
curl --head  --url http://127.0.0.1:8000/add_two/2   --header 'content-type: application/json' -v
```
returns
```
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 8000 (#0)
> HEAD /add_two/2 HTTP/1.1
> Host: 127.0.0.1:8000
> User-Agent: Mozilla/5.0 (compatible; MSIE 9.0; Windows NT 6.1; Trident/5.0)
> Accept: */*
> Referer:
> content-type: application/json
>
< HTTP/1.1 200 OK
HTTP/1.1 200 OK
< content-type: text/plain
content-type: text/plain
< content-length: 15
content-length: 15
< date: Mon, 19 Nov 2018 04:53:33 GMT
date: Mon, 19 Nov 2018 04:53:33 GMT

<
* Connection #0 to host 127.0.0.1 left intact
```

Perhaps I am missing something though. Thoughts?

DeltaManiac:
@tzilist 

I had observed the same behavior while testing #31. 
Guess i missed something too :)

aturon:
[...]

PR #49, opened 2018-11-19
strip body on head request

tzilist:
Hopefully fixes issue #45. Not sure exactly where this logic belongs but seems to make sense to me. :)

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -118,6 +118,10 @@ impl<Data: Clone + Send + Sync + 'static> Service for Server<Data> {
                         resp = await!(m.response(&mut data, &head, resp))
                     }
 
+                    if method == http::Method::HEAD {
+                        resp = resp.map(|_| Body::empty())
+                    }
+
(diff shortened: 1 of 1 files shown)


tirr-c:
The response header of `HEAD` should be identical to those of `GET`, but it is tricky because of `Content-Length` header. It is handled by `Payload` trait from Hyper, which is implemented for `hyper::Body`.

If you give empty body to the response, `Content-Length` will be 0, which leads to the mismatch of headers. We need to implement `Payload` manually, but it requires `Poll` from futures 0.1 (from function signature).

tzilist:
Oh wow, that is a great point! I'll dig into this a bit more then. :)

tzilist:
Holding off on this until questions in #45 are resolved

aturon:
Given how #45 has gone, I'm gonna close this out. It'd be great to get a PR documenting Hyper's behavior and how Tide relies on it, though!

PR #52, opened 2018-11-19
Logging middleware

bIgBV:
This PR is to establish a base to build the logging infrastructure upon. The discussion is still ongoing in #8 , but this PR introduces a basic logging instance to start displaying information regarding the requests being served by the application. The output of running an application now shows up in the terminal as follows:

screen shot 2018-11-19 at 17 16 45

As you can see, the method, path and the status code of a request are logged to the terminal. This provides immediate feedback for the user.

Right now I'm instantiating an instance of the `SimpleLogger` middleware and adding it to the list of middlewares by default.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -14,8 +14,8 @@ use crate::{
     extract::Extract,
     middleware::RequestContext,
     router::{Resource, RouteResult, Router},
+    logger::RootLogger,
     Middleware, Request, Response, RouteMatch,
-    logger::RootLogger
 };
 
 /// The top-level type for setting up a Tide application.
--- a/src/app.rs
+++ b/src/app.rs
@@ -15,6 +15,7 @@ use crate::{
     middleware::RequestContext,
     router::{Resource, RouteResult, Router},
     Middleware, Request, Response, RouteMatch,
+    logger::SimpleLogger
 };
 
 /// The top-level type for setting up a Tide application.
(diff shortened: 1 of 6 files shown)


src/app.rs
```diff
@@ -30,10 +31,11 @@ pub struct App<Data> {
 impl<Data: Clone + Send + Sync + 'static> App<Data> {
     /// Set up a new app with some initial `data`.
     pub fn new(data: Data) -> App<Data> {
+        let logger = RootLogger::new();
         App {
             data,
             router: Router::new(),
-            middleware: Vec::new(),
+            middleware: vec![Box::new(logger)],
```
aturon: Let's hold off on setting up default middleware just yet -- maybe open an issue on it so we can discuss how we want to approach it?

src/lib.rs
```diff
@@ -17,6 +17,7 @@ pub mod body;
 mod endpoint;
 mod extract;
 pub mod head;
+mod logger;
```
aturon: We should consider putting this into a submodule of `middlware` and exposing it publicly that way.

src/logger.rs
```diff
@@ -0,0 +1,48 @@
...
+
+use futures::future::FutureObj;
+
+use crate::{head::Head, Middleware, Response};
+
+/// Root logger for Tide. Wraps over logger provided by slog.SimpleLogger
+///
+/// Only used internally for now.
+pub(crate) struct RootLogger {
+    // drain: dyn slog::Drain,
```
aturon: Should clean up this commented-out line.

aturon: Also, if/when we expose this publicly, we should bikeshed the name a bit. (No strong opinion here)

aturon:
Thank you!

PR #57, opened 2018-11-24
Relieve middleware FutureObj lifetime bounds

tirr-c:
Side effect: `Request` and `Response` will be given by reference.

### DIFF
--- a/src/middleware/mod.rs
+++ b/src/middleware/mod.rs
@@ -15,8 +15,8 @@ pub trait Middleware<Data>: Send + Sync {
         data: &'a mut Data,
         req: &'a mut Request,
         params: &'a RouteMatch<'_>,
-    ) -> FutureObj<'a, Option<Response>> {
-        FutureObj::new(Box::new(async { None }))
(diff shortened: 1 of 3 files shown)


src/middleware/mod.rs
```diff
@@ -10,23 +10,23 @@ pub use self::default_headers::DefaultHeaders;
...
-        req: Request,
-        params: &RouteMatch<'_>,
-    ) -> FutureObj<'static, Result<Request, Response>> {
-        FutureObj::new(Box::new(async { Ok(req) }))
+    fn request<'a>(
+        &'a self,
+        data: &'a mut Data,
+        req: &'a mut Request,
+        params: &'a RouteMatch<'_>,
+    ) -> FutureObj<'a, Option<Response>> {
```
aturon: We may want to use `Result<(), Response>` if we make this change, to integrate better with `?`

tirr-c: That makes sense! I'll change this right now.

aturon:
LGTM! Let's merge after Travis is happy.

Issue #54, opened 2018-11-19
Consider "around" middleware design


aturon:
The current middleware design is based on a "before"/"after" split. We could instead have each middleware invoke the rest of the middleware/endpoint chain.

aturon:
(I have a bunch of thoughts/design constraints here, will write them up ASAP)

tirr-c:
[Koa] also uses "around" middleware approach. Koa goes one step further; everything is middleware, even [the router]! In this way, fallback handler can be implemented as the deepest middleware on the chain. When the router fails to route, it'd invoke the next middleware.

[Koa]: https://github.com/koajs/koa
[the router]: https://github.com/alexmingoia/koa-router

bIgBV:
I know this was merged recently, but can you guys go into a little detail as to the pros of such a middleware design?

aturon:
For sure! Here are some thoughts:

- First, we can easily provide constructors for before/after-style middleware as well. At some point in the PR, @tirr-c had both available. So we don't lose anything, and if we find it's more convenient to write in that style, we can give a way to do it.

- This "around" style middleware makes it easier to pass information from "before" to "after" computations; you don't have to use `extensions` to do so any more, you just use `let`.

- The original middleware style forced an allocation for every middleware on every request. In this new setup, if you're doing *synchronous* work in middleware (which is often the case) you can usually avoid this allocation.

- The middleware interface itself is a bit simpler (IMO)

Because it's possible to provide easy before/after constructors on top of the new interface, I don't see a lot of downsides.

bIgBV:
Awesome, that makes a lot of sense. One thing I'm still not entirely sure of is how data is shared between different middleware and between middleware and the endpoint.

Am I right in thinking that the `Data` in a `RequestContext` will be used for sharing data between middleware and endpoints? I'm still not sure how it will be shared between two middlewares though.

aturon:
This was completed in https://github.com/rust-net-web/tide/pull/58

PR #58, opened 2018-11-25
Experiment: Wrap-around middleware

tirr-c:
Experimental implementation of #54.

* Do we need to keep request-response middleware trait?

### DIFF
--- a/src/endpoint.rs
+++ b/src/endpoint.rs
@@ -8,14 +8,14 @@ use crate::{extract::Extract, head::Head, IntoResponse, Request, Response, Route
 /// implemented directly by Tide users.
 pub trait Endpoint<Data, Kind>: Send + Sync + 'static {
     /// The async result of `call`.
-    type Fut: Future<Output = (Head, Response)> + Send + 'static;
+    type Fut: Future<Output = Response> + Send + 'static;
 
     /// Invoke the endpoint on the given request and app data handle.
     fn call(&self, data: Data, req: Request, params: RouteMatch<'_>) -> Self::Fut;
 }
 
 type BoxedEndpointFn<Data> =
-    dyn Fn(Data, Request, RouteMatch) -> FutureObj<'static, (Head, Response)> + Send + Sync;
+    dyn Fn(Data, Request, RouteMatch) -> FutureObj<'static, Response> + Send + Sync;
 
 pub(crate) struct BoxedEndpoint<Data> {
     endpoint: Box<BoxedEndpointFn<Data>>,
@@ -38,7 +38,7 @@ impl<Data> BoxedEndpoint<Data> {
         data: Data,
         req: Request,
         params: RouteMatch<'_>,
-    ) -> FutureObj<'static, (Head, Response)> {
+    ) -> FutureObj<'static, Response> {
         (self.endpoint)(data, req, params)
     }
 }
@@ -68,7 +68,7 @@ macro_rules! end_point_impl_raw {
                 $X: Extract<Data>
             ),*
         {
-            type Fut = FutureObj<'static, (Head, Response)>;
+            type Fut = FutureObj<'static, Response>;
 
             #[allow(unused_mut, non_snake_case)]
             fn call(&self, mut data: Data, mut req: Request, params: RouteMatch<'_>) -> Self::Fut {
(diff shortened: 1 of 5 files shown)


src/middleware/mod.rs
```diff
@@ -1,13 +1,15 @@
...
-use crate::{head::Head, Request, Response, RouteMatch};
+use crate::{endpoint::BoxedEndpoint, head::Head, Request, Response, RouteMatch};
 
 mod default_headers;
 
 pub use self::default_headers::DefaultHeaders;
 
 /// Middleware for an app with state of type `Data`
-pub trait Middleware<Data>: Send + Sync {
+pub trait ReqResMiddleware<Data>: Send + Sync {
```
aturon: I think we can probably drop this trait for now -- unless you think there's a strong need for it?

aturon: If we do keep it, we could consider providing a blanket impl rather than an adapter:

```rust
impl<M, Data> Middleware<Data> for M where M: ReqResMiddleware<Data> { ... }
```

src/middleware/mod.rs
```diff
@@ -40,3 +49,99 @@ pub trait Middleware<Data>: Send + Sync {
...
+                        res,
+                        head,
+                    }
+                }
+            },
+        ))
+    }
+}
+
+pub struct ResponseContext<Data> {
```
aturon: So, here's an interesting thing: the main reason to have the `Head` type at all (and to return it from the endpoint) was to support "after"-style middleware getting access to most of the request. But with around middleware, we can maybe simplify.

Because of compiler limitations, we aren't currently able to have endpoints get `&Request` but have to pass full ownership. Eventually we'll be able to pass in a borrow.

But with around middleware, if you want to do something with request data after the endpoint runs, you can just compute what you need *before* calling `next`, and then refer to it later.

----

I'm thinking we just drop `head`, `Head`, and `ResponseContext` altogether, and just return a `Response`. This would also let us resolve https://github.com/rust-net-web/tide/issues/41#issuecomment-438496200

What do you think?

src/router.rs
```diff
@@ -198,9 +200,19 @@ mod tests {
...
+    use crate::{
+        body::Body,
+        middleware::{RequestContext, ResponseContext},
+        AppData, Response,
+    };
+
+    fn passthrough_middleware<Data: Clone + Send>(
+        ctx: RequestContext<Data>,
+    ) -> FutureObj<ResponseContext<Data>> {
+        FutureObj::new(Box::new(ctx.next()))
```
aturon: nit: no need to make a new `FutureObj` here -- I think this can literally just be `ctx.next()`.

More generally, one *really cool* think about the middleware setup in this PR is that a lot of middleware won't have to allocate a new `FutureObj` at all!

src/middleware/mod.rs
```diff
@@ -40,3 +49,99 @@ pub trait Middleware<Data>: Send + Sync {
...
+impl<'a, Data> RequestContext<'a, Data> {
+    /// Consume this context, and respond with given `Response`.
+    pub fn respond(self, res: Response) -> ResponseContext<Data> {
+        ResponseContext::from_request_response(self, res)
+    }
+}
+
+impl<'a, Data: Clone + Send> RequestContext<'a, Data> {
+    /// Consume this context, and run remaining middleware chain to completion.
+    pub fn next(mut self) -> FutureObj<'a, ResponseContext<Data>> {
```
aturon: This is brilliant!

src/middleware/mod.rs
tirr-c: I've tried to add a blanked impl, but it made a conflict with impl for `DefaultHeaders`.

src/middleware/mod.rs
tirr-c: I've totally missed that point! We can definitely simplify this.

src/middleware/mod.rs
aturon: Ahha, makes sense! Do you want to just nix this trait for the time being?

src/endpoint.rs
```diff
@@ -79,11 +79,11 @@ macro_rules! end_point_impl_raw {
                     let head = Head::from(parts);
```
tirr-c: Endpoints might want to extract `Head`, so I left this as-is for now.

src/middleware/mod.rs
tirr-c: Done!

PR #59, opened 2018-11-25
Extract `url_table` into `path_table` crate

### LINKED ISSUE #2: Factor out `url_table` into its own crate
The [`url_table` module](https://github.com/rust-net-web/tide/blob/master/src/url_table.rs) implements *generic* routing that is not tied to any other aspects of Tide. It should be pulled out into its own crate, after deciding on a final name.

tirr-c:
Resolves #2.

Renamed `url_table` to `path_table` as "URL" feels like a broader term than "path" (which the crate is actually dealing with).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,8 @@ rust:
 before_script: |
   rustup component add rustfmt-preview clippy-preview
 script: |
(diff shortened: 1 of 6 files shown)


aturon:
Thank you!

PR #72, opened 2018-11-26
Pin a specific nightly version for travis builds

hoodie:
Todays nightly has a [bug that breaks the builds](https://travis-ci.org/hoodie/tide/builds/459748501). but it [seems to build](https://travis-ci.org/hoodie/tide/builds/459753610) with the version from November 24th

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - nightly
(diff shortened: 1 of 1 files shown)


caulagi:
Its being tracked here - https://github.com/rust-lang/rust/issues/56237

aturon:
Thanks!

PR #74, opened 2018-11-26
Expose Computed Values Api

hoodie:
This adds an example for how to use computed values. We had to make the
Compute trait public for that.


aturon:
Awesome, thanks!

Issue #18, opened 2018-11-08
Document routing syntax


aturon:
The `at` method should be documented with the router syntax and fallback semantics.

hseeberger:
I think I'd like to give it a shot, because I'm a Rust and Tide newbie and hence a documentation task seems a good entry point.

I'm not quite sure though what kind of documentation to create – I guess we are talking about documentation comments (`///`), right?

tirr-c:
I see this is resolved now, is it okay to close?

hseeberger:
Thanks. Yes, for me that's fine.

PR #76, opened 2018-11-26
Document routing syntax

hseeberger:
I took a first stab at #18.

Some items for discussion:
- I suggest we consistently use one of the terms "component" or "segment" for the "/"-separated parts of a URL path; currently both are used in `lib.rs` and other places; Akka HTTP uses "segment" which I also prefer
- ~The doc fist test (around line 33) fails: "help: add #![feature(async_await)] to the crate attributes to enable"; but these are enabled, what to do?~
- How can I refer from one `at` method (in "app.rs") to another one (in "router.rs")? Is the way I did it correct?

### DIFF
--- a/src/router.rs
+++ b/src/router.rs
@@ -22,20 +22,49 @@ pub(crate) struct RouteResult<'a, Data> {
 }
 
 impl<Data: Clone + Send + Sync + 'static> Router<Data> {
-    /// Create a new top-level router.
-    pub(crate) fn new() -> Router<Data> {
-        Router {
-            table: PathTable::new(),
-            middleware_base: Vec::new(),
-        }
-    }
-
-    /// Add a new resource at `path`, relative to this router.
+    /// Add a new resource at the given `path`, relative to this router.
+    ///
+    /// Routing means mapping an HTTP request to an endpoint. Here Tide applies a "table of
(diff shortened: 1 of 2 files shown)


src/router.rs
```diff
@@ -22,15 +22,40 @@ pub(crate) struct RouteResult<'a, Data> {
...
+    /// # let mut app = tide::App::new(());
+    /// app.at("/").get(async || "Hello, world!");
+    /// ```
+    ///
+    /// A path is comprised of zero or many segments, i.e. non-empty strings separated by '/'. There
+    /// are two kinds of segments: concrete and wildcard. A concrete segment is used to exactly
+    /// match the respective part of the path of the incoming request. A wildcard segment on the
+    /// other hand extracts and parses the respective part of the path of the incoming request to
+    /// pass it along to the endpoint as an argument. A wildcard segment is either defined by "{}"
+    /// or by "{name}" for a so called named wildcard segment which must have an implementation of
```
aturon:
```suggestion
    /// or by "{name}" for a so called named wildcard segment which can be extracted using
```

aturon:
This looks great, thank you so much!

hseeberger:
Thanks for the quick merge, @aturon. Actually I had some questions I wrote in a comment above. I'll open an new PR for renaming component to segment.

Issue #78, opened 2018-11-27
Use the same port in the examples


ibaryshnikov:
Some examples contain
```
app.serve("127.0.0.1:8000");
```
And others are
```
app.serve("127.0.0.1:7878")
```
Can we have the same port for all of the examples?
Also, is it a good idea to print the url? Like

**Server is listening on http://127.0.0.1:8000**

to be able to navigate to the browser by click

aturon:
Closed by #79

PR #79, opened 2018-11-27
Use the same port in examples

ibaryshnikov:
Solves #78. Updated all of the examples to use the same port, also added the address line to the output.


aturon:
Lovely, thank you!

PR #80, opened 2018-11-27
Extend cookie example

simonasker:
This extends the cookie example to handle multiple cookies.


aturon:
This is great, thanks!

One thought for a follow-up PR: we should consider moving the `Cookies` struct into `request.rs` or `head.rs` and providing it directly with Tide, rather than only having the implementation in this example.

aturon:
Opened [follow-up issue](https://github.com/rust-net-web/tide/issues/84)

PR #81, opened 2018-11-27
Implemented Deref for body types and some Path types, updated examples

ibaryshnikov:
The following body types are covered:
- Str
- StrLossy
- Bytes
- Form
- MuliypartForm
- Json
And some types from head.rs:
- Path
- Named
Are all the cases covered? I'm actually not sure about Named, too.


aturon:
Looks great! Can you also include `DerefMut` implementations in all cases?

aturon:
Perfect!

PR #88, opened 2018-11-27
Rename path component to path segment

hseeberger:
I know that naming is hard, but I nevertheless give it a try: Like asked in #76 I suggest to consistently use either of the terms "component" or "segment" for the "/"-separated parts of the path of a URL. Currently both terms are mixed in lib.rs and other places. I suggest to use "segment" which I think is more specific in the context of a path than "component".

### DIFF
--- a/examples/named_path.rs
+++ b/examples/named_path.rs
@@ -1,10 +1,10 @@
 #![feature(async_await, futures_api)]
 
-use tide::head::{Named, NamedComponent};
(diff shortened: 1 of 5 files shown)


aturon:
Ah, sorry I missed that this wasn't complete in the earlier PR. I'm on board with segment! Will merge after a rebase.

hseeberger:
Cool. Rebased.

Issue #12, opened 2018-11-08
Nail down wildcard/fallback rules


aturon:
The initial implementation of the router uses *greedy* matching for URL segments. Imagine you have two routes:

- `foo/{}/baz`
- `foo/new/bar`

The URL `foo/new/baz` will *fail to be routed*, because the `new` component will be greedily matched (concrete segments are always preferred over wildcards) and there is no backtracking.

This behavior is the simplest to implement, but it's not clear that it's the most *obvious* behavior. OTOH, it's not clear that the URL *should* match here. And in general, this kind of routing situation seems like an anti-pattern, and is perhaps something we should detect and disallow.

Thoughts welcome!

tirr-c:
Maybe we can use regex to approach this. We can convert path expression into regex mechanically. Might contain some errors, but here's the rules:
* Prepend `^` and append `$`
* `{}` &rarr; `([^/]+)`
* `{}*` &rarr; `(.*)`
* `{foo}` &rarr; `(?P<foo>[^/]+)`
* `{foo}*` &rarr; `(?P<foo>.*)`

This way, `/foo/{}/baz` becomes `^/foo/([^/]+)/baz$` and `/foo/new/bar` becomes `^/foo/new/bar$`. Testing both regexes on `/foo/new/baz`, we can easily find out the path matches only the first one.

Bullet points:
* This can be done easily, and resolves most of the counterintuitive cases
* `regex` crate is highly optimized, and we can benefit from that
* We should think about detecting duplicated routes

yoshuawuyts:
A scenario I'd like us to consider is the way GitHub does routing for users / organizations. Essentially I'd like us to be able to support a similar URL structure.

|Route | Example
|----|---
| `/:user/:repo` | `/yoshuawuyts/blog`
| `/:user/:repo/settings` | `/yoshuawuyts/blog/settings`
| `/:user/:repo/tree/:branch` | `/yoshuawuyts/blog/tree/master`
| `/:user/:repo/tree/:branch` | `/yoshuawuyts/blog/tree/master`
| `/:user/:repo/blob/:branch/*` | `/yoshuawuyts/blog/blob/master/content/2019-01-13-wasm-2019.md`
| `/pulls` | `/pulls`

bIgBV:
But ins't the routing structure specific to the way the app is structured? I mean you could have a RESTful API or you could have something arbitrary right?

Nemo157:
[...]

PR #93, opened 2018-11-28
Implement "catch-all" endpoint

tirr-c:
This PR adds the concept of "wildcard modifers". Only one modifier is implemented for now: `*`, the zero-or-more modifier. It matches zero or more path segments. For example, `/static/{path}*` matches `/static/foo.txt`, `/static/images/bar.jpg`, and `/static`. `path` will be `"foo.txt"`, `"bar.jpg"`, and `""`, respectively.

* `{}*` can be written as `*`.
* Since this "catch-all" is still a wildcard, concrete segments take precedence. With `/static/foo/bar` and `/static/*`, `/static/foo` will **fail to route**. We might need to solve this problem (#12).

### DIFF
--- a/path_table/src/lib.rs
+++ b/path_table/src/lib.rs
@@ -215,15 +215,9 @@ impl<R> PathTable<R> {
 
             let wildcard_opt = if segment.starts_with('{') {
                 if segment.ends_with('}') {
-                    Some((
-                        &segment[1..segment.len() - 1],
-                        WildcardKind::Segment,
-                    ))
+                    Some((&segment[1..segment.len() - 1], WildcardKind::Segment))
                 } else if segment.ends_with("}*") {
-                    Some((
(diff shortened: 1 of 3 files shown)


path_table/src/lib.rs
```diff
@@ -17,9 +17,25 @@ pub struct PathTable<R> {
     wildcard: Option<Box<Wildcard<R>>>,
 }
 
+#[derive(Copy, Clone, PartialEq)]
+enum WildcardCountModifier {
```
tzilist: It might make sense to rename this enum to `WildcardKind` with the variants of `Segment` and `Catchall` or something akin to this. `Segment` would be used for path segments and `Catchall` would be used for the catchall mechanism you've written here.

tzilist:
looks great! feel free to merge whenever you are ready @tirr-c

Issue #17, opened 2018-11-08
Add code examples to API docs


aturon:
The existing API docs were written quickly and don't have examples yet. Let's get some added!

pbvie:
I'd like to take on this one. Seems like a great way to get to know the project.

zacck:
How come this is open, do we need more code examples?

yoshuawuyts:
@zacck that was the idea. But we already have quite a few docs, so there's not too much use in keeping this open. Perhaps we should do a docs push at some point though, but we can open a new issue for that.

PR #97, opened 2018-11-29
Add examples to API docs

pbvie:
Add code examples to API docs.

## Description
Added some examples to most of the core components (app, endpoint, body, head).

## Motivation and Context
See #17 

## How Has This Been Tested?
Tested with `cargo test --doc`

## Types of changes

## Checklist:

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -24,6 +24,61 @@ use crate::{
 /// Apps are equipped with a handle to their own state (`Data`), which is available to all endpoints.
 /// This is a "handle" because it must be `Clone`, and endpoints are invoked with a fresh clone.
 /// They also hold a top-level router.
+///
+/// # Examples
+///
+/// You can start a simple Tide application that listens for `GET` requests at path `/hello`
+/// on `127.0.0.1:7878` with:
+///
+/// ```rust, no_run
+/// #![feature(async_await)]
+///
+/// let mut app = tide::App::new(());
+/// app.at("/hello").get(async || "Hello, world!");
+/// app.serve("127.0.0.1:7878")
+/// ```
+///
+/// Create a new `App` with a `Data` handle and make a clone of that handle available to endpoints:
+///
+/// ```rust, no_run
+/// #![feature(async_await, futures_api)]
+///
+/// use std::sync::Arc;
+/// use std::sync::Mutex;
+/// use tide::AppData;
+/// use tide::body;
+///
+/// #[derive(Clone)]
+/// struct Database {
+///     contents: Arc<Mutex<Vec<String>>>,
+/// }
+///
+/// impl Database {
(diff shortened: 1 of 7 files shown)


src/app.rs
```diff
@@ -24,6 +24,61 @@ use crate::{
...
+/// ```
+///
+/// Create a new `App` with a `Data` handle and make a clone of that handle available to endpoints:
+///
+/// ```rust, no_run
+/// #![feature(async_await, futures_api)]
+///
+/// use std::sync::Arc;
+/// use std::sync::Mutex;
+/// use tide::AppData;
```
bIgBV: Is it possible to elaborate on `AppData`? Right now it seems a little opaque as to why the `Database` I passed as a parameter to `tide::App::new` is accessible as an `AppData` to my endpoint.

pbvie: Thanks for the feedback. I've updated the description for the example - let me know what you think.

src/body.rs
```diff
@@ -2,7 +2,78 @@
...
+//!     println!("String: {}", *msg);
+//!     format!("{}", *msg)
+//! }
+//!
+//! async fn echo_string_lossy(msg: body::StrLossy) -> String {
+//!     println!("String: {}", *msg);
+//!     format!("{}", *msg)
+//! }
+//!
+//! async fn echo_vec(msg: body::Bytes) -> String {
```
aturon: Why not return a `Vec<u8>` here?

src/app.rs
bIgBV: Yes, this explanation is really good!

src/body.rs
pbvie: Thank you!

Returning a `Vec<u8>` is not working for me...maybe because there's no `IntoResponse` defined for it?

src/app.rs
pbvie: Thanks!

src/body.rs
tzilist: that is probably it - that should be pretty easy to add though. You can see (here)[https://github.com/rust-net-web/tide/blob/master/src/response.rs#L27] we are turning a string into a `Vec<u8>` anyways :P

pbvie: I'll add it :+1: But I think it would be cleaner to create a separate PR since it has nothing to do with #17 - thoughts?

aturon: I don't mind it being in the same PR, it can still be a separate commit.

pbvie: Ok, added https://github.com/rust-net-web/tide/pull/97/files#diff-6cfc163228cc31bed4a7214fa0b69300R23 and updated the examples for body types.

aturon:
Wonderful, thank you so much!

PR #98, opened 2018-11-30
add elapsed field to logger

leaxoy:


KodrAus:
Hi @leaxoy! :wave:

It's definitely useful to be able to collect timings about how long a piece of middleware takes to execute, but in my opinion this is best solved through some kind of contextual logging where we can choose what blocks of code to wrap for timings.

leaxoy:
Wrap the user logic maybe useful. Then we can add some code around `RequestContext.next` call with insert time field to `Request.extensions`, which can be used whenever it's needed.

prasannavl:
Hi @leaxoy! Does #222 resolve this? :)

leaxoy:
Looks cover this.

Issue #60, opened 2018-11-25
[Discussion] 0.1 release criteria


aturon:
Let's brainstorm about what we'd like to get done before making a 0.1 release to crates.io!

tirr-c:
It will be nice if we have these features:

* Custom default handler (#10)
  - @tzilist is implementing this.
  - In addition, we might want to run a chain of middleware *even if* routing fails. Some middleware like compression should run regardless of the routing result.
* Configuration (#5)
  - I have a vague idea on the design, which can cover default configuration, initialization from file/env, and per router/endpoint overriding. But we don't need to implement all at once; I think it's okay to start with defaults and overriding.

sendilkumarn:
For 0.1, would love to see #51 landing. This will enable us to create a complete web app (easily) with a separate frontend with any framework and backend with tide and diesel. 

P.S. I have a working example with tide + diesel + frontend.

aturon:
@tirr-c 

> In addition, we might want to run a chain of middleware even if routing fails. Some middleware like compression should run regardless of the routing result.

Yeah, I was wondering how we might want to approach that. Want to open up an issue for further discussion?

I'm going to set up a 0.1 milestone and we can start marking issues, including the ones y'all have mentioned here. I'll also open some additional issues around docs.

aturon:
OK, here's the [0.1 milestone](https://github.com/rust-net-web/tide/milestone/1), populated with most of the things mentioned here, other than a new issue for "always-applied" middleware.

aturon:
@sendilkumarn 

> I have a working example with tide + diesel + frontend.

😍

tirr-c:
Opened #61 and added it to the milestone.

tzilist:
> @tirr-c
> 
> > In addition, we might want to run a chain of middleware even if routing fails. Some middleware like compression should run regardless of the routing result.
> 
> Yeah, I was wondering how we might want to approach that. Want to open up an issue for further discussion?
> 
> I'm going to set up a 0.1 milestone and we can start marking issues, including the ones y'all have mentioned here. I'll also open some additional issues around docs.

In regards to this, I am thinking that I will have the router handle this instead of how we are handling it [here](https://github.com/rust-net-web/tide/blob/master/src/app.rs#L114-L118). I suppose my idea here is that instead of having to have separate logic around whether a route is found or not, the user can just specify a response (just like any other endpoint) and the middleware will behave the same. Not sure what you guys think about this but let me know :)

[...]

PR #100, opened 2018-11-30
[WIP] Proof of proof of concept for Tide-global state

bIgBV:
## Description
Proof of concept for global state internal to Tide.

## Motivation and Context
Currently we have the following form of state in Tide:

(Quoting @tirr-c ):

> - App-global across-middleware data at Data
> - App-global middleware-local data inside each Middleware
> - Request-local across-middleware data at extension (or wherever)
> - Request-local middleware-local data inside handle

But this does not provide a way to save state internal to Tide, but accessible globally within the framework itself, as an example design for #5.

This means that things which need to be accessible internally to Tide such as say a `Config` struct and a root logger are no longer accessible.

This PR provides a proof of concept by using a typemap wrapped in a reader-writer lock to provide synchronized access to this global state.

## How Has This Been Tested?
I modified the logging middleware and added a dummy `Config` struct to demonstrate particular usecases for such a system. Right now if an example is run, we log a message mentioning what environment the server is running in and what address is it bound to.

The biggest issue I have with this design is:

### Having to access the `STATE` variable directly. 
I think a better design would be to wrap the `STATE` typemap with our own struct and provide a `pub get` method and a `pub(crate insert` method. This way we provide a way for the users of the framework to only read the application configuration, while internally within the framework we have read-write access.

This would mean that we can provide route level configuration as mentioned in #5 as well.

I tried to implement this in this design, but couldn't make it work with the current design.

## Types of changes

## Checklist:


tirr-c:
If two `App`s are created, they'll share the same state. It's not very likely to happen though... Is this intended?

bIgBV:
@tirr-c no it isn't. I didn't realize that possibility. I think one way to handle this is to wrap it in a struct with a id of some sort and instantiate one for every app instance.

bIgBV:
@aturon is the general idea behind an approach like this viable? I feel like this is something like this will be required for #60 as it ties directly into two major concerns, being logging and app configuration.

bIgBV:
I've been trying to somehow add a `Entry` API to the underlying `AnyMap` which also includes handling locking, but it is a little beyond my rust skills 😅. I think it is possible, but the issue I'm running into right now is, with the AnyMap, you get `Option<&T>` or `Option<T>` values, which doesn't really work when trying to add an `entry` method to it.

On the other hand `TypeMap` exposes a `Entry` struct, but wrapping the map in a lock means that it you end up with something similar to the current API anyway. 

Basically what I'm looking to return from functions such as `get` and `insert` is something along the lines of `<RwLockReadGuard<Entry<K>> where K: Key`.

aturon:
@bIgBV so I agree that the key missing thing right now is a way to communicate data with middleware and endpoints during the "static" phase (that is, for configuration). And I also agree that we'll ultimately want this configuration to be easily loadable from the external environment.

That said, there are a few reasons that I'd *prefer* to achieve these goals more along the lines of https://github.com/rust-net-web/tide/issues/5:

- It allows for "static configuration" to be tweaked at any point in the router hierarchy, rather than just globally

- It avoids the need to build in a global `RwLock`, because (in principle) you should have mutable access via the `App` during setup.

However, the key thing to making that work will be getting some variant of typemap working for configuration. I left a comment with some ideas [here](https://github.com/rust-net-web/tide/issues/5#issuecomment-446367707), lemme know what you think!

bIgBV:
With the discussion in #5 and the new PR for configuration #109 , I feel this PR is no longer needed.

Issue #96, opened 2018-11-29
Consider renaming `App::at()` to `App::path()`


kud1ing:
I propose to rename `App::at()` to `App::path()`.

`at()` is short and thus easy to type, but it bears not much context.

I think

    app.path("/").get(index_endpoint)

would be slightly more descriptive/readable than

    app.at("/").get(index_endpoint)

yoshuawuyts:
It seems most of the discussion around this is currently happening on the PR #103.

PR #103, opened 2018-12-03
change app to path for App

Dylan-DPC:
## Description
Changes `at()` to `path()` for App

## Motivation and Context
#96 

## How Has This Been Tested?
No new tests added. Existing tests modified to reflect the change

## Types of changes

## Checklist:

### DIFF
--- a/examples/body_types.rs
+++ b/examples/body_types.rs
@@ -41,11 +41,11 @@ async fn echo_form(msg: body::Form<Message>) -> body::Form<Message> {
 fn main() {
     let mut app = tide::App::new(());
 
-    app.at("/echo/string").post(echo_string);
-    app.at("/echo/string_lossy").post(echo_string_lossy);
-    app.at("/echo/vec").post(echo_vec);
-    app.at("/echo/json").post(echo_json);
-    app.at("/echo/form").post(echo_form);
+    app.path("/echo/string").post(echo_string);
+    app.path("/echo/string_lossy").post(echo_string_lossy);
+    app.path("/echo/vec").post(echo_vec);
+    app.path("/echo/json").post(echo_json);
+    app.path("/echo/form").post(echo_form);
 
     let address = "127.0.0.1:8000".to_owned();
     println!("Server is listening on http://{}", address);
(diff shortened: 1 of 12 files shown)


tzilist:
this looks good to me, just want to cc @aturon to make sure he's alright with it!

aturon:
I don't have a strong preference here, but I'd like to make sure others are on board! cc @bIgBV @tirr-c any thoughts/preferences?

bIgBV:
Both `at` and `path` to me are pretty similar from an implementation stand point. Though that said, I feel that `.at()` communicates the intent of the operation better

> You are mounting an _endpoint_ for the given _method_, _at_ the given _path_

So calling the method `.path()` changes that, and consequently, the documentation surround explaining the concepts becomes more complex.

> You are mounting an _endpoint_ for the given _method_, at the _path_ by calling the `app.path()` method.

But, there might be a better way to communicate this idea, and in that case `.path()` works.

DeltaManiac:
My understanding of `at` was _do something_ _at_ a _path/resource_ , very much like how @bIgBV put it.
Even though `path` might work, I'd still prefer it to be the `at` a path/resource.

Dylan-DPC:
There isn't much to pick between the 2, but imo `path` feels more readable to me.

connectkushal:
as a beginner i think `path` would be more distinct and understandable than `at`. Im more comfortable with a function name that is understandable when looking at the code rather than a function name which makes language of the documentation proper. (code readability over documentation convinience)

mmrath:
I was just playing with tide and `at` feels better to me.

yoshuawuyts:
Something I haven't seen raised before is that `path` already has a meaning within Rust. Namely [`Path`](https://doc.rust-lang.org/std/path/struct.Path.html) and [`PathBuf`](https://doc.rust-lang.org/std/path/struct.PathBuf.html) are used to describe _file system locations_.

It might be a gotcha that Tide's `.path()` method does not take a `Into<Path>` but a string slice. The fact that path now has subtly different behaviors might cause potential issues down the line.

This can be entirely evaded by not using `.path()`, but using a different name instead. For example the existing `.at()` method evades this. Or else something like `.route()` might work (because we add _routes_ to the _router_ using the _`.route()`_ API).

But personally I think `.at()` is good enough actually, and don't think it's worth changing.

Dylan-DPC:
Going by the comments, `at()` looks like a better option so I'm okay with closing this. Thanks

PR #105, opened 2018-12-06
Add Cookie

### LINKED ISSUE #84: Ship cookie parsing as a computed value
As a follow up to https://github.com/rust-net-web/tide/pull/80, the [cookie example](https://github.com/rust-net-web/tide/blob/master/examples/computed_values.rs#L8-L27) should be polished up and moved into Tide proper, so that applications can use it.

c410-f3r:
Integrates cookie into Tide.

## Description

This PR adds the `Cookie` and the `CookieBuilder` structs into Tide, however, it doesn't have a default implementation for the `Compute` trait and it also doesn't include a test suite to ensure that implementations follow the rules specified in [RFC 6265](https://tools.ietf.org/html/rfc6265).

## Motivation and Context

Fixes #84 

## How Has This Been Tested?

Added some documentation tests.

## Types of changes

## Checklist:

### DIFF
--- a/src/cookie.rs
+++ b/src/cookie.rs
@@ -1,0 +1,313 @@
+// Based on https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Cookie.java
+// and http://doc.qt.io/qt-5/qnetworkcookie.html.
+
+use std::time::Duration;
+
+/// Represents an HTTP cookie. It has a name, a single value and optional parameters to
+/// maintain persistent information across requests.
+///
+/// Check out the [official specification](https://tools.ietf.org/html/rfc6265).
+#[derive(Clone, Debug)]
+pub struct Cookie<S> {
+    domain: Option<S>,
+    expires_at: Option<Duration>,
+    is_host_only: bool,
+    is_http_only: bool,
+    is_persistent: bool,
+    is_secure: bool,
+    name: S,
+    path: Option<S>,
+    value: S,
+}
+
+impl<S> Cookie<S>
+where
+    S: AsRef<str>,
+{
+    /// Creates a new [`Cookie`](Cookie) from the full range of possible parameters.
+    ///
+    /// # Parameters
+    ///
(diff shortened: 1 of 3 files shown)


c410-f3r:
Sorry, I will be away for a couple of days and won't be able to fix things in the meanwhile.

src/cookie.rs
```diff
@@ -0,0 +1,252 @@
...
+{
+    /// If any, returns the domain associated with this
+    /// cookie. E.g.: `Set-Cookie: name=value; Domain=.foo.com`.
+    pub fn domain(&self) -> Option<&str> {
+        self.domain.as_ref().map(|x| x.as_ref())
+    }
+
+    /// If any, returns the time this cookie expires.
+    ///
+    /// A value less than the current time means that the cookie has been expired.
```
yoshuawuyts: Perhaps we should clarify that this is a duration since the unix epoch. It took a bit of browsing to figure out *which* duration we were talking about, haha.

yoshuawuyts: Alternatively this API could perhaps expose the value as a `SystemTime` instance so that expiration dates become non-relative, and possibly easier to work with.

The tradeoff might be that the API no longer matches the cookie API 1:1.

src/cookie.rs
```diff
@@ -0,0 +1,252 @@
...
+    host_only: bool,
+    http_only: bool,
+    name: Option<S>,
+    path: Option<S>,
+    persistent: bool,
+    secure: bool,
+    value: S,
+}
+
+impl<S> CookieBuilder<S>
```
yoshuawuyts: All of the methods on `CookieBuilder` should probably be tagged with `#[inline]`.

yoshuawuyts:
@c410-f3r Left some comments, but overall this looks pretty good! You mentioned you don't have much time; I'd be happy to push some fixes if you prefer!

@aturon I'd be keen to hear what you think of the overall design.

c410-f3r:
Closing this PR since most people seem to prefer #114.

PR #109, opened 2018-12-14
Per-endpoint configuration

### LINKED ISSUE #5: Add configuration (for extractors and middleware)
Most extractors and middleware should be configurable. But that presents some challenges:

- Extractors are given as *argument types* to endpoints, so there's no obvious value to attach configuration to.
- Middleware configuration may want to be tweaked in route-specific ways.

One approach for solving these problems is to use a typemap for configuration, allowing extractors and middleware to provide their knobs as public types that can be installed into the map. This map would then be built as part of constructing the router:

```rust
let mut app = App::new();

// configure json extractors to use a 4k maximum payload by default
app.config(json::MaxPayload(4096));

// set up the `foo` endpoint and override its configuration
app.at("foo")
    .put(foo_endpoint)
    .config(json::MaxPayload(1024));

// for all routes starting at `bar`...
app.at("bar").nest(|router| {
    // configure the json extractor to return a 400 on parsing failure
    router.config(json::OnError(StatusCode::BAD_REQUEST));

    // ...
})
```

(Note: the `nest` example uses the feature proposed [here](https://github.com/rust-net-web/tide/issues/4)).

Note that this functionality *could* be implemented using middleware and the `extensions` part of the request, but that would mean the configuration is constructed fresh during request handling, rather than when the app is constructed.

Let's reach consensus on a design, and then implement this!

tirr-c:
## Description
This PR adds per-endpoint configuration mechanism.

Each configuration item is represented as a value of its own type. Items can be applied to endpoints and (sub)routers, and child configuration inherits from that of parent. The point when `config` is called does not matter; when the router setup ends, all configuration items of the router are applied to the children (except overridden).

The first three commits implements the feature with `serde` and `toml`. The other commits use a minimal, custom typemap for configuration.

## Motivation and Context
Resolves #5.

Currently we don't provide any way to configure middleware, extractors etc. Using configuration, we can adjust various aspects of how components of Tide work, like limiting request body size.

## How Has This Been Tested?
* An example application (`examples/configuration.rs`). Tested with HTTPie and works as expected.
* New unit test, some for `Configuration` itself and some for `Router`.

## Types of changes
  - Some function signatures have changed.

## Checklist:


src/configuration.rs
```diff
@@ -0,0 +1,128 @@
...
+        let id = TypeId::of::<T>();
+        self.0
+            .insert(id, Box::new(value) as Box<dyn ConfigurationItem>);
+    }
+}
+
+/// An extractor for reading configuration from endpoints.
+///
+/// It will try to retrieve the given configuration item. If it is not set, the extracted value
+/// will be `None`.
```
bIgBV: Wouldn't this be `Ok(None)`?

src/middleware/mod.rs
```diff
@@ -28,11 +29,15 @@ pub struct RequestContext<'a, Data> {
     pub app_data: Data,
     pub req: Request,
     pub params: Option<RouteMatch<'a>>,
-    pub(crate) endpoint: &'a BoxedEndpoint<Data>,
+    pub(crate) endpoint: &'a EndpointData<Data>,
     pub(crate) next_middleware: &'a [Arc<dyn Middleware<Data> + Send + Sync>],
 }
 
 impl<'a, Data: Clone + Send> RequestContext<'a, Data> {
+    pub fn get_config<T: Any + Clone + Send + Sync>(&self) -> Option<&T> {
```
bIgBV: Since this is returning a config item, I feel like `get_config_item` would be a more descriptive name.

tirr-c: Agreed! I've changed the name.

src/configuration.rs
tirr-c: I meant the inner value, so `None` would be okay.

examples/configuration.rs
```diff
@@ -0,0 +1,25 @@
+#![feature(async_await, futures_api)]
+
+use tide::{head::Path, ExtractConfiguration};
+
+#[derive(Clone, Default)]
+struct IncreaseBy(i32);
+
+async fn add(
+    Path(base): Path<i32>,
+    ExtractConfiguration(amount): ExtractConfiguration<IncreaseBy>,
```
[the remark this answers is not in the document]
tirr-c: Sure! I've added comments to the example.

bIgBV:
Looks good to me! I'd like to build on top of this to get a default config set up. @tirr-c do you think I should create a branch off of your branch, or wait until the PR is merged after @aturon can have a chance to look at it?

tirr-c:
@bIgBV Go ahead! It would be really helpful if I can get some feedback about the API design.

src/configuration.rs
```diff
@@ -0,0 +1,128 @@
+//! Types for managing and extracting configuration.
+
+use std::any::{Any, TypeId};
+use std::collections::HashMap;
```
yoshuawuyts: A possible performance improvement here would be to use the [hashbrown](https://github.com/Amanieu/hashbrown) crate.

examples/configuration.rs
```diff
@@ -0,0 +1,29 @@
+#![feature(async_await, futures_api)]
+
+use tide::{head::Path, ExtractConfiguration};
+
+/// A type that represents how much value will be added by the `add` handler.
+#[derive(Clone, Default)]
```
yoshuawuyts: Should this also include `Debug`?

src/configuration.rs
```diff
@@ -0,0 +1,128 @@
...
+impl Clone for Box<dyn ConfigurationItem> {
+    fn clone(&self) -> Box<dyn ConfigurationItem> {
+        (&**self).clone_any()
+    }
+}
+
+/// A cloneable typemap for saving per-endpoint configuration.
+///
+/// Configuration is mostly managed by `App` and `Router`, so this is normally not used directly.
+#[derive(Clone)]
```
yoshuawuyts: How should people debug this? If we're not adding the `Debug` trait, perhaps it would be useful to leave a note on how to inspect the values, or mention that it's currently not possible.

src/configuration.rs
tirr-c: AFAICT hashbrown is used only in rustc, not libstd.

src/configuration.rs
tirr-c: Well, I think there are two options on implementing `Debug` for `Store`:
* Require `Debug` for all store items.
* Enable specialization, and specialize `Debug::fmt` for types that implements `Debug`.

First option works without any unstable features, but all items should implement `Debug`. Second one is unstable but potentially flexible as it doesn't require `Debug`. It seems like the former is better as deriving `Debug` is pretty trivial, but I'm not sure. How do you think, @yoshuawuyts?

tirr-c:
@bIgBV No worries, I'll handle those!

src/app.rs
```diff
@@ -137,12 +168,18 @@ impl<Data: Clone + Send + Sync + 'static> App<Data> {
     /// Start serving the app at the given address.
     ///
     /// Blocks the calling thread indefinitely.
-    pub fn serve<A: std::net::ToSocketAddrs>(self, addr: A) {
+    pub fn serve(self) {
+        let configuration = self.get_item::<Configuration>().unwrap();
+        let addr = format!("{}:{}", configuration.address, configuration.port)
+            .parse::<std::net::SocketAddr>()
+            .unwrap()
+            .clone();
```
bIgBV: Removing the .clone() call should remove the warning. 

```
    let addr = format!("{}:{}", configuration.address, configuration.port)
            .parse::<std::net::SocketAddr>()
            unwrap();
```

As an aside, github should really allow multi-line suggestions.

src/configuration/default_config.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub address: String,
+    pub port: u16,
+}
+
+pub struct ConfigurationBuilder {
+    pub env: Environment,
+    pub address: String,
+    pub port: u16,
+}
+
```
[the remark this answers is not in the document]
tirr-c: Yep, I've extracted `new` to the `Default` impl. The constructor of `ConfigurationBuilder` is still there but now it's less visible (to humans).

src/configuration/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+pub struct ExtractConfiguration<T>(pub Option<T>);
+
+impl<S: 'static, T: Any + Clone + Send + Sync + 'static> Extract<S> for ExtractConfiguration<T> {
+    type Fut = FutureObj<'static, Result<Self, Response>>;
+
+    fn extract(
+        data: &mut S,
+        req: &mut Request,
+        params: &Option<RouteMatch<'_>>,
+        config: &Store,
```
PeteJodo: Incredibly pedantic but it seems like the naming of this is inconsistent with everything else implementing `extract`; across the board it's been named `store`. Just wanted to point that out

tirr-c: Oops, seems like I've missed this one!

src/configuration.rs
yoshuawuyts: @tirr-c I think requiring `Debug` on all items is a reasonable constraint. In terms of semver it should probably be OK to loosen that constraint eventually (e.g. with specializations, maybe?), but we adding `Debug` as a constraint down the line would be a breaking change.

So I think it would be good to go with the first option, and we can always decide to loosen the constraint in the future.

yoshuawuyts:
Builds are failing because nightly can't be installed https://github.com/rust-lang/rust/issues/57488. I think we should be good to merge once we can get tests to pass!

src/configuration.rs
tirr-c: @yoshuawuyts I've added `Debug` impl for `Store`! It looks pretty good, check the `configuration` example.

PeteJodo:
I love this but I have some issues with it that I can't quite put my finger on exactly yet. I guess one issue is the naming of some things e.g. `ExtractConfiguration` seems odd since other types that implement Extract don't have `Extract-` prepended. For example, `Path` isn't named `ExtractPath`. Now I don't have any thing better to suggest myself since there's already the `Configuration` type. I'll probably play around with this once it's merged to see what I can come up with.

bIgBV:
So looks like travis was failing on installing nightly in the container. I just restarted the build and looks like whatever issue was occurring a couple of days ago is now solved.

Are there any other blockers for this PR as I want to get another PR open for a default config file and the related supporting code open. Plus once the discussion in #8 reaches some consensus, I'm sure this will help a lot for that as well.

yoshuawuyts:
Travis is passing again, and from the comments there seem to be no blockers. I propose we merge!

src/configuration.rs
yoshuawuyts: Oh I missed this. But yeah, it's landing in libstd; people made a PR during RustFest Rome for this!

PR #114, opened 2018-12-28
Basic cookie extractor

### LINKED ISSUE #84: Ship cookie parsing as a computed value

mmrath:
## Description
Introduces a Cookies struct which wraps CookieJar from [cookie-rs](https://github.com/alexcrichton/cookie-rs). The example files has usage details.

Most likely this will fail in test but that is because of existing issues around feature-preview crate

## Motivation and Context
Fixes #84 

## How Has This Been Tested?
Example is written which can be executed directly. I would like a proper test case to be written but at the moment I am not too sure how the testing infrastructure need to be setup.

## Types of changes

- [*] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,4 +38,7 @@ version = "0.3.0-alpha.12"
 
 [dev-dependencies]
 juniper = "0.10.0"
-basic-cookies = "0.1.2"
+
+[dev-dependencies.basic-cookies]
+git = "https://github.com/yoshuawuyts/basic-cookies"
+branch = "fix-cookie-token"


src/cookies.rs
```diff
@@ -0,0 +1,48 @@
...
+use crate::response::IntoResponse;
+use crate::{Extract, Request, Response, RouteMatch};
+
+#[derive(Clone, Debug)]
+pub struct Cookies {
+    content: CookieJar,
+}
+
+impl Cookies {
+    pub fn get(&self, name: &str) -> Option<&Cookie<'static>> {
```
yoshuawuyts: This might be good to mark as `#[inline]`

bIgBV:
My vote goes to this PR, having `Extract` implemented for cookies feels a lot more ergonomic as compared to using a computed value.

examples/cookie_extractor.rs
```diff
@@ -0,0 +1,16 @@
+#![feature(async_await, futures_api)]
+
+use tide::Cookies;
+
+async fn hello_cookies(cookies: Cookies) -> String {
```
bIgBV: Some documentation explaining what the `Cookies` type is could be helpful. In the long run, this could be an example of how the `Extract` trait works in a guide-like doc.

src/cookies.rs
```diff
@@ -0,0 +1,48 @@
...
+        self.content.get(name)
+    }
+}
+
+impl<S: 'static> Extract<S> for Cookies {
+    // Note: cannot use `existential type` here due to ICE
+    type Fut = future::Ready<Result<Self, Response>>;
+
+    fn extract(data: &mut S, req: &mut Request, params: &Option<RouteMatch<'_>>) -> Self::Fut {
+        match req.headers().get("Cookie") {
```
tirr-c: I came up with an idea that might make the code easier to read:
1. Make `Result<CookieJar, ParseError>` at the `match req.headers().get("Cookie")` expression. For example, The `None` arm would return `Ok(CookieJar::new())`.
1. Transform that value into `Result<Cookies, Response>`.
1. Wrap it with `future::Ready`.

By doing this, we can easily see we're extracting `CookieJar` and then wrapping it with `Cookies`.

src/cookies.rs
```diff
@@ -0,0 +1,48 @@
+use cookie::Cookie;
```
PeteJodo: `use cookie::{Cookie, CookieJar, ParseError};` or am I missing something?

mmrath: > `use cookie::{Cookie, CookieJar, ParseError};` or am I missing something?

I was lazy and left it to rustfmt. Looks like neither rustfmt nor CLion is that smart yet. I will fix it.

src/cookies.rs
mmrath: good way. I updated the PR.

examples/cookie_extractor.rs
mmrath: @blgBV you mean to say that Cookies type is a wrapper around CookieJar and open for future extensions? something in those lines? or more documentation on the example or both?
I am happy to add more documentation but need some pointers

yoshuawuyts: @mmrath I can't speak for @bIgBV but I was thinking explaining high-level what this function does, and why it works this way might be useful. Something like:

> Extract the cookies from the HTTP header into a Map. This is done through the `Cookies` extractor.

To people unfamiliar with Tide this method might seem rather magical, and having a few pointers on where to look might be useful.

Separate from this we should probably write some docs on how extractors work, in addition to the [existing blog post](https://rust-lang-nursery.github.io/wg-net/2018/10/16/tide-routing.html).

yoshuawuyts:
ping @aturon do you have any input on the design?

examples/cookie_extractor.rs
bIgBV: Yup, this is basically what I had in mind.

mmrath:
@yoshuawuyts @aturon  is this ok to merge? Please let know if any more work required on this. build issue appears to be not related to this PR.

aturon:
The design LGTM! Thanks @mmrath!

Cargo.toml
```diff
@@ -24,6 +24,7 @@ serde_qs = "0.4.1"
 slog = "2.4.1"
 slog-term = "2.4.0"
 slog-async = "2.3.0"
+cookie="0.11"
```
PeteJodo: from what I understand, to use `secure` cookies, you need to enable that [feature](https://docs.rs/cookie/0.11.0/cookie/#features). Seems like that's something we'd want to include

mmrath: Yes that is true. But I wanted have the basic cookie extractor there before looking into secure cookie. Secure cookie needs `ring` which I believe is not a dependency yet.

Issue #26, opened 2018-11-08
Middleware for compression


aturon:
A good test of the middleware system would be to implement gzip compression as middleware (transforming the body and headers on the way in/out).

yoshuawuyts:
This would tie in nicely with https://github.com/rust-lang-nursery/wg-net/issues/52

fairingrey:
I took up this issue up roughly two weeks ago through https://github.com/rustasync/tide/pull/194 and progress is almost finished. There are some things I want to wrap up before I submit it for review.

PR #117, opened 2019-01-15
[WIP] New Tide middleware: Automatic compression handling

kimsnj:
## Description

This PR aims at brining a new compression handling middleware to Tide.
The goal of it would to automatically:

- decompress incoming data if it's compressed
- detect the preferred encoding from the request and encode the response accordingly.

This is still work-in-progress for several reasons:

### Missing features:
Some target features are not there yet (but I guess this could come with time)

- Decompression of request
- Brotli support
- Quality value support (by the way, I expected to find their parsing in http/hyper without success. Did I miss something?)

### Streaming compression
I struggled (and actually didn't manage) to compress the body as a stream of chunks.

Trying to use the `Stream` impl of `hyper::Body`, I encountered a few challenges:
- Compiler complaints as I get mixed up between futures 0.1 and 0.3 versions
- I was not familiar enough with the `flate2` interface. Can the `AsyncRead`/`AsyncWrite` version be used?

I guess I should probably manually implement a `Stream` and call [`flate2::Compress`](https://docs.rs/flate2/1.0.6/flate2/struct.Compress.html) raw struct.
Any hint on how to best go about this are very welcome :)

Meanwhile, to get something working, I used `read_to_vec` on `tide::Body`. This works well for small response… But will put more memory pressure payloads (e.g. streaming a file compressed).

### Error handling
For some errors, it is quite straight-forward to have a fallback behaviour (e.g. if compression fails, return body as is), but others are un-clear to me (e.g. what should happen in `read_to_vec` fails?).

## Motivation and Context
This follows from issue #26

## How Has This Been Tested?
- (automated) Unittests in the compression middleware
- (manual) Integrated to the hello world example and ping via `curl` with various headers

## Types of changes

## Checklist:

### DIFF
--- a/src/middleware/compression.rs
+++ b/src/middleware/compression.rs
@@ -1,0 +1,204 @@
+use futures::future::FutureObj;
+
+use crate::body::Body;
+use crate::{middleware::RequestContext, Middleware, Response};
+use http::header::{HeaderValue, ACCEPT_ENCODING, CONTENT_ENCODING};
+
+use std::io::Read;
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum Encoding {
+    Gzip,
+    Deflate,
+    Identity,
+}
+
+impl Encoding {
+    fn parse(i: &str) -> Option<Encoding> {
+        use self::Encoding::*;
+        match i {
+            "gzip" => Some(Gzip),
+            "deflate" => Some(Deflate),
+            "identity" => Some(Identity),
+            _ => None,
+        }
+    }
+
+    fn header_value(self) -> HeaderValue {
+        use self::Encoding::*;
+        match self {
+            Gzip => HeaderValue::from_str("gzip").unwrap(),
+            Deflate => HeaderValue::from_str("deflate").unwrap(),
+            Identity => HeaderValue::from_str("identity").unwrap(),
+        }
+    }
+}
+
+/// Tide's compression middleware.
+/// Will pick preferred supported format from "Accept-Encoding" header
+/// and serialize response accordingly
+pub struct Compression {
+    default: Encoding,
+}
+
+impl Compression {
+    pub fn new() -> Self {
+        Compression {
+            default: Encoding::Gzip,
+        }
+    }
+
+    pub fn with_default(default: Encoding) -> Self {
+        Compression { default }
+    }
+
+    fn preferred_encoding(headers: &http::HeaderMap) -> Option<Encoding> {
+        headers
+            .get_all(ACCEPT_ENCODING)
+            .iter()
+            .filter_map(|hval| hval.to_str().ok())
+            .flat_map(|val| val.split(|c| c == ','))
+            .filter_map(|encoding| Encoding::parse(encoding))
+            .next()
+    }
+
+    fn apply(body: &[u8], encoding: Encoding) -> Option<Vec<u8>> {
+        let mut buf = Vec::new();
+        match encoding {
+            Encoding::Gzip => {
+                let mut gz = flate2::bufread::GzEncoder::new(body, flate2::Compression::fast());
+                gz.read_to_end(&mut buf).ok()?;
+            }
+
+            Encoding::Deflate => {
+                let mut deflate =
+                    flate2::bufread::DeflateEncoder::new(body, flate2::Compression::fast());
+                deflate.read_to_end(&mut buf).ok()?;
+            }
+
+            _ => return None,
+        };
+        Some(buf)
+    }
+}
+
+impl Default for Compression {
+    fn default() -> Self {
+        Compression::new()
+    }
+}
+
+/// Picks best ecnoding from request and accordingly set response header and body.
+impl<Data: Clone + Send> Middleware<Data> for Compression {
+    fn handle<'a>(&'a self, ctx: RequestContext<'a, Data>) -> FutureObj<'a, Response> {
+        FutureObj::new(Box::new(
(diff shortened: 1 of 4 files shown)


src/middleware/compression.rs
```diff
@@ -0,0 +1,204 @@
...
+}
+
+impl Encoding {
+    fn parse(i: &str) -> Option<Encoding> {
+        use self::Encoding::*;
+        match i {
+            "gzip" => Some(Gzip),
+            "deflate" => Some(Deflate),
+            "identity" => Some(Identity),
+            _ => None,
```
yoshuawuyts: What is the difference between `Some(Identity)` and `None`? My understanding is that [they're both treated the same way](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding#Directives), so perhaps we could could combine these to make matching on them easier?

yoshuawuyts:
Overall this looks very promising! Thanks for putting in the work @kimsnj. Keen to see how this PR evolves!

src/middleware/compression.rs
kimsnj: Right now, the behavior for both is the same.
I suspect it might come in handy to have the explicit differentiation once quality value. 
A request containing `identity;q=0,unknown-compressio-algo` could (should?) respond [406 Not Acceptable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406)

aturon:
@kimsnj just checking in on this -- are you still planning to finish this PR?

kimsnj:
Hi @aturon,

TLDR: I must admit I have a bit abandoned it the last few months due to other priorities, but could be back to it in the upcoming weeks if it's not taken in the meantime.

----
Since the last update, I took a bit the time to play with `flate2` to try to figure out how to use compression for a `Stream`.
My findings so far would be to directly use [the raw Compress struct](https://docs.rs/flate2/1.0.7/flate2/struct.Compress.html) with a buffer held by the middleware for the (temporary) output, feeding each generated chunk to `flate2` and streaming back each of its outputs.
I have tried it successfully for Deflate compression in a stand-alone synchronous example, but didn't have time to integrate it.
GZip would then need an extra header and checksum.

Does this approach seems fine?

Regarding this PR, should I close it in the meantime to avoid cluttering the pending ones on this repo, and re-open it once I have an update ?

src/middleware/compression.rs
```diff
@@ -0,0 +1,204 @@
+use futures::future::FutureObj;
+
+use crate::body::Body;
+use crate::{middleware::RequestContext, Middleware, Response};
```
pickfire: Why not merge both the above as:
```rust
use crate::{body::Body, middleware::RequestContext, Middleware, Response};
```

pickfire: Why not merge both into
```rust
use crate::{body::Body, middleware::RequestContext, Middleware, Response};
```

src/middleware/compression.rs
```diff
@@ -0,0 +1,204 @@
...
+        Compression { default }
+    }
+
+    fn preferred_encoding(headers: &http::HeaderMap) -> Option<Encoding> {
+        headers
+            .get_all(ACCEPT_ENCODING)
+            .iter()
+            .filter_map(|hval| hval.to_str().ok())
+            .flat_map(|val| val.split(|c| c == ','))
+            .filter_map(|encoding| Encoding::parse(encoding))
```
pickfire:
```suggestion
            .filter_map(Encoding::parse)
```
IIRC, clippy have a warning related to that.

src/middleware/compression.rs
```diff
@@ -0,0 +1,204 @@
...
+            .next()
+    }
+
+    fn apply(body: &[u8], encoding: Encoding) -> Option<Vec<u8>> {
+        let mut buf = Vec::new();
+        match encoding {
+            Encoding::Gzip => {
+                let mut gz = flate2::bufread::GzEncoder::new(body, flate2::Compression::fast());
```
pickfire: Should we not use `flate2::Compression::default()` rather than `flate2::Compression::fast()`?

src/middleware/compression.rs
```diff
@@ -0,0 +1,204 @@
...
+            .get_all(ACCEPT_ENCODING)
+            .iter()
+            .filter_map(|hval| hval.to_str().ok())
+            .flat_map(|val| val.split(|c| c == ','))
+            .filter_map(|encoding| Encoding::parse(encoding))
+            .next()
+    }
+
+    fn apply(body: &[u8], encoding: Encoding) -> Option<Vec<u8>> {
+        let mut buf = Vec::new();
```
pickfire: I wonder if this `buf` is being allocated even if there is no encoding to be done (see the `None` block below)?

As well, I think it might be possible to not have an `Option` returned here?

fairingrey:
@kimsnj I'm willing to try working on this WIP middleware if you're busy, though it might happen in a separate PR.

kimsnj:
@fairingrey I'll close this PR for now so that you can pick it up if you want and are available.

If this is of any help, here is a [gist](https://gist.github.com/kimsnj/102e30722256287428a31130db2dca69) where I played around with flate API. I planned to use something like `raw_compress` to read from a stream and create a new one.

fairingrey:
@kimsnj Much appreciated! It will help, actually, as I'm figuring out a strategy for encoding the Body stream.

The other PR is here if you're interested in taking a look: https://github.com/rustasync/tide/pull/194

PR #124, opened 2019-01-19
change 'Content-Type' for IntoResponse::into_response(Vec<u8>)

DCjanus:
replace 'text/plain' with 'application/octet-stream'

## Description

There is no guarantee that Vec<u8> is valid UTF-8 byte sequence.

## Types of changes

## Checklist:


yoshuawuyts:
@DCjanus would you be interested in adding a test for this in the same file? It'd be great if we could prevent this from ever regressing on accident. I'd also be happy to pick this up if you're unavailable!

DCjanus:
Unit testing is a great thing, but I don’t think it’s necessary to write unit tests in such a short code.

yoshuawuyts:
@DCjanus It's because it's such a small detail that I think adding unit tests are a good idea. If a big refactor comes around, I could easily see this be the kind of detail that gets missed when converting code. Having a test around is both cheap, and helps make our code more resilient to refactors.

Anyway, no worries! -- I've added some simple tests for this to the PR :sparkles:

DCjanus:
reopen for rerun CI

DCjanus:
If there are no more questions, could you please re-run the CI and merge?

Issue #123, opened 2019-01-18
Using config-rs for layered configuration


bIgBV:
Now that #109 has landed, I am looking to extend it so that there are other possible ways of setting up the execution environment for a Tide application. When it comes to application deployments, the [12-factor app](https://12factor.net/) idea is standard.

The [config-rs](https://docs.rs/config/0.9.2/config/) crate provides good support for layering configuration from different sources. It also seems to be the most downloaded crate in the configuration category according [crates.rs](https://crates.rs/config) (To be fair, that metric is pretty short sighted).

The issue is that it makes the API for accessing application configuration becomes a little verbose. I feel that the flexibility afforded is worth is.

What are your thoughts on this?

yoshuawuyts:
> The twelve-factor app stores config in environment variables (often shortened to env vars or env). Env vars are easy to change between deploys without changing any code; unlike config files, there is little chance of them being checked into the code repo accidentally (...)

@bIgBV doesn't the 12 factor app standard disencourage the use of config files? From the docs `config-rs` seems to take in configuration from a lot of places, only one of which is the environment.

In my Rust applications I've been using `structopt` to define both flags and environment variables. For example: https://github.com/rust-clique/clap-port-flag/blob/master/src/lib.rs#L38.

I think it might be useful to suggest methods of configuration with Tide, but probably not to build into the framework directly.

yoshuawuyts:
> It just has to provide the mechanisms for arbitrary config data.

@PeteJodo The way I think about this relationship is:
- structopt gets data from the CLI, either from flags or env vars
- in the main fn this data gets passed into Tide's app + middleware constructors. This happens only once, at boot
- every request now is passed through middleware, and each `Store` (formerly `Config`) is populated with the corresponding data

---

> Tide really only relies on host and port

Structopt supports reusing flags! For address & port: https://github.com/rust-clique/clap-port-flag

bIgBV:
@yoshuawuyts you're right, it does discourage config files, I figured it wouldn't be too difficult to add support for that on top of CLI flags and env vars. Rocket uses a config file and I thought it would be useful to have a similar mechanism.

On the other hand, I did not know about structopt and I think it would be a really handy way of managing configuration. 

> * structopt gets data from the CLI, either from flags or env vars
> * in the main fn this data gets passed into Tide's app + middleware constructors. This happens only once, at boot
> * every request now is passed through middleware, and each Store (formerly Config) is populated with the corresponding 

[...]

PR #129, opened 2019-01-21
Structopt integration

bIgBV:
## Description

Integrate [`structopt`](https://docs.rs/structopt/) with the default configuration as discussed in #123 . This provides users a way provide application configuration through command line options, environment variables and programmatically override the configuration at app launch

## Motivation and Context

Environment specific configuration is a requirement for web application frameworks as depending on the environment, the application will have different means of communicating with external entities. These may be databases or external APIs. This PR provides a starting point for default application configuration which currently only includes the environment, address and port but could go on to include important items like database connection strings and template directory paths.

## How Has This Been Tested?

Manually tested the output for example applications.

## Types of changes

## Checklist:

    - Added new doctests for existing configuration builder


bIgBV:
@PeteJodo I purposefully avoided using a database host as a config option as I figured that there would be more discussion around connecting to external data stores and therefore I didn't want to make any assumptions. And yes, I think a rename would be better, but I think that should be a different PR. I'll open it once this lands.

bIgBV:
@yoshuawuyts so when you say an example, are you thinking of something along the lines of struct stored in the app store defined by the user, on which `StructOpt` is derived?

yoshuawuyts:
@bIgBV I was indeed thinking something along the lines of @PeteJodo's example.

bIgBV:
@yoshuawuyts @PeteJodo I'll open a PR adding a similar example and close this one. We can discuss weather or not we want to include this in the framework itself in a separate issue.

PR #139, opened 2019-02-21
Refactor to use HttpService internally

aturon:
## Description

This PR refactors Tide to internally use the new [http-service crate](https://github.com/rust-net-web/http-service). That crate provides an `HttpService` trait, meant to provide a Rack-like interface between low-level http servers (like hyper) and frameworks or applications that sit on top of them.

The refactoring isolates all hyper-specific shim code to a new module, `serve`, which presents an `HttpService`-based interface to the rest of the crate. Eventually this shim code should be moved into an external crate.

## Motivation and Context

The ultimate goals here are:

- To move the shim code to work with hyper into a separate crate
- To support a standardized layering such that the base web server can easily be switched out (in Tide and other web frameworks)

## How Has This Been Tested?

As a pure refactoring, no new tests are added.

## Types of changes

## Checklist:

### DIFF
--- a/src/body.rs
+++ b/src/body.rs
@@ -73,24 +73,14 @@
 //!
 //! ```
 //!
-use futures::{future::FutureObj, prelude::*};
+use futures::future::FutureObj;
 use http::status::StatusCode;
 use http_service::Body;
 use multipart::server::Multipart;
-use pin_utils::pin_mut;
-use std::io::{self, Cursor};
+use std::io::Cursor;
 use std::ops::{Deref, DerefMut};
 
 use crate::{configuration::Store, Extract, IntoResponse, Request, Response, RouteMatch};
-
-pub(crate) async fn body_to_vec(body: Body) -> Result<Vec<u8>, io::Error> {
-    let mut bytes = Vec::new();
-    pin_mut!(body);
-    while let Some(chunk) = await!(body.next()) {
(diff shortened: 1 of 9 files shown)


src/body.rs
```diff
@@ -73,118 +73,23 @@
...
-    }
-}
-
-impl Into<hyper::Body> for Body {
-    fn into(self) -> hyper::Body {
-        match self.inner {
-            BodyInner::Fixed(v) => v.into(),
-            BodyInner::Streaming(s) => hyper::Body::wrap_stream(s.compat()),
-        }
+pub(crate) async fn body_to_vec(body: Body) -> Result<Vec<u8>, io::Error> {
```
yoshuawuyts: Why was this chosen to become a free function rather than a method on the body? I think from reading uses of this, the method seemed to be a bit nicer.

aturon: Because the type moved out of the crate, so we can't provide it as an inherent method. A couple alternatives:

- Add it via an extension trait
- Add it to the http-service crate

I prefer the latter.

yoshuawuyts: D'oh, yes that makes perfect sense. Filed a tracking issue https://github.com/rust-net-web/http-service/issues/2.

aturon:
Alright, I've updated to use the new `Body::into_vec` method on http-service, so this is good to go!

PR #142, opened 2019-02-27
Add external crate http_service_mock, delete local TestBackend

gruberb:
This solves [ust-net-web/tide#15](https://github.com/rust-net-web/tide/issues/15). We wanted to extract the `HttpBackend` inside the tests, and offer it inside the `http-service` create (https://github.com/rust-net-web/http-service/pull/6).

## Description
I removed the `HttpBackend` from the `wildcard.rs` tests, and used the external crate (which needs to be merged and pushed first inside `http-service`).

## Motivation and Context
After the change, we are able to offer a TestBackend which is not tied to `Tide`.

## How Has This Been Tested?
The PR in `http-service` needs to be merged first. I tested it locally and it ran fine. 

## Types of changes

## Checklist:

### DIFF
--- a/tests/wildcard.rs
+++ b/tests/wildcard.rs
@@ -1,34 +1,11 @@
 #![feature(futures_api, async_await)]
 
-use futures::{executor::block_on, prelude::*};
-use http_service::{Body, HttpService, Request, Response};
+use http_service::{Body};
+use futures::{executor::block_on};
 use tide::{
-    head::{Named, NamedSegment},
-    Server,
+    head::{Named, NamedSegment}
 };
-
-struct TestBackend<T: HttpService> {
-    service: T,
-    connection: T::Connection,
-}
-
-impl<T: HttpService> TestBackend<T> {
-    fn wrap(service: T) -> Result<Self, <T::ConnectionFuture as TryFuture>::Error> {
-        let connection = block_on(service.connect().into_future())?;
-        Ok(Self {
-            service,
-            connection,
-        })
-    }
-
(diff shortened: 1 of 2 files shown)


tests/wildcard.rs
```diff
@@ -1,34 +1,11 @@
...
-
-    fn simulate(&mut self, req: Request) -> Result<Response, <T::Fut as TryFuture>::Error> {
-        block_on(
-            self.service
-                .respond(&mut self.connection, req)
-                .into_future(),
-        )
-    }
-}
+use http_service_mock::{make_server};
```
bIgBV: Since this is only being used in the test module, wouldn't moving it inside the tests module reduce overall binary size?

gruberb: Could you elaborate on this or expand your explanation? I don't seem to get it! As I see it, `http_servicemock` is already a `dev-dependency` and therefore won't be included in the build?

bIgBV: My bad, I didn't see that it was as already a dev-dependency.

tests/wildcard.rs
```diff
@@ -77,7 +49,8 @@ fn wildcard() {
 
 #[test]
 fn invalid_segment_error() {
-    let mut server = make_server();
+    let app = tide::App::new(());
```
tirr-c: We need to set up routers for *all* test functions, as functions can't see each other.

gruberb: I added the routes. I wonder if there is a better way of initializing `app` and add the routes, so every test `fn` can access it...

gruberb:
Ping @tirr-c Tests are passing

PR #146, opened 2019-03-06
Update to nightly-2019-02-27 and make it default

tirr-c:
This fixes CI breakage.

EDIT: Versions from `nightly-2019-03-03` cause ICE on tests. Rolled back to the latest version that works well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,8 @@
 language: rust
 rust:
-  - nightly
+  - nightly-2019-02-27
 
 before_script: |
-  rustup install nightly-2019-02-27
-  rustup default nightly-2019-02-27
   rustup component add rustfmt clippy
 script: |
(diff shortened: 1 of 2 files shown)


tirr-c:
```
error: internal compiler error: src/librustc_codegen_llvm/debuginfo/metadata.rs:661: debuginfo: unexpected type in type_metadata: <tide::head::Named<Number> as tide::extract::Extract<()>>::Fut

thread 'rustc' panicked at 'Box<Any>', src/librustc_errors/lib.rs:620:9
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
thread '<unnamed>' panicked at 'Metadata module not compiled?', src/libcore/option.rs:1038:5
error: aborting due to previous error
```

Well, this is unexpected.

PR #151, opened 2019-03-12
Use Arc<Store> instead of Store for lazy loading

Murakami-Int:
## Description

Change store's type used in `Router` and `Extract` to wrapped one, to load configurations after extraction.
This change affects to public trait.

## Motivation and Context

Sometimes we define services for SoC. They would be dependent on another services.
Tide has per-endpoint configuration, so every services should be able to read it.

One way for it is passing a store's reference in services' constructor recursively while extraction. This looks good, but if some services depend on same services, duplicate object will be created.
Another way is extract (partial) clone of store and share it across services. If the store is big, per-request clone will be high cost.

This change helps latter one. Instead of cloning a store directly, we can clone just shared reference with few overhead.

## How Has This Been Tested?

There are no additional tests.
Implementations of `Extract` are affected, but it can be easily fixed.

## Types of changes

## Checklist:

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -86,7 +86,7 @@ impl<Data: Clone + Send + Sync + 'static> App<Data> {
             router: Router::new(),
             default_handler: EndpointData {
                 endpoint: BoxedEndpoint::new(async || http::status::StatusCode::NOT_FOUND),
-                store: Store::new(),
+                store: Arc::new(Store::new()),
             },
         };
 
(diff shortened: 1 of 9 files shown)


mmrath:
Why would services depend on Store? I would expect services to be independent of any of the types in `tide`.
Would you have some example code showing the benefit/usage?

Murakami-Int:
@mmrath Sorry, actualy I have no Idea.
At the first I thought is to inject slog logger and configure logging strategies by per-endpoint, per-service, but I can not determine whether this use case is appropriate because of lack of experiences.

aturon:
Closing this one, since the relevant concepts were removed in #156.

Issue #141, opened 2019-02-26
Consider using route_recognizer instead of path-table


tirr-c:
## Detailed Description
For now `Router` is using `path-table` for its route matching. Consider changing the algorithm by replacing it with [`route_recognizer`].

The published version of `route_recognizer` is unmaintained for about 2 years, and seems to have reversed precedence of path segment matching (conduit-rust/route-recognizer.rs#21). Maybe we can fork the repository and apply some patches.

[`route_recognizer`]: https://github.com/conduit-rust/route-recognizer.rs

## Context
Related to #12. `route_recognizer` seems to have more sophiscated route matching algorithm than `path-table`. I expect it to fix most of the counterintuitive cases.

krircc:
There is [actix_router](https://github.com/actix/actix-net/tree/master/router), but it is Independent，and dont dependent Actix* . It is General

aturon:
Note: this is the algorithm ultimately used by crates.io:

- https://github.com/conduit-rust/conduit-router
- https://github.com/rust-lang/crates.io/blob/master/Cargo.toml#L76

so it's "maintained", just hasn't needed a lot of updating. Definitely a worthy candidate to look at, and I  expect the current owners would be  interested in moving it into the rust-net-web org.

bIgBV:
One thing to consider for the router API is a way of providing reverse lookups #24. I'll go into more detail about the small survey I did of current implementation of such an API in django and flask, but both of them leverage the router to provide a reverse lookups of endpoints to paths for building URLs.

tirr-c:
So, we need a router that:
* has intuitive path matching algorithm
* can be nested
* has ability to perform reverse lookups

`path-table` currently used by Tide addresses only the second point, and both `route_recognizer` and `actix-router` don't have reverse lookup feature (as far as I understand). I think we need to fork and add needed functionality anyway, so how about experimenting with `route_recognizer`? It might be challenging with state machines, though.

fundon:
I wrote a https://github.com/trek-rs/path-tree.

* supports reverse lookups

* supports `named` and `catch-all` likes https://github.com/julienschmidt/httprouter, but more flexible.

Currently, I am thinking about how to generate urls. Maybe the `{}` is best syntax.

aturon:
[...]

PR #156, opened 2019-03-21
Revamp Tide, dropping Extractors and simplifying the framework

aturon:
This commit reworks Tide, with the goal of reducing the total number of concepts
in the framework. The key new idea is to remove the notion of `Extractor`s, which
in turn allows us to remove or simplify several other concepts in Tide.

We'll first lay out the new design, and then discuss the tradeoffs made in this
simplification.

Here's a full list of the concepts in Tide after this commit:

| Concept  | Description |
| ----- | ----------- |
| `App` | Builder for Tide applications |
| `Route` | Builder for an individual route |
| `Endpoint` | Trait for actual endpoints |
| `Context` | The request context for an endpoint |
| `IntoResponse` | A trait for converting into a `Response` |
| `Middleware` | A trait for Tide middleware |

Previously, the `Endpoint` trait was treated as a somewhat magical internal
abstraction, and we used a macro to provide `Endpoint` implementations for
actual endpoints (with varying numbers of extractor arguments).

In this commit, an `Endpoint` is just an asynchronous function from a `Context`
to a `Response`:

```rust
pub trait Endpoint<AppData>: Send + Sync + 'static {
    /// The async result of `call`.
    type Fut: Future<Output = Response> + Send + 'static;

    /// Invoke the endpoint.
    fn call(&self, cx: Context<AppData>) -> Self::Fut;
}
```

For convenience, this trait is implemented for async functions that return any
value that implements `IntoResponse`:

```rust
impl<AppData, F, Fut> Endpoint<AppData> for F
where
    F: Fn(Context<AppData>) -> Fut,
    Fut: Future
    Fut::Output: IntoResponse,
    // ...
```

This implementation is in contrast to the macro-generated implementations we
previously had, which allowed endpoints with varying numbers of `Extractor`
arguments. The intent is for endpoints to perform their own extraction directly
on the `Context`, as we'll see next.

The `Context` type contains all of the request and middleware context an
endpoint operates on. You can think of it as wrapping an `http_service::Request`
with some additional data.

It's easiest to understand `Context` through the APIs it provides. First, we have
methods for getting basic http request information, mirroring the `http` APIs:

```rust
impl<AppData> Context<AppData> {
    pub fn method(&self) -> &Method;
    pub fn uri(&self) -> &Uri;
    pub fn version(&self) -> Version;
    pub fn headers(&self) -> &HeaderMap;
}
```

The context also has a handle to application data, which typically would store
database connection pools and other "application-global" state. This API
replaces the old `AppData` extractor:

```rust
impl<AppData> Context<AppData> {
    pub fn app_data(&self) -> &AppData {
        &self.app_data
    }
}
```

Similarly, we provide a *direct* API for extracting any "route parameters"
(i.e. placeholders in the route URL), replacing the need for `NamedSegment` and
the like:

```rust
impl<AppData> Context<AppData> {
    pub fn route_param(&self, key: &str) -> Option<&str>;
}
```

Basic body extraction is likewise built in via `Context` methods, replacing the
`Str`, `Bytes`, and `Json` extractors:

```rust
impl<AppData> Context<AppData> {
    pub async fn body_bytes(&mut self) -> std::io::Result<Vec<u8>>;
    pub async fn body_string(&mut self) -> std::io::Result<String>;
    pub async fn body_json<T: serde::de::DeserializeOwned>(&mut self) -> std::io::Result<T>;
}
```

Looking at the [message database example](https://github.com/rustasync/tide/blob/master/examples/messages.rs#L44),
we previously had endpoints like this:

```rust
async fn new_message(mut db: AppData<Database>, msg: body::Json<Message>) -> String {
    db.insert(msg.clone()).to_string()
}

async fn set_message(
    mut db: AppData<Database>,
    id: head::Path<usize>,
    msg: body::Json<Message>,
) -> Result<(), StatusCode> {
    if db.set(*id, msg.clone()) {
        Ok(())
    } else {
        Err(StatusCode::NOT_FOUND)
    }
}
```

These endpoints would now be written something like this (where `Error` is
intended as a general error type, convertible into a response):

```rust
async fn new_message(cx: Context<Database>) -> Result<String, Error> {
    let msg = await!(cx.body_json())?;

    cx.app_data().insert(msg).to_string()
}

async fn set_message(cx: Context<Database>) -> Result<(), Error> {
    let msg = await!(cx.body_json())?;

    if cx.app_data().set(cx.route_param("id"), msg) {
        Ok(())
    } else {
        Err(StatusCode::NOT_FOUND)
    }
}
```

The endpoint code is a bit more verbose, but also arguably easier to follow,
since the extraction (and error handling) is more clear.

In addition, the basic extraction approach is *more discoverable*, since it
operates via normal methods on `Context`.

Part of the idea of the old `Extractor` trait was that Tide would provide an
*extensible* system of extractors; you could always introduce new types that
implement `Extractor`. But now most of the existing extractors are built-in
`Context` methods. How do we recover extensibility?

Easy: we use Rust's ability to extend existing types with new methods, via
traits! (Note: this is directly inspired by the Gotham framework).

Let's say we want to provide cookie extraction. Previously, we'd have a `Cookies`
type that you could use as an endpoint argument for extraction. Now, instead, we
can introduce a `Cookies` *trait* that's used to extend `Context` with new APIs:

```rust
trait Cookies {
    fn cookies(&self) -> CookieJar;
}

impl<AppData> Cookies for Context<AppData> { ... }
```

This pattern is called an "extension trait" -- a trait whose sole purpose is to
extend an existing type with new methods. There are several nice properties of
this approach:

- The resulting extraction API is just a direct and natural as the built-in
  ones: just a method call on the `Context` object.

- The methods that are available on `Context` are controlled by what traits are
  in scope. In other words, if you want to use a custom extractor from the
  ecosystem, you just bring its trait into scope, and then the method is
  available. That makes it easy to build a robust ecosystem around a small set
  of core Tide APIs.

One of the major benefits of moving extraction into the endpoint body, rather
than via `Extractor` arguments, is that it's much simpler to provide
configuration. For example, we could easily provide a customized json body
extractor that limited the maximum size of the body or other such options:

```rust
impl<AppData> Context<AppData> {
    pub async fn body_json_cfg<T: serde::de::DeserializeOwned>(&mut self, cfg: JsonConfig) -> std::io::Result<T>;
}
```

As a result, we can drop much of the complexity in `App` around configuration.

Following the spirit of the changes to extractors, response generation for
non-standard Rust types is now just done via a free function:

```rust
mod response {
    pub fn json<T: serde::Serialize>(t: T) -> Response { ... }
}
```

As before, there's a top-level `App` type for building up a Tide application.
However, the API has been drastically simplified:

- It no longer provides a configuration system, since extractors can now be
  configured directly.
- It no longer allows for the middleware list to be customized per route;
  instead, middleware is set up only at the top level.

These simplifications make the programming model much easier to understand;
previously, there were inconsistencies between the way that middleware nesting
and configuration nesting worked. The hope is that we can get away with this
much simpler, top-level model.

When actually adding routes via `at`, you get a `Route` object (which used to be
`Resource`). This object now provides a *builder-style* API for adding
endpoints, allowing you to chain several endpoints. Altogether, this means we
can drop nested routing as well.

The middleware trait is more or less as it was before, adjusted to use `Context`
objects and otherwise slightly cleaned up.

This commit also switches to using the route-recognizer crate, rather than the
path-table crate, as the underlying routing mechanism. In addition to being more
efficient, route-recognizer provides a more intuitive semantics for "ambiguous"
routing situations. See issue #12 and issue #141 for more details.

# TODO

- [x] Update docs
- [x] Add back forms and cookies
- [x] Update tests
- [x] Update examples
- [ ] Support `App` nesting


bIgBV:
I really like makes the framework extremely approachable. The mental model for someone using the framework now is simple, as all they have to concern themselves with is how to get data from either the request or the context, and how to return a response.

I think this makes for an excellent base for third party extensions, I can already see how a generic session interface could look like.

I agree with @tirr-c , nesting routes is a very useful form of encapsulation. A possible idea for a future time would be a way to specify extension traits on a nested `App`/`Route`.

aturon:
@tirr-c agreed re: nesting!

To clarify a bit: before we were using nesting for a few different purposes, including customizing configuration and middleware for specific subroutes. I was wanting to simplify *that* story, but totally agree that we need some way to nest full apps. I'll get that added to this PR :+1:

aturon:
Update:

- Rebased
- Re-added cookie extraction
- Re-added `RootLogger` and `DefaultHeaders` middleware

src/context.rs
```diff
@@ -0,0 +1,87 @@
...
+#[derive(Clone)]
+pub struct Context<AppData> {
+    app_data: Arc<AppData>,
+    context_data: Arc<ContextData>,
+}
+
+struct ContextData {
+    meta: http::request::Parts,
+    body: Mutex<http_service::Body>,
+    extensions: Mutex<Extensions>,
```
tirr-c: I think `Context` isn't required to be `Send` nor `Sync` because only one thread will be accessing the struct during the lifetime (as each request is processed in its own thread.) Did I get it wrong?

src/context.rs
```diff
@@ -0,0 +1,87 @@
...
+        let body_bytes = await!(self.body_bytes())?;
+        Ok(serde_json::from_slice(&body_bytes).map_err(|_| std::io::ErrorKind::InvalidData)?)
+    }
+
+    pub fn replace_body(&mut self, body: Body) -> Body {
+        std::mem::replace(&mut self.context_data.body.lock().unwrap(), body)
+    }
+
+    pub fn insert_extension<T: Send + Sync + 'static>(&mut self, ext: T) -> Option<T> {
+        self.context_data.extensions.lock().unwrap().insert(ext)
```
tirr-c: *(worries about locking and unlocking mutexes often)*

src/context.rs
skade: I would not agree. It's certainly rare, but also not unusual to spawn further threads during a request execution that might have access to the request data.

aturon:
@tirr-c I'm also uneasy about the mutexes, but one note: this is currently in place to work around the compiler limitations mentioned in https://github.com/rustasync/tide/issues/3. In the long run, we'll be able to pass the endpoint `&mut Context`, which will allow middleware to inspect context modifications *after* the endpoint runs, without needing to clone and share state.

aturon:
Update: the PR is now at feature parity with master, including form extraction and response generation.

Next up: 

- docs, examples, and tests
- nested apps

After that, I think we will be ready to land based on stakeholder feedback so far!

src/context.rs
yoshuawuyts: I think this should be okay performance-wise. This seems like a synchronization point that needs to happen, and we can always optimize later in some form or shape.

In any case, once we have some examples we can load test + analyze them, and quickly figure out which points form the bottleneck.

src/app.rs
```diff
@@ -71,116 +62,99 @@ use crate::{
...
-        self.default_handler = endpoint;
-        &mut self.default_handler
-    }
-
-    /// Apply `middleware` to the whole app. Note that the order of nesting subrouters and applying
-    /// middleware matters; see `Router` for details.
-    pub fn middleware(&mut self, middleware: impl Middleware<Data> + 'static) -> &mut Self {
-        self.router.middleware(middleware);
-        self
-    }
```
chrisdickinson: Silly q: I was poking around trying to build an app with middleware against this branch (I'm excited for these changes!), but noticed that `app.middleware()` no longer works – was this method removed intentionally / is it coming back in a different form? (Or have I just jumped the gun entirely? 😅)

yoshuawuyts: @chrisdickinson Good question! -- I hadn't realized this was missing until now. Middleware is staying as a concept, so it should be coming back in some form. Probably more on this this week!

jesskfullwood:
I feel a bit uneasy commenting on this PR since I haven't taken an interest in `tide` until now and I may have misunderstood how it works, but here goes. I'm not in favor of removing the `Extractor`s. As far as I can see, it is functionally very similar to `actix-web`'s `FromRequest` trait. I maintain a moderately complex `actix-web` server and I have implemented `FromRequest` numerous times as it allows a form of declarative programming which makes it really explicit what each request handler is doing, just from the signature. I use it as a way of adding "capabilities" to request handlers.

For example, I have an extractor which looks at the request for an `Authorization` header, extracts the session id, validates it with Redis and returns an opaque type `AuthedUser` (which cannot be constructed any other way). Now, when I want to ensure that a user can only reach an endpoint when logged in, I just put `auth: AuthedUser` in the handler signature and it guarantees that the endpoint is secure. The `AuthedUser` is then required to be passed to any other 'secured' functions (e.g. database updates). I have granted certain capabilities to this endpoint via the signature.

It seems like with this PR equivalent would be extend `Context` to have an `Authorize` method which is called in the body. To me this is much less explicit (less _declarative_) than above.

Just my opinion, of course - I appreciate that this framework is taking a different path from `actix-web` and `rocket` (which has similar "request guards").

gruberb:
@jesskfullwood I actually had the same thought in mind. However, I have two signatures in mind which would read much nicer.

1. `app.at("/admin/users").verify(is_admin).get(list_admin_users)`
2. `app.at("/admin/users").get(verify, list_admin_users)`

I haven't thought about the implementation details yet since I am still building up enough knowledge, but I think this approach would be much cleaner then relying on the right parameter in `list_admin_users`.

bIgBV:
The thing is, the extraction of value from the `Context` is pretty repetitive, and I remember @aturon mentioning on discord that it'd be straightforward to provide a macro which extracts values out. This I think would bring the same level of declarativeness to the code (while also syntactically looking very similar to rocket), while still preserving the explicitness of understanding how it's doing it. Because right now, the way extractors work, it's really difficult for anyone new to understand how it works under the hood. I think it's valuable since one of the goals for tide was to provide a reference point for people wanting to build their own software using async/await.

I'm sure @aturon can clarify this more as I might be mis-remembering somethings.

tirr-c:
We might need to think about how we can pass "subpath" information to nested apps, if we make `App: Endpoint`. Mounted apps (being endpoints) cannot see where they are mounted on, so they need some info.

prasannavl:
Love these changes!

- What's the expected way to grab a peer socket address out of this? I don't see a straight forward way yet. Since this is a major rework, I think it would make sense to include a way to do that along. 

Current with Hyper, we can pull it out with `MakeServiceFn`. I suppose one way is to add it as `extension` to request, and be pull it back out. But I feel it would far better to expose it in Tide since `tide`.

aturon:
Yes, @bIgBV, that's exactly how I'm thinking about it! Right now the focus is to settle on the "core" APIs, and we'll have to continue to experiment to see whether we can make those ergonomic enough to use directly, or whether we want some (macro or trait-based) layer on top.

**Update**: I've pushed up code that gets the test suite and examples working, and adds a bunch of helpful error infrastructure (which makes endpoints quite a bit nicer to write).

**At this point the code is ready for a final review**, while I work on rewriting the docs :-)

prasannavl:
@aturon Did you give any thought to https://github.com/rustasync/tide/pull/156#issuecomment-479545253? 

I was thinking - one way would be to expose it in the `Context`. (With the actually `peer_addr` to be extracted out in the `http-service` layer, and possibly just shoved into the `Request` as an extension that can be pulled back out and exposed via the Context)

mmrath:
@aturon I was wondering how can we use context in middleware. If in routes/middleware some changes are done, how we can get the changes on to response. e.g. for the following middleware handler function we need a way to get some details on to response.

```rust
fn handle<'a>(&'a self, cx: Context<Data>, next: Next<'a, Data>) -> FutureObj<'a, Response> {
        box_async! {
            // Some changes to context here
            let mut res = await!(next.run(cx)); //Context is moved here
            // How to extract some bits from context to response here?
        }
}
```

crlf0710:
There's an unfortunate name clash with this `Context` and the std::task::Context, i think :( Hope it won't become too confusing.

yoshuawuyts:
@prasannavl could you perhaps open an issue about that? I agree we should have a way to do this, and like the approach you're proposing.

However it probably works best to make it a follow-up to this PR, as it can be added without any breaking changes to Tide (:

prasannavl:
@yoshuawuyts - Sure, makes sense. Will do that in a bit. Just wanted to make sure it was in @aturon's radar, in case he had something else in mind through the redesign. 

@aturon - Just went through the PR -- just thinking out loud here - could `Context` also be generic over per `RequestData`? It could be analogous to `AppData` which is for app level data - `RequestData` could be one that's created once per request. This could include structures that have request id, auth data, etc in a more structured format than using traits and `extensions` and as such can be shared without a Mutex that extensions today are behind. This could also open up the option of middlewares that can be specialised for certain RequestData (this could debated as both an advantage and disadvantage). 

I haven't given much thought into it, it could potentially make the ergonomics much worse too. But just wanted to spend a tiny bit exploring it.

prasannavl:
Should `IntoReponse` really be `TryIntoResponse`, and have it return `Result`? That gives the option to propagate some errors like de-serialisation and other finalizations into a final handler gracefully, instead of panic, or bogus `Response` that becomes tedious to examine?

src/response.rs
```diff
@@ -1,11 +1,15 @@
...
 
-use crate::body;
-
-/// An HTTP response.
-///
-/// A convenient alias for the `http::Response` type, using Tide's `Body`.
-pub type Response = http::Response<Body>;
+pub type Response = http_service::Response;
+
+pub fn json<T: serde::Serialize>(t: T) -> Response {
```
prasannavl: Perhaps return a `Result` to be more rust idiomatic? https://github.com/rustasync/tide/pull/156#issuecomment-480596914

src/response.rs
```diff
@@ -1,11 +1,15 @@
...
+    // TODO: remove the `unwrap`
+    http::Response::builder()
+        .status(http::status::StatusCode::OK)
+        .header("Content-Type", "application/json")
+        .body(Body::from(serde_json::to_vec(&t).unwrap()))
+        .unwrap()
+}
 
 /// A value that is synchronously convertable into a `Response`.
 pub trait IntoResponse: Send + Sized {
```
prasannavl: Please consider `TryIntoResponse` instead.  https://github.com/rustasync/tide/pull/156#issuecomment-480596914

aturon:
**Update**: 

- No more mutexes!
- Docs rewritten
- Rebased

Once we pass Travis I think we're good to go here!

@prasannavl 

> What's the expected way to grab a peer socket address out of this?

> I was thinking - one way would be to expose it in the Context.

Yep, agreed! Let's tackle that in follow-up.

> could Context also be generic over per RequestData?

Yes, but the primary use-case here is for communicating with middleware -- and *all* the middleware need to agree on the context type in use. Using extensions makes it easy to do this, since it is an extensible type map.

> Should IntoReponse really be TryIntoResponse, and have it return Result? 

That's a really interesting question! I don't want to delay this PR further, though, so would you mind opening up an issue about it where folks can discuss?

yoshuawuyts:
We already have an issue around error handling which I think overlaps with whether we should return `Result<TryInto<Response>, _>`. I suggest we continue our conversation there: https://github.com/rustasync/tide/issues/138.

cc/ @prasannavl

aturon:
OK, we're ready to merge! @tirr-c @yoshuawuyts @bIgBV, any concerns before we do so?

src/context.rs
```diff
@@ -0,0 +1,146 @@
...
+        self.request.headers()
+    }
+
+    /// Access the entire request.
+    pub fn request(&self) -> &http_service::Request {
+        &self.request
+    }
+
+    /// Mutably access the request body
+    pub fn body(&mut self) -> &mut http_service::Body {
```
bIgBV: I've generally seen two methods for mutable and immutable access. Any particular reason why we have a single function returning a mutable reference? I ask because I figure the body will only be read more often than not.

src/context.rs
```diff
@@ -0,0 +1,146 @@
...
+    ///
+    /// # Errors
+    ///
+    /// Yields an `Err` if the parameter was found but failed to parse as an
+    /// instance of type `T`.
+    ///
+    /// # Panics
+    ///
+    /// Panic if `key` is not a parameter for the route.
+    pub fn param<T: FromStr>(&self, key: &str) -> Result<T, T::Err> {
```
bIgBV: Generics are so cool :D

src/cookies.rs
```diff
@@ -1,44 +1,37 @@
...
-        future::ready(resp)
+impl<AppData> ExtractCookies for Context<AppData> {
+    fn cookie(&mut self, name: &str) -> Option<Cookie<'static>> {
+        let cookie_data = self
+            .extensions_mut()
+            .remove()
+            .unwrap_or_else(|| CookieData {
+                content: self
+                    .headers()
+                    .get("Cookie")
```
bIgBV: Maybe a more descriptive name like `tide-cookie`? Just a thought.

src/endpoint.rs
```diff
@@ -20,154 +22,53 @@ use crate::{
...
-        }
-    };
-}
-
-macro_rules! end_point_impl {
-    ($($X:ident),*) => {
-        end_point_impl_raw!([Head] $($X),*);
-        end_point_impl_raw!($($X),*);
+    fn poll(self: Pin<&mut Self>, waker: &std::task::Waker) -> std::task::Poll<Response> {
+        let inner = unsafe { self.map_unchecked_mut(|wrapper| &mut wrapper.fut) };
```
bIgBV: Maybe a comment explaining why we need `unsafe` here?

src/context.rs
aturon: The reason this one is mutable is because, since the body is a stream, to read it you *must* mutate it.

But more generally you're right that this API design doesn't quite smell right. And in practice, you end up using `take_body` anyway! So I think I'll drop this method for now.

src/endpoint.rs
aturon: Better yet -- nixed the unsafe :)

aturon:
All comments addressed, merging! Thanks everybody for your patience on this one!

examples/graphql.rs
```diff
@@ -3,26 +3,26 @@
...
 use std::sync::{atomic, Arc};
-use tide::{body, App, AppData, IntoResponse, Response};
+use tide::{error::ResultExt, response, App, Context, EndpointResult};
 
-// First, we define `Context` that holds accumulator state. This is accessible as App data in
+// First, we define `Data` that holds accumulator state. This is accessible as App data in
 // Tide, and as executor context in Juniper.
 #[derive(Clone, Default)]
-struct Context(Arc<atomic::AtomicIsize>);
+struct Data(Arc<atomic::AtomicIsize>);
```
pickfire: `app_data` is already `Arc<AppData>`, do we still need `Arc` here?

Issue #162, opened 2019-04-10
RFC: isolating a stable core


aturon:
Tide is designed to be a highly modular, extensible framework, and #156 pushes us further in that direction.

There are a set of core concepts and interfaces:

- The core routing system (`App` and `Route`)
- The `Endpoint` interface, including `Context`
- The `Middleware` interface

Beyond that, we have a number of implementations and extensions:

- Various "extractors", which are now implemented as extension traits that work on `Context`. See [forms](https://github.com/rustasync/tide/blob/dc28fd5601d8b8d071de999e080f2c347c3fd61d/src/forms.rs#L12-L18) for example.

- Middleware implementations, such as [logging](https://github.com/rustasync/tide/blob/dc28fd5601d8b8d071de999e080f2c347c3fd61d/src/middleware/logger.rs).

- Response builders, like [this one](https://github.com/rustasync/tide/blob/master/src/forms.rs#L48-L55).

Tide's current design allow for a strong decoupling between the core interfaces and these various extensions, much like Flask. This raises some questions about how to approach development, as well as the UX for Tide.

## Proposal: follow the Flask model (for now)

Since Tide is explicitly intended to foster re-usable, modular components that can work with other frameworks, it makes a lot of sense to take full advantage of the decoupling above. That would follow in Flask's footsteps, where there's an extensive ecosystem of independent extensions to the tiny core framework.

I propose that we:

- Limit the `tide` crate itself to the core interfaces and APIs, and perhaps some *extremely common* extractors (the ones that are currently directly provided on `Context`, like `body_json`.

- Move all other extractors, middleware, and response builders into small, separate crates -- and generally encourage a spirit of experimentation with lots of small crates trying out competing approaches.

- Provide a unified "guide to Tide's ecosystem" as an entry point to all these extensions. This could be as simple as a list of links to crate docs.

Eventually, as we gain more experience, I imagine that some of these extensions will become *de facto* standard, and we might want to consider moving them into core. Likewise, I expect we'll want some "bundling" crates that set up a middleware stack for you, to make it very easy to get going with Tide. But these things can come later.

This approach would make it feasible to get the Tide core to 1.0 status relatively quickly, with the expectation that the APIs would remain stable for a while (let's say six months at least). Meanwhile, the various middlware crates etc can be versioned independently, freeing them to make breaking changes without any coordination.

[...]

PR #166, opened 2019-04-12
Re-export http crate

pbzweihander:
## Description

I added `pub use http;` in `lib.rs`

## Motivation and Context

Since `tide` heavily depends on `http` crate, it would be great if we can use `http` crate without adding to `Cargo.toml`.

## How Has This Been Tested?

I checked via `cargo doc` that the `http` crate is re-exported.

## Types of changes

## Checklist:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -40,3 +40,5 @@ pub use crate::{
     response::Response,
     route::Route,
 };
+
+pub use http;


secretfader:
You might want to consider `pub use http` as a better approach. In [Rust 2018](https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html) `extern crate` is unnecessary since dependencies are managed in `Cargo.toml`.

yoshuawuyts:
@pbzweihander how are you using the `http` crate? If there's anything in particular missing from what we're exposing it'd be great to know.

pbzweihander:
@yoshuawuyts Especially `StatusCode`. If I want to use `with_status` of `IntoResponse`, I have to import `StatusCode` from `http`

secretfader:
I also frequently import the `http` crate as well. Usually, I'm creating redirect responses or other functionality that I expect will eventually be implemented in Tide.

Until Tide includes conveniences for all possible responses, exporting `http` seems like a decent option.

yoshuawuyts:
> Until Tide includes conveniences for all possible responses, exporting http seems like a decent option.

We're thinking of how to reduce the Tide API to just the core parts (#162), so I'm a bit cautious around additions that we know we don't want to keep in the long term.

That said, it does feel like exporting `http` will almost always remain useful, even if we expose shorthands, so it doesn't seem unreasonable to include this as part of core.

PR #168, opened 2019-04-12
example: remove asterisk from catch_all route params

### LINKED ISSUE #167: example 'catch all' panicked

secretfader:
This PR removes the asterisk from the `catch_all` route params, resolving the issue raised in #167.


secretfader:
Follow up to this PR is #171, because I misunderstood the route-recognizer API.

PR #169, opened 2019-04-13
fix: catch all in 'catch_all'

### LINKED ISSUE #167: example 'catch all' panicked

DCjanus:
Fix #167 in the right way

In example 'catch_all', I do belive that we should catch 'http://127.0.0.1:8000/echo_path/foo' and 'http://127.0.0.1:8000/echo_path/foo/foo', which means, [this fix](https://github.com/rustasync/tide/pull/168) is Wrong

## Types of changes

## Checklist:

### DIFF
--- a/examples/catch_all.rs
+++ b/examples/catch_all.rs
@@ -3,12 +3,12 @@
 use tide::Context;
 
 async fn echo_path(cx: Context<()>) -> String {
-    let path: String = cx.param("path").unwrap();
+    let path: String = cx.param("path*").unwrap();
     format!("Your path is: {}", path)
 }
 
 fn main() {
     let mut app = tide::App::new(());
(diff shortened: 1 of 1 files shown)


secretfader:
The `panic` concern raised by your issue was resolved. What we've yet to determine is whether the behaviour you believe to be correct actually is, a question I consider unresolved.

#156 migrated Tide to a new routing library ([`route-recognizer`](https://github.com/conduit-rust/route-recognizer.rs). IMO, it's premature to declare the fix "wrong" until we know what to expect.

In fact, after some time with [`route-recognizer`](https://github.com/conduit-rust/route-recognizer.rs) this morning, I can confirm that it panics when provided with `:path*`, too. So the correct response is to change position of the star parameter, not roll back the changes I made yesterday.

DCjanus:
Sounds fair to me.

PR #170, opened 2019-04-13
Cookie Improvement

mmrath:
## Description
This change adds support for adding, removing cookies on top of currently supported retrieval.

### Breaking changes:
1. Added a new middleware. This middleware is required to be added for the features to work
2. `ExtractCookies` is renamed to `CookiesExt` because this is no longer just extraction feature

## Motivation and Context
Every web application needs to deal with cookies at some stage. Hopefully this approach is simple and ergonomic

## How Has This Been Tested?
Test cases written for simple cases. Examples included for basic features.

## Types of changes

## Checklist:


mmrath:
/ping @aturon @yoshuawuyts @bIgBV 
Ready for initial feedback

examples/cookies.rs
```diff
@@ -0,0 +1,28 @@
+#![feature(async_await, futures_api)]
+
+use cookie::Cookie;
+use tide::{cookies::CookiesExt, middleware::CookiesMiddleware, Context};
+
+/// Tide will use the the `Cookies`'s `Extract` implementation to build this parameter.
```
bIgBV: Now that there is no `Extract` trait, shouldn't this change? Something in the lines of:

```
/// Tide will use the `CookiesExt` extension trait to extract the cookie from the request.
```

bIgBV: Now that there is no `Extract` trait, shouldn't this change? Something in the lines of:

```
/// Tide will use the `CookiesExt` extension trait to extract the cookie from the request.
```

src/cookies.rs
```diff
@@ -1,46 +1,64 @@
...
+    pub(crate) content: Arc<RwLock<CookieJar>>,
+}
+
+impl CookieData {
+    pub fn from_headers(headers: &HeaderMap) -> Self {
+        CookieData {
+            content: Arc::new(RwLock::new(
+                headers
+                    .get(http::header::COOKIE)
+                    .and_then(|raw| parse_from_header(raw.to_str().unwrap()).ok())
```
bIgBV: I think we should terminate early and send an error response back here.

src/middleware/cookies.rs
```diff
@@ -0,0 +1,46 @@
...
+
+            cx.extensions_mut().insert(cookie_data);
+            let mut res = await!(next.run(cx));
+            let headers = res.headers_mut();
+            for cookie in cookie_jar.read().unwrap().delta() {
+                let hv = HeaderValue::from_str(cookie.encoded().to_string().as_str());
+                 if let Ok(val) = hv {
+                    headers.append(http::header::SET_COOKIE, val);
+                 };
+                 // TODO: raise error in case of Error?
```
bIgBV: As mentioned above, I think we should terminate early and send an error response.

examples/cookies.rs
mmrath: Correct! I will have to change much of the documentation.

src/cookies.rs
mmrath: Yes, mostly we should send an error response back early. I think `CookiesExt` methods should be changed to return a `Result`. Not sure we can send an error response(http response) from here.

src/middleware/cookies.rs
mmrath: Agree

mmrath:
I think I should change `CookiesExt` to return Result. And send an internal error response in case of middleware.

src/cookies.rs
bIgBV: Yeah ideally, the extension should return a `Result` and the tide middleware handler should return an error response if any of the middleware functions return an error. I think that was the case previously, but it might have changed now. I'll have to check.

mmrath:
Ready for merge. Build fail because of futures api change.

src/middleware/cookies.rs
```diff
@@ -7,6 +7,11 @@ use crate::{
     Context, Response,
 };
 
+/// `CookiesMiddleware middleware is required for `CookiesExt` implementation on `Context` to work.
```
yoshuawuyts: Could you add an empty line after this? That way we can have concise preview docs, but show more details when expanded.

src/middleware/cookies.rs
```diff
@@ -38,7 +43,7 @@ impl<Data: Send + Sync + 'static> Middleware<Data> for CookiesMiddleware {
                 if let Ok(val) = hv {
                     headers.append(http::header::SET_COOKIE, val);
                 } else {
-                    //TODO Log error here
+                    // It would be useful to log this error here.
```
yoshuawuyts: Could you keep the TODO here? Those can be super useful to search for in the codebase.

src/middleware/cookies.rs
mmrath: Done. Thanks.

src/middleware/cookies.rs
mmrath: Done. Thanks

secretfader:
LGTM. The only thing I'd say (and this is just commentary) is that we may want to come up with a better term for these extension traits. If `_Ext` is to be the standard (which I'm fine with), I should probably rename the trait in #175.

Then again, we can rename them both so long as it happens before a release, now that `0.1.0` is out in the wild.

src/cookies.rs
```diff
@@ -1,46 +1,93 @@
...
-        self.extensions_mut().insert(cookie_data);
+            .extensions()
+            .get::<CookieData>()
+            .ok_or_else(|| StringError(MIDDLEWARE_MISSING_MSG.to_owned()))?;
 
-        cookie
+        let arc_jar = cookie_data.content.clone();
+        let locked_jar = arc_jar
+            .read()
+            .map_err(|e| StringError(format!("Failed to get write lock: {}", e)))?;
```
aturon: should be "read lock"

src/cookies.rs
```diff
@@ -1,46 +1,93 @@
...
-                    .unwrap_or_default(),
-            });
-        let cookie = cookie_data.content.get(name).cloned();
-        self.extensions_mut().insert(cookie_data);
+            .extensions()
+            .get::<CookieData>()
+            .ok_or_else(|| StringError(MIDDLEWARE_MISSING_MSG.to_owned()))?;
 
-        cookie
+        let arc_jar = cookie_data.content.clone();
```
aturon: Is this clone required?

src/cookies.rs
```diff
@@ -1,46 +1,93 @@
...
+            .map_err(|e| StringError(format!("Failed to get write lock: {}", e)))?;
+        Ok(locked_jar.get(name).cloned())
+    }
+
+    fn set_cookie(&mut self, cookie: Cookie<'static>) -> Result<(), StringError> {
+        let cookie_data = self
+            .extensions()
+            .get::<CookieData>()
+            .ok_or_else(|| StringError(MIDDLEWARE_MISSING_MSG.to_owned()))?;
+        let jar = cookie_data.content.clone();
```
aturon: Is this clone required?

src/cookies.rs
```diff
@@ -1,46 +1,93 @@
...
+        Ok(())
+    }
+
+    fn remove_cookie(&mut self, cookie: Cookie<'static>) -> Result<(), StringError> {
+        let cookie_data = self
+            .extensions()
+            .get::<CookieData>()
+            .ok_or_else(|| StringError(MIDDLEWARE_MISSING_MSG.to_owned()))?;
+
+        let jar = cookie_data.content.clone();
```
aturon: Is this clone required?

aturon:
@secretfader Agreed re: needing a solid convention here. I've [opened an issue](https://github.com/rustasync/tide/issues/187).

aturon:
I'm gonna go ahead and merge and will fix the nits in follow-up.

Thanks so much @mmrath, this is great work!

PR #175, opened 2019-04-16
Add query string extension trait

secretfader:
This commit adds query string extraction via the `ExtractQuery` trait, implemented for `tide::Context`.

## Description
Added a module, `querystring`, which includes the `ExtractQuery` trait to help with validating and deserializing HTTP query parameters.

## Motivation and Context

Processing query parameters is a common concern of most HTTP frameworks.

## How Has This Been Tested?

Simulated requests to a Tide endpoint. This PR doesn't remove or alter existing functionality.

## Types of changes

## Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ slog = "2.4.1"
 slog-async = "2.3.0"
 slog-term = "2.4.0"
 typemap = "0.3.3"
+serde_urlencoded = "0.5.5"
 
 [dependencies.http-service-hyper]
 optional = true
--- a/src/querystring.rs
+++ b/src/querystring.rs
@@ -1,0 +1,80 @@
+use crate::{error::Error, Context};
+use http::StatusCode;
+use serde::Deserialize;
+
+/// An extension trait for `Context`, providing query string deserialization.
+pub trait ExtractQuery<'de> {
+    fn url_query<T: Deserialize<'de>>(&'de self) -> Result<T, Error>;
+}
+
+impl<'de, Data> ExtractQuery<'de> for Context<Data> {
+    #[inline]
+    fn url_query<T: Deserialize<'de>>(&'de self) -> Result<T, Error> {
+        let query = self.uri().query();
+
+        if query.is_none() {
+            return Err(Error::from(StatusCode::BAD_REQUEST));
+        }
+
+        Ok(serde_urlencoded::from_str(query.unwrap())
+            .map_err(|_| Error::from(StatusCode::BAD_REQUEST))?)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use futures::executor::block_on;
+    use http_service::Body;
+    use http_service_mock::make_server;
+    use serde_derive::Deserialize;
+
+    #[derive(Deserialize)]
+    struct Params {
+        msg: String,
+    }
+
+    async fn handler(cx: crate::Context<()>) -> Result<String, Error> {
+        let p = cx.url_query::<Params>()?;
+        Ok(p.msg)
+    }
+
+    fn app() -> crate::App<()> {
+        let mut app = crate::App::new(());
+        app.at("/").get(handler);
+        app
+    }
+
+    #[test]
+    fn successfully_deserialize_query() {
+        let app = app();
+        let mut server = make_server(app.into_http_service()).unwrap();
+        let req = http::Request::get("/?msg=Hello")
+            .body(Body::empty())
+            .unwrap();
+        let res = server.simulate(req).unwrap();
+        assert_eq!(res.status(), 200);
+        let body = block_on(res.into_body().into_vec()).unwrap();
+        assert_eq!(&*body, &*b"Hello");
+    }
+
+    #[test]
+    fn unsuccessfully_deserialize_query() {
+        let app = app();
+        let mut server = make_server(app.into_http_service()).unwrap();
+        let req = http::Request::get("/").body(Body::empty()).unwrap();
+        let res = server.simulate(req).unwrap();
+        assert_eq!(res.status(), 400);
+    }
+
+    #[test]
+    fn malformatted_query() {
+        let app = app();
+        let mut server = make_server(app.into_http_service()).unwrap();
+        let req = http::Request::get("/?error=should_fail")
+            .body(Body::empty())
+            .unwrap();
+        let res = server.simulate(req).unwrap();
+        assert_eq!(res.status(), 400);
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -31,6 +31,7 @@ pub mod cookies;
 mod endpoint;
 pub mod forms;
 pub mod middleware;
+pub mod querystring;
 pub mod response;
 mod route;
 mod router;


yoshuawuyts:
Heh, looks like you and @badboy had the same idea at the same time -- https://github.com/badboy/tide-query-params

badboy:
He, yeah. I saw this PR _after_ I had written my code, so I didn't bother fixing it up, adding tests and submitting it as a PR, instead just pushing the code undocumented (and unlicensed).

secretfader:
@yoshuawuyts Oops.

@badboy's implementation looks very similar, with a few minor differences:

1. I used `serde_urlencoded` over `serde_qs`,  which doesn't support nested parameters. I was concerned we'd run into [this issue](https://github.com/samscott89/serde_qs/issues/17), which is currently un-patched in the library (a [PR exists](), though).

2. I return `Result<_, tide::error::Error>` instead of `Option`, which means this PR returns an error if parameters fail to deserialize. I suppose users can convert to `Option` if needed, but I'm not entirely sure which method I prefer.

3. This PR uses explicit lifetimes instead of `serde::DeserializeOwned`. I'm unsure which is best in this scenario.

I'd like to have a convenient example to show in tonight's presentation, so that was my primary goal in writing this PR. I'm happy to defer, though.

src/querystring.rs
```diff
@@ -0,0 +1,70 @@
...
+impl<'de, Data> ExtractQuery<'de> for Context<Data> {
+    fn query<T: Deserialize<'de>>(&'de self) -> Result<T, Error> {
+        let querystr = self.uri().query();
+
+        if querystr.is_none() {
+            return Err(Error::from(StatusCode::BAD_REQUEST));
+        }
+
+        let query = serde_urlencoded::from_str(querystr.unwrap())
+            .map_err(|_| Error::from(StatusCode::INTERNAL_SERVER_ERROR))?;
```
yoshuawuyts: Shouldn't this be a 400 error if we fail to parse?

src/querystring.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    #[test]
+    fn unsuccessfully_deserialize_query() {
+        let app = app();
+        let mut server = make_server(app.into_http_service()).unwrap();
+        let req = http::Request::get("/").body(Body::empty()).unwrap();
+        let res = server.simulate(req).unwrap();
+        assert_eq!(res.status(), 400);
+    }
+}
```
yoshuawuyts: Could we perhaps also include a test for a malformed query string? E.g. passing `?error=should_fail` or something similar, where `msg=Hello` would be expected.

yoshuawuyts:
@secretfader I think what you just said makes a lot of sense. The decisions you've made made a lot of sense!

I left a few on the issue I'm not sure about, but otherwise this is looking :100:

src/querystring.rs
secretfader: Done. I should've caught this, but it's always good to have a reviewer. Thanks!

src/querystring.rs
secretfader: Agreed. I returned `BAD_REQUEST` earlier, not sure why I didn't here. It has been corrected.

secretfader:
@yoshuawuyts I'm seeing issues with the clippy lints that were just added (`#![forbid(rust_2018_idioms)]`) is erroring on the `#[derive(Deserialize)]` macro. Have you seen that before? That's the only thing preventing a successful CI run, I think.

Cargo.toml
```diff
@@ -28,6 +28,7 @@ slog = "2.4.1"
 slog-async = "2.3.0"
 slog-term = "2.4.0"
 typemap = "0.3.3"
+serde_urlencoded = "0.5.5"
```
niklasf: Could we use `serde_qs`? Maybe more powerful than nescessary, but we already depend on it for forms.

secretfader: See my response to this question [here](https://github.com/rustasync/tide/pull/175#issuecomment-484542425).

secretfader: IMO, it could make sense to move away from `serde_qs` for form handling, too, unless we can get a maintainer to help cover issues for that library. 

This isn't to say I don't want to use `serde_qs` -- I understand why the library exists, but I don't want to end up depending on a library that has fixes ready to go but are unapplied.

secretfader:
@aturon Do we want to include this before your re-org or after?

secretfader:
@prasannavl This PR seems to conflict with some of the recent changes to Clippy lints, notably `#![forbid(rust_2018_idioms)]`. It seems like we'd prefer to favor Rust 2018 idioms. Maybe I misunderstand the point of this lint. Can you assist?

prasannavl:
> This PR seems to conflict with some of the recent changes to Clippy lints, notably #![forbid(rust_2018_idioms)]. It seems like we'd prefer to favor Rust 2018 idioms. Maybe I misunderstand the point of this lint. Can you assist?

@secretfader - I haven't look at the PR yet, but as answering your question quickly - `rust_2018_idioms` is the name of the lint. It makes sure you use the 2018 idioms. `forbid` is to indicate that the linter/compiler forbids you from NOT using 2018 idioms. :)

prasannavl:
Alright, so this seems to be since `serde_derive` is still a bit problematic with 2018.  https://github.com/serde-rs/serde/blob/master/serde_derive/src/dummy.rs#L26. This wasn't a problem before since Tide didn't need to use a derive so far. 

@yoshuawuyts - Should we drop the compiler lint down to `deny` to allow working with serde? (I personally haven't been able to find a workaround so far if serde derive is needed)

secretfader:
Serde [also has changed usage guidelines](https://github.com/serde-rs/serde/issues/1441) when compiling with Rust 2018. So I tried what the issue suggested: removing the explicit dependency  on `serde_derive` and enabling the `derive` feature on Serde itself.

While it makes imports cleaner, it did nothing to satisfy the lint in question.

src/querystring.rs
```diff
@@ -0,0 +1,79 @@
+use crate::{error::Error, Context};
+use http::StatusCode;
+use serde::Deserialize;
```
prasannavl: Could you perhaps, switch this to `serde::de::Deserialize` instead? (https://github.com/serde-rs/serde/issues/1441)

src/querystring.rs
```diff
@@ -0,0 +1,79 @@
...
+
+/// An extension trait for `Context`, providing query string deserialization.
+pub trait ExtractQuery<'de> {
+    fn query<T: Deserialize<'de>>(&'de self) -> Result<T, Error>;
```
prasannavl: Should this be `query_string` instead? I would find it confusing, in the land of web, and database queries.

src/querystring.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+impl<'de, Data> ExtractQuery<'de> for Context<Data> {
+    fn query<T: Deserialize<'de>>(&'de self) -> Result<T, Error> {
```
prasannavl: Same as above.

src/querystring.rs
secretfader: I opted for `query` because I was concerned about the opposite: that users would expect Rust methods ending in  `_string` methods might return `String`.

I'd like to hear more feedback on this question, though, because I don't mind either way.

prasannavl: I don't really think that's the rust convention. But since you mentioned, I just ran this on the crates registry, just to see for myself. 

`find . -type f -name "*.rs" -exec grep "_string(" /dev/null {} \;`

Unless the prefix part of the `string` has something that indicates `to_`/`as_`/`gen_`, or any other explicit notion that it returns a `String`, I don't think there was any. Languages like Go use that convention, but I doubt rust does.

prasannavl: Actually, I'm curious now. Let's invoke the big guns - @aturon, thoughts? (also if I'm in the vicinity of being right, that just brings about more questions - `body_as_string`/`body_as_json` etc?

src/querystring.rs
secretfader: This already matches the preferred import style of Serde in Rust 2018.

secretfader:
@prasannavl I have the Serde 2018 update ready to PR. I'll push it momentarily.

src/querystring.rs
secretfader: See response above. `serde::Deserialize` is preferred over `serde::de::Deserialize` in Rust 2018 imports.

src/querystring.rs
prasannavl: uh, really? I'm curious to know where you found this. Perhaps you're confusing the macro from the trait? 

Everywhere in tide, @aturon seem to have used it explicitly: https://github.com/rustasync/tide/blob/master/src/forms.rs#L14

Everywhere in serde docs: it's used explicitly: 
https://serde.rs/impl-serialize.html

I'm confused where you got this notion from. (Anyway, while this was just a suggestion - but I'd appreciate it if you don't just mark things as resolved in a whim without either proper rationale or providing a chance to respond)

secretfader: [`Deserialize` is exported at the crate root.](https://github.com/serde-rs/serde/blob/master/serde/src/lib.rs#L231)

It's my understanding that when the `derive` feature is implemented, it also become a valid import for usage as `#[derive(Deserialize)]`. Tests pass, which seem to confirm this understanding.

It makes sense that Serde would use the full path in their docs, because it's technical writing and they need to be exact. But `Deserialize` is also an often-used API, which is why it's exported at the crate root: as a convenience for users. In our case, I'd suggest it pays dividends to think like the user (since that's essentially what we are in this context) and unify import paths where reasonable.

src/querystring.rs
yoshuawuyts: Agree this might be confusing -- "query string" might be indeed interpreted as: "returns a string". Without reading the code I wouldn't know what type it returns.

Also "query" might be confusing the moment we start hooking up databases to context. One could reasonably expect that the `query` method allows you to construct a database query.

I'm seeing a few other naming options:
- `url_query`, similar to [`UrlQuery`](https://docs.rs/url/1.7.2/url/struct.UrlQuery.html), making it clear it's the query portion of the URL
- `search` or `search_params` as in url search params, which [is used by node.js](https://nodejs.org/api/url.html#url_url_search) and part of [the WhatWG url spec](https://url.spec.whatwg.org/#url-class), but is less used and doesn't have precedent in the `url` crate.

secretfader: +1 for @yoshuawuyts `url_query`

secretfader:
Rebased this branch against master and made the API changes suggested by @yoshuawuyts. `url_query` is an ideal name for this method.

src/querystring.rs
```diff
@@ -0,0 +1,79 @@
...
+use http::StatusCode;
+use serde::Deserialize;
+
+/// An extension trait for `Context`, providing query string deserialization.
+pub trait ExtractQuery<'de> {
+    fn url_query<T: Deserialize<'de>>(&'de self) -> Result<T, Error>;
+}
+
+impl<'de, Data> ExtractQuery<'de> for Context<Data> {
+    fn url_query<T: Deserialize<'de>>(&'de self) -> Result<T, Error> {
```
yoshuawuyts: I think we could mark this function as `#[inline]` without a problem. Though we should probably do a full pass over Tide to inline functions where possible.

secretfader: I'm not entirely sure what that attribute does. Do you think it's something I should add now or wait for a later, crate-wide audit?

secretfader: No sooner had I asked than I found this. https://doc.rust-lang.org/nightly/reference/attributes/codegen.html#the-inline-attribute

It seems like a good idea to mark this as inline.  I'll try it when I get back to the computer.

secretfader: No sooner had I asked than I found this. https://doc.rust-lang.org/nightly/reference/attributes/codegen.html#the-inline-attribute

It seems like a good idea to mark this as inline.  I'll try it when I get back to the computer.

secretfader: `#[inline]` has been added.

secretfader:
I think I've resolved all feedback. This seems ready to go in, but I'd prefer to let someone else push the "merge" button.

aturon:
Merged, thank you!

PR #179, opened 2019-04-21
add test that checks for empty body on HEAD req

fairingrey:
Attempts to address https://github.com/rustasync/tide/issues/45.

## Description

This adds a test that checks if a response with an empty body is returned on a HEAD request to a resource where GET behavior exists.

For some reason it fails, so there might be something in `http_service_mock` that needs to be fixed? Attempting an HTTP request to a live server via `curl`/Insomnia exhibits the normal behavior e.g. https://github.com/rustasync/tide/issues/45#issuecomment-439771815, but a non-empty body is returned on `server.simulate(req).unwrap()`, as if it was making a GET request normally.

EDIT: See my comment below https://github.com/rustasync/tide/pull/179#issuecomment-485635546

## Types of changes

## Checklist:


fairingrey:
Looking at it more in detail, I think I know why this doesn't work. Also related to https://github.com/rustasync/tide/issues/45#issuecomment-442195043.

My intuition was somehow that the http service and http service mock crates were using Hyper underneath -- this is wrong. Instead, the app server just has an implementation for `HttpService` via https://github.com/rustasync/tide/blob/master/src/app.rs#L246-L278 which http-service-mock emulates, but doesn't properly simulate the behavior that http-service-hyper exhibits when the app is probably served.

I'll probably start figuring out more on how to do this, but hmm...

tests/head_response_empty_body.rs
```diff
@@ -0,0 +1,23 @@
+#![feature(futures_api, async_await)]
+
+use futures::executor::block_on;
+use http_service::Body;
+use http_service_mock::make_server;
+use tide::{error::ResultExt, Context};
+
+async fn ok(_cx: Context<()>) -> String {
+    String::from("this shouldn't exist in the body of a HEAD response")
```
yoshuawuyts: I think if we want to make this a persistent test, just having this panic might be reasonable to signify it should never be executed.

yoshuawuyts:
@fairingrey hmm, not sure either. I guess the obvious choice here would be to expand http-service-mock to include this behavior. But that does kind of feel like a band-aid.

Hmm, we could also fix this at the tide-level, where if the request is a HEAD we abort early?

Not sure which is better here.

tests/head_response_empty_body.rs
fairingrey: We could do that... Maybe something like this, I'm guessing?
```suggestion
    panic!("should not go here");
    String::from("this shouldn't exist in the body of a HEAD response")
```

yoshuawuyts:
Ping @fairingrey -- do you have an idea how to move forward here?

fairingrey:
Mmm... I think I would rather just change http-service-mock to fit the test here. It is a bandaid, admittedly -- but since we already know how hyper behaves, it should be fine.

That said, I do think it won't necessitate the panic as you suggested originally here: https://github.com/rustasync/tide/pull/179#discussion_r277760925

PR #180, opened 2019-04-21
Migrate to serde_urlencoded, and away from serde_qs

secretfader:
As mentioned in #175, `serde_qs` isn't nearly as well maintained as `serde_urlencoded`. The only difference between the two libraries is the former's support for nested fields (a feature we can implement if need be).

So we don't have two different libraries doing nearly the same thing, I think it makes sense to use `serde_urlencoded` in forms if we're also using it for query string extraction (#175). 

## Description

Substitute `serde_urlencoded` for `serde_qs`. For uses contained in this PR, the API is identical. 

## Motivation and Context

Push Tide towards a single library for handling serialization of URL-encoded data.

## How Has This Been Tested?

Existing tests pass, and the `multipart-form` example was evaluated.

## Types of changes

## Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,11 +23,11 @@ route-recognizer = "0.1.12"
 serde = "1.0.90"
 serde_derive = "1.0.90"
 serde_json = "1.0.39"
-serde_qs = "0.4.5"
 slog = "2.4.1"
 slog-async = "2.3.0"
 slog-term = "2.4.0"
 typemap = "0.3.3"
+serde_urlencoded = "0.5.5"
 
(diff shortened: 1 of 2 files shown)


secretfader:
This PR is blocked on a `cargo fmt` error in CI. I'll correct it shortly.

secretfader:
Thanks, @yoshuawuyts. I'm still feeling out processes of who needs to approve and when, so while this feels like a simple PR to merge, I'd prefer if someone else (other than myself) did it.

aturon:
Merged, thanks!

PR #181, opened 2019-04-21
fix: error on example code

sangheestyle:
## Description

The example code on README.md got an error due to `?`.

## How Has This Been Tested?

Tested on `rustc 1.36.0-nightly (8aaae4294 2019-04-19)`

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,10 +59,10 @@ __Hello World__
 ```rust
 #![feature(async_await)]
 
-fn main() -> Result<(), failure::Error> {
+fn main() -> Result<(), std::io::Error> {
     let mut app = tide::App::new(());
     app.at("/").get(async move |_| "Hello, world!");
-    app.serve("127.0.0.1:8000")?;
+    Ok(app.serve("127.0.0.1:8000")?)
 }
 ```
 


secretfader:
Thanks for the report, @sangheestyle! You're correct, example code in the `README` doesn't work as-is. I left feedback on your commit, which resolves the issue, but doesn't address the function's return type. Ideally, I'd like to correct the function's type signature and error handling rather than simply `unwrap()`ing the value (which may cause a `panic`).

README.md
```diff
@@ -62,7 +62,7 @@ __Hello World__
 fn main() -> Result<(), failure::Error> {
     let mut app = tide::App::new(());
     app.at("/").get(async move |_| "Hello, world!");
-    app.serve("127.0.0.1:8000")?;
+    app.serve("127.0.0.1:8000").unwrap();
```
secretfader: Instead of `unwrap`ing the value here (which would probably cause a `panic`), would you be open to the following solution?

```rust
  // ...body omitted
  Ok(app.serve("127.0.0.1:8000")?)
}
```

I've confirmed that this code works without including `failure` or other error handling crates, which was another issue in our `README` example.

sangheestyle: Ah! I needed to check it thoroughly. My bad :) 
Thank you for the kind explanation.

sangheestyle:
> Thanks for the report, @sangheestyle! You're correct, example code in the `README` doesn't work as-is. I left feedback on your commit, which resolves the issue, but doesn't address the function's return type. It's a minor issue, really, but I'd prefer to correct the function's type signature and error handling rather than simply `unwrap()`ing the value (which may cause a `panic`).

Ah! :) I just revised my PR.

secretfader:
On behalf of the Tide team, thanks for your PR. It's my goal to make the Tide ecosystem a welcoming place, especially for new contributors. If there's anything I can do to help as you investigate further, please reach out!

I'll get this merged now so other folks don't hit the same issue. :)

sangheestyle:
@secretfader It was an excellent experience. I will do my best!

PR #185, opened 2019-04-22
docs: add supported rust versions

prasannavl:
Closes https://github.com/rustasync/tide/issues/173

## Description
Update docs to include supported rust versions

## Motivation and Context

https://github.com/rustasync/tide/issues/173

## How Has This Been Tested?

N/A

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,10 +91,11 @@ it has the following specific breakpoints that align with std future API changes
 
 Tide | Rust
 --- | ---
+&le; v0.1.0 | &le; nightly-2019-04-07
 &ge; v0.1.1 | &ge; nightly-2019-04-08
-&le; v0.1.0 | &le; nightly-2019-04-07*
 
-**Note:** Due to the fact that these changes are in `std`, projects with dependencies that use conflicting versions of  `std::futures` will not build successfully.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -84,6 +84,19 @@ Read about the design here:
...
 
+### Supported Rust Versions
+
+Tide is built against the latest Rust nightly releases and as such, due to it's use of `std` futures,
+it has the following specific breakpoints that align with std future API changes:
+
+Tide | Rust
+--- | ---
+&le; v0.1.0 | &le; nightly-2019-04-07
+&ge; v0.1.1 | &ge; nightly-2019-04-08
```
yoshuawuyts: I'm liking this a lot, but because we're still actively developing + futures aren't stable yet, I think we could get away with only keeping a single number around. I guess I'm wondering how useful this will still be after, say, 10 releases.

Or do you think we should keep a habit of tracking these numbers?

prasannavl: :100 with you. I was actually thinking 2-3 releases up ahead, we can trim 0.1.0 to a small note instead - and then completely remove it having after a few more releases, after which this section will likely be just one or two lines.

PR #189, opened 2019-04-25
add App::default

yoshuawuyts:
## Description
Adds `Default` to `App`.

## Motivation and Context
This allows us to make building applications a bit less noisy, especially for folks new to Rust, by removing `(())` from the base examples.

This also allows us to more gradually introduce concepts in Tide, allowing us to leave `AppData` out for an initial introduction, and gradually ease people into using more features.

__current__
```rust
fn main() -> Result<(), Error> {
    app.at("/").get(async move |_| "Hello, world!");
    app.serve("127.0.0.1:8080")?;
```

__proposed__
```rust
fn main() -> Result<(), Error> {
    let mut app = tide::App::default();
    app.at("/").get(async move |_| "Hello, world!");
    app.serve("127.0.0.1:8080")?;
}
```

## How Has This Been Tested?

It compiles, which in this case is enough.

## Types of changes

## Checklist:

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -232,6 +232,12 @@ impl<AppData: Send + Sync + 'static> App<AppData> {
     }
 }
 
+impl Default for App<()> {
+    fn default() -> Self {
+        Self::new(())
+    }
+}
+
 /// An instantiated Tide server.
 ///
 /// This type is useful only in conjunction with the [`HttpService`] trait,


gruberb:
How about removing the `pub` from `pub fn new(data: AppData) -> App<AppData> {` to make it private? I think it's confusing to have both `App::new(())` and `App::default()` as an option. 

This would be a breaking change, but we wouldn't expose a method we don't want users to use.

yoshuawuyts:
@gruberb I feel like we're on a pretty similar page here. I agree we should only have two constructors. What this PR proposes is:

- `app::default()` creates an application without app data
- `app::new(app_data)` creates a new application with app data

But I guess another option would be to not use default, but instead add a new method directly and change new. Becoming:
- `app::new()` creates an application without app data
- `app::with_context(app_data)` creates an application with app data.

We then have the question whether `with_context` is the right name, because it doesn't really take [a `Context` argument](https://docs.rs/tide/0.1.1/tide/struct.App.html#method.new). So perhaps `with_data` might be more on point:

- `app::new()` creates an application without app data
- `app::with_data(app_data)` creates an application with app data.

---

I don't care too much about which version we end up going for. I think both default / changing the behavior of `new` are valid options. I think I might have a slight preference for the latter, but it requires breaking changes. Tho what I care most about is having a short-hand constructor so we can keep `app_data` out as a concept, at least for the first few examples.

gruberb:
Agree. I also don't care too much about the names, just wanted to raise that it might be confusing to have a "hidden" `new(())`. 

Is there a use case where I have a database but don't init/pass it with `App::new(DB)`? In case no, in my opinion, we can be a bit more strict and say "you either create an app without a database or with a database". 

Then we could have 
- `App::empty()` and `App::with_data()`
- `App::default()` and `App::with_data()` 
-`App::new()` and `App::new_with_db(DB)`

So many options. Don't want to start a whole discussion about that, since people will get used to either one of them. My only thing would be to already have the two versions ready with this PR.

yoshuawuyts:
> Is there a use case where I have a database but don't init/pass it with App::new(DB)?

There's lots of things that could go in here. To name a few:
- connection pools to auth services
- connections to credential stores
- connections to caches
- configuration, read from either disk or the network
- handles to 3rd party service providers
- database connection pools

With the point being that we don't want to box ourselves in too much, and kind of allow this to be something Tide users get to fill in to match their needs.

---

Thinking about this more, I'm thinking of perhaps indeed going with the last option, and proposing we move to `App::new()` and `App::with_data(app_data)`, as I think it provides the most clarity.

fairingrey:
I'm most in support of going with the option yoshua is proposing, with `App::new()` and `App::with_data(app_data)`. It seems most clear to me that way, and is easiest to read.

However, I might also recommend `App::with_state(app_data)` as that's what users coming from actix-web are more familiar with (related example [here](https://actix.rs/docs/databases/)), and it might be more fitting to call it as such since as stated in the App docs:

* Application state is user-defined, and is provided via the [`App:new`] function. The state is available as a shared reference to all app endpoints.

In contrast to `App::with_data`, which may sound a bit more unclear to unsuspecting users.

And as yoshua said above, lots of things could be shared between whatever struct/object in passed into the shared app state. One use case I often think of is having two different connection pools to complimentary databases (such as having r2d2 pools for postgres for long-term storage and redis as a caching mechanism). And this could interplay with a lot of services also -- as already mentioned, one could even maybe stick in an async reqwest client to validate requests through external services and the like.

yoshuawuyts:
@fairingrey thanks for sharing; I like what you're proposing a lot. It seems to me as if we've come to a point where we're talking about larger changes than just adding a single method. I think it might be good to perhaps file an RFC (or similar!) to create concensus before moving forward with an implementation.

So I'll go ahead and close this PR, and submit a document for review in a bit!

PR #191, opened 2019-04-26
add rfc template

yoshuawuyts:
## Description
This adds an RFC template to Tide, similar to [rustwasm's RFC template](https://github.com/rustwasm/rfcs).

## Motivation and Context

When making changes to an API it can be useful to get consensus on the design first in a structured manner before moving to implementation. The Rust project has proven that filing RFCs has been a a good strategy to discuss these kinds of changes, and we're seeing a similar process work well for [`rustwasm/gloo`](https://github.com/rustwasm/gloo/pull/76) and [`crossbeam`](https://github.com/crossbeam-rs/rfcs/tree/master/text).

I've chosen to add the RFC template directly to Tide so we can discuss changes in the same repo. Having a separate RFC directory could be another option, but I feel that could be something we can  move to at a later stage if we think that's desirable. The main goal for now is to provide us with a more structured way of proposing API changes, for which this should be the right way to go.

## How Has This Been Tested?

No tests needed.

## Types of changes

## Checklist:

### DIFF
--- a/rfcs/000-template.md
+++ b/rfcs/000-template.md
@@ -1,0 +1,66 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- Tracking Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+One paragraph explanation of the proposal.
+
+# Motivation
+[motivation]: #motivation
+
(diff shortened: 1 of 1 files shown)


xenith:
This has received no objections since being raised, going to go ahead and merge so we can build on top of it.

PR #192, opened 2019-04-26
app constructor RFC

yoshuawuyts:
[rendered](https://github.com/rustasync/tide/blob/rfc-app-new/rfcs/001-app-new.md)

This proposes changes to the `App` constructor, and renaming `AppData` to `State`.

Also in general keen to hear if this process works for people; my hope is that by starting by introducing RFCs for API changes we can split discussions about the problem space from discussions the implementation details.

Continuation of https://github.com/rustasync/tide/pull/189

### DIFF
--- a/rfcs/001-app-new.md
+++ b/rfcs/001-app-new.md
@@ -1,0 +1,89 @@
+- Start Date: 2019-04-26
+- RFC PR: (leave this empty)
+- Tracking Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+We propose to rename the `AppData` generic argument to `State`, and expose two constructors on `App`
+to create new Tide applications.
+
+# Motivation
+[motivation]: #motivation
+
+With most introductory examples, passing the `AppData` parameter around isn't necessary. That's why
+most examples we author use `App::new(())`.
+
+However this can be confusing for people new to Tide ("why is the use of that parameter?"), and even
+stranger for people new to Rust ("what does `(())` mean?"). So it would be useful to have a way of
+constructing new Tide applications without needing to start off by explaining what `AppData` does.
+
+Another thing worth thinking about here is that the `AppData` argument is rather verbose. It's also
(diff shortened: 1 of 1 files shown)


fairingrey:
Following up on my comment https://github.com/rustasync/tide/pull/189#issuecomment-487004611, I'm in favor of renaming `AppData` to `State`, as I think its' much clearer as to what it does -- store shared state between apps. In addition to that we'd probably want to rename `Context::app_data` to `Context::state` also (the function [here](https://docs.rs/tide/0.1.1/tide/struct.Context.html#method.app_data)).

Such a change would likely bring it more in line with the naming scheme Actix-web has went with (their `State` documentation is listed [here](https://actix.rs/api/actix-web/stable/actix_web/struct.State.html)). In my opinion this would also be a simpler API to grasp. In Actix-web one can use shared application in an extractor-like fashion by including it in their endpoint handler (via `State<App>`) *or* they can call [`state`](https://actix.rs/api/actix-web/stable/actix_web/struct.HttpRequest.html#method.state) on a request context like `HttpRequest<State>`.

Following that, I would be in favor of adopting constructors `App::new` (which takes no arguments and just creates the app with associated type `()`) and `App::with_state` which takes a shared state object. I understand this would make it very similar to how [`App` is used in Actix-web](https://actix.rs/api/actix-web/stable/actix_web/struct.App.html), but I think it makes a lot of sense!

k-nasa:
I think the changes that @fairingrey says are good.
If this RFC is merged, may I implement it?

yoshuawuyts:
I've included the note on `Context::{app_data, state}` by @fairingrey. This has been open well over a week, has been read by quite a few people, and there seem to be no objections -- so I'm going to go ahead and merge this.

@k-nasa definitely; happy to review any PR!

k-nasa:
I will work from now on. Thanks!

k-nasa:
Then, the rust version in this PR's CI seems to be old.
 It would be nice to rebase the master.

PR #194, opened 2019-04-28
Middleware-based compression and decompression

fairingrey:
Provides compression middleware for outgoing HTTP responses and decompression middleware for incoming HTTP requests.

## Description

These two middlewares do the following:

* Automatically handle incoming request decompression for `Content-Encoding` headers that contain one or more applications of `gzip`, `deflate`, `br`, or `zstd`.
* Encode outgoing responses with preferred encoding behavior for any incoming request with a valid `Accept-Encoding` header.
* Allow one to define compression levels for each compression algorithm available.

The decompression middleware actually depends on getting a mutable handle on the request from the context, hence why a new function was added [here](https://github.com/rustasync/tide/blob/b6bcbde42007bf3e2a7c45b38309431c0a918971/src/context.rs#L59-L61).

## Motivation and Context

This closes https://github.com/rustasync/tide/issues/26.

Tide currently has no way of handling encoded requests or handling outgoing request compression. This middleware is one approach to solve that problem, although there exist a multitude of approaches that may work. This middleware may also serve as a point of discussion for https://github.com/rustasync/team/issues/52.

## How Has This Been Tested?

Tests are written! They're not very thorough, but that's because it's only to test their functionality. More rigorous tests for each of the stream encoders/decoders are in the async-compression crate.

## Types of changes

## Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 members = [
     "tide",
+    "tide-compression",
     "examples",
 ]
 
--- a/tide-compression/Cargo.toml
+++ b/tide-compression/Cargo.toml
@@ -1,0 +1,30 @@
+[package]
+authors = [
+    "Tide Developers",
+]
+description = "Compression-related middleware for Tide"
+documentation = "https://docs.rs/tide-compression"
+keywords = ["tide", "web", "async", "middleware", "compression"]
+categories = ["network-programming", "compression", "asynchronous"]
+edition = "2018"
+license = "MIT OR Apache-2.0"
+name = "tide-compression"
+readme = "README.md"
+repository = "https://github.com/rustasync/tide"
+version = "0.1.0"
+
+[dependencies]
+tide = { path = "../tide" }
+accept-encoding = "0.2.0-alpha.2"
+bytes = "0.4.12"
+futures-preview = "0.3.0-alpha.16"
+http = "0.1"
+http-service = "0.2.0"
+
+[dependencies.async-compression]
+default-features = false
+features = ["stream", "gzip", "zlib", "brotli", "zstd"]
+version = "0.1.0-alpha.1"
+
+[dev-dependencies]
+http-service-mock = "0.2.0"
--- a/tide-compression/README.md
+++ b/tide-compression/README.md
@@ -1,0 +1,16 @@
+# tide-compression
+
+This crate provides compression-related middleware for Tide.
+
+## Examples
+
+Examples are in the `/examples` folder of this crate.
+
+__Simple Example__
+
+You can test the simple example by running `cargo run --example simple` while in this crate's directory, and then running either of the following commands:
+
+```console
+$ curl http://127.0.0.1:8000/ -v
+$ curl http://127.0.0.1:8000/echo -v -d "why hello there"
+```
--- a/tide-compression/README.md
+++ b/tide-compression/README.md
@@ -12,5 +12,5 @@ You can test the simple example by running `cargo run --example simple` while in
 
 ```console
 $ curl http://127.0.0.1:8000/ -v
-$ curl http://127.0.0.1:8000/echo -v -d "why hello there"
+$ echo 'why hello there' | gzip | curl -v --compressed -H 'Content-Encoding: gzip' 'http://127.0.0.1:8000/echo' --data-binary @-
 ```
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,8 @@ before_script: |
   rustup component add rustfmt clippy
 script: |
   cargo fmt --all -- --check &&
-  cargo clippy --all -- -D clippy::all &&
+  cargo clippy --all --all-features -- -D clippy::all &&
   cargo build --no-default-features --verbose &&
-  cargo build --all --verbose &&
-  cargo test  --all --verbose
+  cargo build --all --all-features --verbose &&
+  cargo test --all --all-features --verbose
 cache: cargo
--- a/tide-compression/examples/simple.rs
+++ b/tide-compression/examples/simple.rs
@@ -18,5 +18,5 @@ pub fn main() {
     app.at("/echo").post(echo_bytes);
     app.middleware(Compression::with_default(Encoding::Brotli));
     app.middleware(Decompression::new());
-    app.serve("127.0.0.1:8000").unwrap();
+    app.run("127.0.0.1:8000").unwrap();
 }
--- a/tide-compression/examples/simple.rs
+++ b/tide-compression/examples/simple.rs
@@ -1,0 +1,22 @@
+#![feature(async_await)]
+use tide::{App, Context};
+use tide_compression::{Compression, Decompression, Encoding};
+
+// Returns a portion of the lorem ipsum text.
+async fn lorem_ipsum(_cx: Context<()>) -> String {
+    String::from("Lorem ipsum dolor sit amet, consectetur adipiscing elit.")
+}
+
+// Echoes the request body in bytes.
+async fn echo_bytes(mut cx: Context<()>) -> Vec<u8> {
+    cx.body_bytes().await.unwrap()
+}
+
+pub fn main() {
+    let mut app = App::new();
+    app.at("/").get(lorem_ipsum);
+    app.at("/echo").post(echo_bytes);
+    app.middleware(Compression::with_default(Encoding::Brotli));
+    app.middleware(Decompression::new());
+    app.serve("127.0.0.1:8000").unwrap();
+}
--- a/tide-compression/src/lib.rs
+++ b/tide-compression/src/lib.rs
@@ -1,0 +1,423 @@
+#![cfg_attr(feature = "nightly", deny(missing_docs))]
+#![cfg_attr(feature = "nightly", feature(external_doc))]
+#![cfg_attr(feature = "nightly", doc(include = "../README.md"))]
+#![cfg_attr(test, deny(warnings))]
+#![feature(async_await)]
+#![deny(
+    nonstandard_style,
+    rust_2018_idioms,
+    future_incompatible,
+    missing_debug_implementations
+)]
+
+pub use accept_encoding::Encoding;
+use async_compression::stream;
+use futures::future::BoxFuture;
+use http::{header::CONTENT_ENCODING, status::StatusCode, HeaderMap};
+use http_service::{Body, Request};
+use tide::{
+    middleware::{Middleware, Next},
+    response::IntoResponse,
+    Context, Error, Response,
+};
+
+macro_rules! box_async {
+    {$($t:tt)*} => {
+        ::futures::future::FutureExt::boxed(async move { $($t)* })
+    };
+}
+
+/// Encode settings for the compression middleware.
+///
+/// This can be modified in the case that you want more control over the speed or quality of compression.
+///
+/// For more information on how to configure each of these settings, see the async-compression crate.
+#[derive(Debug)]
+pub struct EncodeSettings {
+    /// Settings for gzip compression.
+    pub gzip: async_compression::flate2::Compression,
+    /// Settings for deflate compression.
+    pub deflate: async_compression::flate2::Compression,
+    /// Settings for brotli compression. Ranges from 0-11. (default: `11`)
+    pub brotli: u32,
+    /// Settings for zstd compression. Ranges from 1-21. (default: `3`)
+    pub zstd: i32,
+}
+
+impl Default for EncodeSettings {
+    fn default() -> Self {
+        Self {
+            gzip: Default::default(),
+            deflate: Default::default(),
+            brotli: 11,
+            zstd: 3,
+        }
+    }
+}
+
+/// Middleware for automatically handling outgoing response compression.
+///
+/// This middleware currently supports HTTP compression using `gzip`, `deflate`, `br`, and `zstd`.
+#[derive(Debug)]
+pub struct Compression {
+    default_encoding: Encoding,
+    settings: EncodeSettings,
+}
+
+impl Default for Compression {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Compression {
+    /// Creates a new Compression middleware. The default encoding is [`Encoding::Identity`] (no encoding).
+    pub fn new() -> Self {
+        Self {
+            default_encoding: Encoding::Identity,
+            settings: Default::default(),
+        }
+    }
+
+    /// Creates a new Compression middleware with a provided default encoding.
+    ///
+    /// This encoding will be selected if the client has not set the `Accept-Encoding` header or `*` is set as the most preferred encoding.
+    pub fn with_default(default_encoding: Encoding) -> Self {
+        Self {
+            default_encoding,
+            settings: Default::default(),
+        }
+    }
+
+    /// Accesses a mutable handle to this middleware's [`EncodeSettings`].
+    ///
+    /// This will allow you to configure this middleware's settings.
+    pub fn settings_mut(&mut self) -> &mut EncodeSettings {
+        &mut self.settings
+    }
+
+    fn preferred_encoding(&self, headers: &HeaderMap) -> Result<Encoding, Error> {
+        let encoding = match accept_encoding::parse(headers) {
+            Ok(encoding) => encoding,
+            Err(_) => return Err(Error::from(StatusCode::BAD_REQUEST)),
+        };
+        Ok(encoding.unwrap_or(self.default_encoding))
+    }
+
+    /// Consumes the response and returns an encoded version of it.
+    fn encode(&self, mut res: Response, encoding: Encoding) -> Response {
+        if res.headers().get(CONTENT_ENCODING).is_some() || encoding == Encoding::Identity {
+            return res; // avoid double-encoding a given response
+        }
+        let body = std::mem::replace(res.body_mut(), Body::empty());
+        match encoding {
+            Encoding::Gzip => {
+                let stream = stream::GzipEncoder::new(body, self.settings.gzip);
+                *res.body_mut() = Body::from_stream(stream);
+            }
+            Encoding::Deflate => {
+                let stream = stream::ZlibEncoder::new(body, self.settings.deflate);
+                *res.body_mut() = Body::from_stream(stream);
+            }
+            Encoding::Brotli => {
+                let stream = stream::BrotliEncoder::new(body, self.settings.brotli);
+                *res.body_mut() = Body::from_stream(stream);
+            }
+            Encoding::Zstd => {
+                let stream = stream::ZstdEncoder::new(body, self.settings.zstd);
+                *res.body_mut() = Body::from_stream(stream);
+            }
+            Encoding::Identity => unreachable!(),
+        };
+        res.headers_mut()
+            .append(CONTENT_ENCODING, encoding.to_header_value());
+        res
+    }
+}
+
+impl<Data: Send + Sync + 'static> Middleware<Data> for Compression {
+    fn handle<'a>(&'a self, cx: Context<Data>, next: Next<'a, Data>) -> BoxFuture<'a, Response> {
+        box_async! {
+            let encoding = match self.preferred_encoding(cx.headers()) {
+                Ok(encoding) => encoding,
+                Err(e) => return e.into_response(),
+            };
+            let res = next.run(cx).await;
+            self.encode(res, encoding)
+        }
+    }
+}
+
+/// Middleware for handling incoming request decompression.
+///
+/// This middleware currently supports HTTP decompression under the `gzip`, `deflate`, `br`, and `zstd` algorithms.
+#[derive(Debug, Default)]
+pub struct Decompression {}
+
+impl Decompression {
+    /// Creates a new Decompression middleware.
+    pub fn new() -> Self {
+        Self {}
+    }
+
+    fn parse_encoding(s: &str) -> Result<Encoding, Error> {
+        match s {
+            "gzip" => Ok(Encoding::Gzip),
+            "deflate" => Ok(Encoding::Deflate),
+            "br" => Ok(Encoding::Brotli),
+            "zstd" => Ok(Encoding::Zstd),
+            "identity" => Ok(Encoding::Identity),
+            _ => Err(Error::from(StatusCode::UNSUPPORTED_MEDIA_TYPE)),
+        }
+    }
+
+    fn decode(&self, req: &mut Request) -> Result<(), Error> {
+        let encodings = if let Some(hval) = req.headers().get(CONTENT_ENCODING) {
+            let hval = match hval.to_str() {
+                Ok(hval) => hval,
+                Err(_) => return Err(Error::from(StatusCode::BAD_REQUEST)),
+            };
+            hval.split(',')
+                .map(str::trim)
+                .rev() // apply decodings in reverse order
+                .map(Decompression::parse_encoding)
+                .collect::<Result<Vec<Encoding>, Error>>()?
+        } else {
+            return Ok(());
+        };
+
+        for encoding in encodings {
+            match encoding {
+                Encoding::Gzip => {
+                    let body = std::mem::replace(req.body_mut(), Body::empty());
+                    let stream = stream::GzipDecoder::new(body);
+                    *req.body_mut() = Body::from_stream(stream);
+                }
+                Encoding::Deflate => {
+                    let body = std::mem::replace(req.body_mut(), Body::empty());
+                    let stream = stream::ZlibDecoder::new(body);
+                    *req.body_mut() = Body::from_stream(stream);
+                }
+                Encoding::Brotli => {
+                    let body = std::mem::replace(req.body_mut(), Body::empty());
+                    let stream = stream::BrotliDecoder::new(body);
+                    *req.body_mut() = Body::from_stream(stream);
+                }
+                Encoding::Zstd => {
+                    let body = std::mem::replace(req.body_mut(), Body::empty());
+                    let stream = stream::ZstdDecoder::new(body);
+                    *req.body_mut() = Body::from_stream(stream);
+                }
+                Encoding::Identity => (),
+            }
+        }
+
+        // strip the content-encoding header
+        req.headers_mut().remove(CONTENT_ENCODING).unwrap();
+
+        Ok(())
+    }
+}
+
+impl<Data: Send + Sync + 'static> Middleware<Data> for Decompression {
+    fn handle<'a>(
+        &'a self,
+        mut cx: Context<Data>,
+        next: Next<'a, Data>,
+    ) -> BoxFuture<'a, Response> {
+        box_async! {
+            match self.decode(cx.request_mut()) {
+                Ok(_) => (),
+                Err(e) => return e.into_response(),
+            };
+            next.run(cx).await
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use async_compression::flate2;
+    use bytes::Bytes;
+    use futures::{
+        executor::{block_on, block_on_stream},
+        stream::StreamExt,
+    };
+    use http::header::ACCEPT_ENCODING;
+    use http_service::Body;
+    use http_service_mock::make_server;
+
+    async fn lorem_ipsum(_cx: Context<()>) -> String {
+        String::from(r#"
+            Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam rutrum et risus sed egestas. Maecenas dapibus enim a posuere
+            semper. Cras venenatis et turpis quis aliquam. Suspendisse eget risus in libero tristique consectetur. Ut ut risus cursus, scelerisque
+            enim ac, tempus tellus. Vestibulum ac porta felis. Aenean fringilla posuere felis, in blandit enim tristique ut. Sed elementum iaculis
+            enim eu commodo.
+        "#)
+    }
+
+    fn lorem_ipsum_bytes() -> Vec<u8> {
+        String::from(r#"
+            Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam rutrum et risus sed egestas. Maecenas dapibus enim a posuere
+            semper. Cras venenatis et turpis quis aliquam. Suspendisse eget risus in libero tristique consectetur. Ut ut risus cursus, scelerisque
+            enim ac, tempus tellus. Vestibulum ac porta felis. Aenean fringilla posuere felis, in blandit enim tristique ut. Sed elementum iaculis
+            enim eu commodo.
+        "#).into_bytes()
+    }
+
+    // Echoes the request body in bytes.
+    async fn echo_bytes(mut cx: Context<()>) -> Vec<u8> {
+        cx.body_bytes().await.unwrap()
+    }
+
+    // Generates the app.
+    fn app() -> tide::App<()> {
+        let mut app = tide::App::new();
+        app.at("/").get(lorem_ipsum);
+        app.at("/echo").post(echo_bytes);
+        app.middleware(Compression::new());
+        app.middleware(Decompression::new());
+        app
+    }
+
+    // Generates a response given a string that represents the Accept-Encoding header value.
+    fn get_encoded_response(hval: &str) -> Response {
+        let app = app();
+        let mut server = make_server(app.into_http_service()).unwrap();
+        let req = http::Request::get("/")
+            .header(ACCEPT_ENCODING, hval)
+            .body(Body::empty())
+            .unwrap();
+        let res = server.simulate(req).unwrap();
+        res
+    }
+
+    // Generates a decoded response given a request body and the header value representing its encoding.
+    fn get_decoded_response(body: Body, hval: &str) -> Response {
+        let app = app();
+        let mut server = make_server(app.into_http_service()).unwrap();
+        let req = http::Request::post("/echo")
+            .header(CONTENT_ENCODING, hval)
+            .body(body)
+            .unwrap();
+        let res = server.simulate(req).unwrap();
+        res
+    }
+
+    #[test]
+    fn compressed_gzip_response() {
+        let res = get_encoded_response("gzip");
+        assert_eq!(res.status(), 200);
+        let body = res.into_body();
+        let stream = stream::GzipDecoder::new(body);
+        let decompressed_body: Vec<u8> = block_on_stream(stream)
+            .map(Result::unwrap)
+            .flatten()
+            .collect();
+        let lorem_ipsum = lorem_ipsum_bytes();
+        assert_eq!(decompressed_body, lorem_ipsum);
+    }
+
+    #[test]
+    fn compressed_deflate_response() {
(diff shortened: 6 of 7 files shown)


fairingrey:
The entire middleware here is feature-gated behind `"compression"` because it imports a number of new crates exclusive to itself. I can remove the feature gate and optional dependencies later if we decide to make it a core part of the framework, but as Tide is still very much in development and this being only one approach to the issue, I'd like to keep things lightweight.

Current implementation issues probably revolve around how to set a new body on a mutable request. As seen here: https://github.com/fairingrey/tide/blob/52d40d04ca5b45337753ad51d298576440ffbac9/src/middleware/compression.rs#L127-L139 I'm currently having a problem trying to figure out how one would allow a new body to bind to the response passed into `encode`. The `'static` lifetime bound in https://docs.rs/http-service/0.1.5/http_service/struct.Body.html#method.from_stream makes this especially difficult, but maybe my approach is flawed somewhere?

src/middleware/compression.rs
```diff
@@ -0,0 +1,229 @@
...
+        }
+    }
+    /// Consumes the response and returns an encoded version of it.
+    fn encode(&self, mut res: Response, encoding: Encoding) -> Response {
+        res.headers_mut()
+            .append(CONTENT_ENCODING, encoding.header_value());
+        match encoding {
+            Encoding::Gzip => {
+                // FIXME: currently does not encode anything yet
+                let mut body = res.body_mut();
```
DoumanAsh: use `mem::swap` to get body out

src/middleware/compression.rs
```diff
@@ -0,0 +1,229 @@
...
+                let stream = body.map(|b| match b {
+                    Ok(b) => {
+                        let reader = b.into_buf().reader();
+                        let mut buffer: Vec<u8> = Vec::new();
+                        let mut encoder = GzEncoder::new(reader, self.settings.gzip_level);
```
DoumanAsh: `GzEncoder` should be created once outside of closure, and then used

Also use `flate2::write::GzEncoder::new(Vec::new()...`

src/middleware/compression.rs
fairingrey: I've tried that, same problem.

src/middleware/compression.rs
```diff
@@ -0,0 +1,233 @@
...
+            }
+            Encoding::Deflate => {
+                let mut body = Body::empty();
+                std::mem::swap(&mut body, res.body_mut());
+                let level = self.settings.deflate_level;
+                let stream = body.map(move |b| match b {
+                    Ok(b) => {
+                        let reader = b.into_buf().reader();
+                        let mut buffer: Vec<u8> = Vec::new();
+                        let mut encoder = DeflateEncoder::new(reader, level);
```
Nemo157: Creating a new encoder for each chunk is not going to work, they implicitly create internal state as they’re encoding the data, and later chunks will depend on this state from earlier chunks. You will need to use the [low-level in memory chunk encoder](https://docs.rs/flate2/1.0.7/flate2/struct.Compress.html) and keep one instance around for the entire stream.

src/middleware/compression.rs
```diff
@@ -0,0 +1,233 @@
...
+            "deflate" => Ok(Encoding::Deflate),
+            "identity" => Ok(Encoding::Identity),
+            "br" => Ok(Encoding::Deflate),
+            "*" => Ok(self.default_encoding),
+            _ => Err(Error::from(StatusCode::UNSUPPORTED_MEDIA_TYPE)),
+        }
+    }
+
+    fn preferred_encoding(&self, headers: &HeaderMap) -> Result<Encoding, Error> {
+        let mut choices: Vec<(Encoding, f32)> = Vec::new();
```
DoumanAsh: No need for Vec here.

Algorithm should be following:
- Default choice is Identity with lowest priority.
- If header is missing, then treat it as `*` which means client has no preference. In this case it is either should be `Identity` or configured by user
- As you go through each value, save the highest priority to variable with default choice.

src/middleware/compression.rs
```diff
@@ -0,0 +1,233 @@
...
+                    Err(e) => Err(e),
+                });
+                *res.body_mut() = Body::from_stream(stream);
+                res
+            }
+            Encoding::Deflate => {
+                let mut body = Body::empty();
+                std::mem::swap(&mut body, res.body_mut());
+                let level = self.settings.deflate_level;
+                let stream = body.map(move |b| match b {
```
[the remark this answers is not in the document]
fairingrey: Sure thing, I'll use that.

src/middleware/compression.rs
fairingrey: Alright, noted. I may need help with this implementation as I am still pretty wet behind the ears when it comes to Stream wrappers. Will try my best though.

src/middleware/compression.rs
fairingrey: I'll fix this up since I kind of wrote it based on my first understanding of the spec.

src/middleware/compression.rs
Nemo157: You can ping me on either discord if you've got questions.

fairingrey: Will do, thank you!

src/middleware/compression.rs
```diff
@@ -0,0 +1,239 @@
+#![cfg(feature = "compression")]
+use brotli2::read::{BrotliDecoder, BrotliEncoder};
```
pickfire: I think we should use `brotli2::bufread` instead of `brotli2::read`, IIRC buffer based read/write are more efficient, not sure if it applies here.

fairingrey: Yeah, it should be -- I was just going off whatever encoders/decoders used `std::io::Read` as their interface (as Bytes exposes a `Read` handle through the `IntoBuf` trait. But this approach isn't likely to work as following another discussion [here](https://github.com/rustasync/tide/pull/194#discussion_r279182624).

For now I'm sort of back at the drawing board, as if I use the low-level in-memory chunk encoder/decoder that both [flate2](https://docs.rs/flate2/1.0.7/flate2/struct.Compress.html) and [brotli2](https://docs.rs/brotli2/0.3.2/brotli2/raw/struct.Compress.html) expose, I'll have to come up with some sort of strategy for decoding potentially large blocks and returning the stream of `Result<Bytes, Error>` piecemeal, unless the Stream generated by a non-compressed body already does this for me (then in which case I can read all the way to EOF, although this might be highly inefficient if I'm wrong).

src/middleware/compression.rs
```diff
@@ -0,0 +1,239 @@
...
+            Encoding::Identity => HeaderValue::from_str("identity").unwrap(),
+            Encoding::Br => HeaderValue::from_str("br").unwrap(),
+        }
+    }
+}
+#[derive(Debug, Default)]
+pub struct EncodeSettings {
+    gzip_level: flate2::Compression,
+    deflate_level: flate2::Compression,
+    brotli_level: u32,
```
pickfire: I am not sure if we should set a default of 0 for brotli, brotli_sys default to 11 in https://docs.rs/brotli-sys/0.3.2/brotli_sys/constant.BROTLI_DEFAULT_QUALITY.html

Maybe we should use CompressParams from https://docs.rs/brotli2/0.3.2/brotli2/struct.CompressParams.html and specify the quality. As seen in https://docs.rs/brotli2/0.3.2/src/brotli2/bufread.rs.html#44

src/middleware/compression.rs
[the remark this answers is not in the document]
pickfire: @Nemo157 does clippy lint these kinds of error?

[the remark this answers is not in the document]
Nemo157: Not currently. There are some similarish lints for iterator methods (e.g. [`iter_skip_next`](https://rust-lang.github.io/rust-clippy/master/index.html#iter_skip_next)), so maybe once combinators move into `std` they could have lints created for them. (Or if clippy had some form of plugin system maybe `futures` could have a plugin with lints specific for it?)

src/middleware/compression.rs
Nemo157: Looking at this, one thing I’ve been wondering is why `body` is a `Stream<Item = Bytes>` instead of an `AsyncRead`. If this is based on implementation details of `hyper` then maybe it would be better to change `http-service` to pass an `AsyncRead` (or more likely `AsyncBufRead`), then there can easily be an adaptor crate that implements just the compression adaptors on top of that which could be used here.

src/middleware/compression.rs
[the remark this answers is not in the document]
fairingrey: I've changed this in my latest push, since I imagine we have to work with the raw compressor now -- I'm just thinking about how one would be able to use a compressor instance that stays alive with the rest of the stream without the stream taking ownership of it...

fairingrey:
So just some thoughts on this if you'll allow me. This is going to be a long comment.

After messing around with possible approaches, I don't think this is going to work unless one consumes the entire stream of bytes into memory and then returns the entire encoded body in a stream that either yields it immediately in a single future or re-chunks it on the way back out through multiple futures each yielding some amount of bytes until it finishes. Both seem pretty inefficient.

I was talking with @Nemo157 in the Rust community discord about possible ways to do this, but other than the above, I don't think anything else is going to work. Each of the raw, in-memory compression handling objects for both [flate2](https://docs.rs/flate2/1.0.7/flate2/struct.Compress.html) and [brotli2](https://docs.rs/brotli2/0.3.2/brotli2/raw/struct.Compress.html) require they basically stay alive for the length that the stream is alive so it understands how to chunk each piece of memory given how it read all the previous chunks.

However that likely can't happen because as mentioned in our discussion, part of the issue is that I might not consume the entire incoming bytes without blocking the stream (thus making async kind of pointless) *and* if not that, I would be producing multiple output blocks for each input block (so streams within streams) which would then require `flat_map`... which then runs into lifetime/ownerships issues because the `Compress` objects _still_ have to be passed around since Stream is poll-based.

With all that considered, I don't think in its current state, Tide should be responsible for handling compression/decompression, as it may be handled more smoothly at either the hyper or http-service layer. But it does bring up a good point of discussion (and this may be related to changing/revamping what constitutes our response `Body`).

FWIW, it would be significantly easier to write an adapter that takes a streaming-like body that impl's `AsyncBufRead` and applies the necessary transforms to provide another body `AsyncRead`, or something similar to that approach. Trying to write a `Stream<Item = Bytes> -> Stream<Item = Bytes>` adapter to create a new streaming body efficiently seems impossible to me at the moment, especially given that the `Compress` object(s) depends on previous blocks to encode new blocks.

fairingrey:
Thanks to everyone here for helping review this PR, however, but I think this can speak for a larger, related discussion to what a tide response's body ought to look like. Although I do understand such a recommended change would necessitate newer breaking changes in related crates http-service and the like.

I'll try to bring it up during the next async WG meeting, and maybe we can discuss possible approaches there.

fairingrey:
Updated my comments to reflect my current status. Once https://github.com/rustasync/tide/pull/193 gets merged, I can start writing some tests for compressing Tide request bodies.

fairingrey:
One other concern I also had was if the request should also upsert the vary header with value `Accept-Encoding` as following https://developer.mozilla.org/en-US/docs/Web/HTTP/Compression#End-to-end_compression. Might be a necessary detail if we're going by what it says there.

fairingrey:
Phew... Finished all of the logic for compression and decompression handling over streams, but both implementations aren't tested yet because of current blockers over the outdated futures API on the latest nightly (deny warnings).

One other concern I had (and this is mentioned in my commit message) is that there is currently no way to get a `&mut Request` handle from the Context. This is another blocker since to decompress the request into plain data it's necessary to alter the incoming request being passed in.

I'll try to bring this concern up during the next meeting, but otherwise I think I've made a lot of progress! It would be good to get more review now while it's appropriate, I think.

kdar:
This could be due to my lack of experience with futures, but when I try to use this with hyper 0.12, I can only ever get it to spit out the first 8000 bytes. Even if I try to block and concat/collect the stream, it still only returns the first 8000 bytes.

Here is how I'm using it in hyper:
```rust
use std::io;
use async_compression::stream::deflate;
use futures::{future, stream::Stream};
use futures03::{compat::Stream01CompatExt, stream::TryStreamExt};

let (mut parts, mut body) = response.into_parts();
let compressed = deflate::DeflateStream::new(
  body
    .map(|v| v.into_bytes())
    .map_err(|e| io::Error::new(io::ErrorKind::Other, e))
    .compat(),
  flate2::Compression::default(),
);

body = Body::wrap_stream(compressed.compat());
let response = hyper::Response::from_parts(parts, body);
```

This ends up setting transfer-encoding to chunked, and only returning 8000 bytes.
I then tried this craziness to see if I could just grab everything from the stream and put it into a bytes::Bytes, but it still only returned 8000 bytes.

```rust
use futures03::executor::block_on;
use futures03::stream::StreamExt;
use hyper::body::Payload;
let payload: Vec<_> = block_on(compressed.map(|v| v.unwrap()).collect());
let mut fin: Vec<_> = vec![];
for b in payload {
  fin.put(b);
}
body = Body::from(fin);
```

I just want to confirm whether this is my issue or not.

fairingrey:
@kdar I'd recommend you open an issue in the [async-compression repo](https://github.com/rustasync/async-compression-rs/issues) and provide your code or steps to reproduce, and I can try helping you there. While this PR is sort of related, this is mostly for tracking progress on this middleware. But perhaps I can figure out a bug from what you're doing.

8000 bytes is the size of the internal buffer for those compression streams though, so if something weird is happening it's likely related to that.

src/middleware/compression.rs
```diff
@@ -0,0 +1,271 @@
...
+        let mut preferred_encoding = self.default_encoding;
+        let mut max_qval = 0.0;
+        'outer: for hval in headers.get_all(ACCEPT_ENCODING).iter() {
+            for cap in RE_ACCEPT_ENCODING.captures_iter(hval.to_str().unwrap()) {
+                match self.parse_encoding(cap.name("a").unwrap().as_str()) {
+                    Ok(encoding) => match cap.name("q") {
+                        Some(qval) => {
+                            let qval = match qval.as_str().parse::<f32>() {
+                                Ok(f) => f,
+                                Err(_) => {
```
pickfire:
```suggestion
                                Err(_) => continue,
```
Why not just inline it?

Cargo.toml
```diff
@@ -23,16 +24,29 @@ route-recognizer = "0.1.12"
...
+serde_urlencoded = "0.5.5"
 slog = "2.4.1"
 slog-async = "2.3.0"
 slog-term = "2.4.0"
 typemap = "0.3.3"
-serde_urlencoded = "0.5.5"
+
+[dependencies.async-compression]
+optional = true
+git  = "https://github.com/fairingrey/async-compression-rs"
```
pickfire:
```suggestion
git = "https://github.com/fairingrey/async-compression-rs"
```

src/middleware/compression.rs
fairingrey: Oh, good point. I'll also fix that along with the other review you left below.

Cargo.toml
```diff
@@ -13,6 +13,7 @@ repository = "https://github.com/rustasync/tide"
 version = "0.2.0"
 
 [dependencies]
+bytes = "0.4.12"
```
pickfire: I wonder what is this used for?

fairingrey: Ah, that can be removed safely. It was a remnant of some stuff I tried several days ago trying to write in the compression/decompression logic directly -- now all of it is in https://github.com/rustasync/async-compression-rs.

I'll go ahead and remove it real quick.

src/middleware/compression.rs
```diff
@@ -0,0 +1,281 @@
...
+        req.headers_mut().remove(CONTENT_ENCODING).unwrap();
+
+        Ok(())
+    }
+}
+
+impl<Data: Send + Sync + 'static> Middleware<Data> for Decompression {
+    fn handle<'a>(&'a self, cx: Context<Data>, next: Next<'a, Data>) -> BoxFuture<'a, Response> {
+        box_async! {
+//            match self.decode(cx.request()) { // <- There needs to be a way to grab a mut reference to the request
```
fairingrey: The logic is pretty much done for the decompression middleware, but this is a blocker.

Currently, there's no existing way to get a `&mut Request` handle from the Request context, and while there is a way to take the body through `take_body(&mut self)`, there's no way to set it through some existing method like `set_body(&mut self)` or so.

https://docs.rs/tide/0.2.0/tide/struct.Context.html#methods

fairingrey:
I'll be writing tests for the compression middleware in the meantime I'm looking to get the issue I'm having unblocked. Will likely start tomorrow, since I have some stuff to do today.

fairingrey:
I'll be moving some of my Accept-Encoding parsing logic to https://github.com/rustasync/accept-encoding under the recommendation of yoshua -- hopefully shouldn't take too long :sweat_smile:

fairingrey:
Currently waiting on both https://github.com/rustasync/accept-encoding/pull/2 and https://github.com/rustasync/async-compression-rs/pull/13 until I continue progress.

Cargo.toml
```diff
@@ -29,11 +29,20 @@ route-recognizer = "0.1.12"
...
+serde_urlencoded = "0.5.5"
 slog = "2.4.1"
 slog-async = "2.3.0"
 slog-term = "2.4.0"
 typemap = "0.3.3"
-serde_urlencoded = "0.5.5"
+
+[dependencies.accept-encoding]
+optional = true
+git = "https://github.com/fairingrey/accept-encoding"
```
pickfire: Should we move to rustasync?

fairingrey: It's there, I'm just waiting for yoshua to review and merge https://github.com/rustasync/accept-encoding/pull/2

tirr-c:
How do you think about using `Encoding` and `Option<Encoding>` accordingly? Currently we have `Encoding::Any` (or `Encoding::None`) to indicate "no specific preference", but it seems like `Encoding::Any` is often special-cased.

fairingrey:
Oh, that's a good approach. Using `Option<Encoding>` actually hadn't crossed my mind while writing it, but I think it'd look pretty decent.

I can rewrite that part in the Accept-Encoding crate such that when its' re-exported here I don't have to special-case it. I'll try it in a moment.

Cargo.toml
fairingrey: This was added back because it turns out I needed it for testing decompression: https://github.com/rustasync/tide/blob/30a3b1404f6b27fdce8bcd97fc04124706b9f684/src/middleware/compression.rs#L253

Apparently, Travis and rustfmt don't like it when it's included in dev-dependencies, likely because it's already included through `http-service`, so it's in the regular dependencies. Not that it should matter much though, aside from it being exposed I suppose.

src/middleware/compression.rs
```diff
@@ -0,0 +1,434 @@
...
+
+    fn preferred_encoding(&self, headers: &HeaderMap) -> Result<Encoding, Error> {
+        let encoding = match accept_encoding::parse(headers) {
+            Ok(encoding) => encoding,
+            Err(e) => return Err(Error::from(StatusCode::BAD_REQUEST)),
+        };
+        match encoding {
+            Some(encoding) => Ok(encoding),
+            None => Ok(self.default_encoding),
+        }
```
tirr-c: You can use `encoding.unwrap_or(self.default_encoding)` here.

src/middleware/compression.rs
```diff
@@ -0,0 +1,434 @@
...
+    }
+
+    /// Consumes the response and returns an encoded version of it.
+    fn encode(&self, mut res: Response, encoding: Encoding) -> Response {
+        // TODO: test and make sure this works
+        match encoding {
```
tirr-c: There might be a way to eliminate duplicate codes... like, early-return on `Encoding::Identity` first, then make `match` evaluate to `Body`?

src/middleware/compression.rs
```diff
@@ -0,0 +1,434 @@
...
+                Err(e) => return e.into_response(),
+            };
+            let res = next.run(cx).await;
+            self.encode(res, encoding)
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+enum Decoding {
```
tirr-c: Can't we just use `Encoding` for this? (I'm not quite familiar with `Content-Encoding` and `Accept-Encoding` things)

src/middleware/compression.rs
tirr-c: (Now I think early-return might not be necessary)

src/middleware/compression.rs
fairingrey: Ah, right. I forgot about `unwrap_or`. I'll do that.

src/middleware/compression.rs
fairingrey: We can -- I had to include this separately because of the existence of `Encoding::Any` previously, but I should be able to simplify this now.

src/middleware/compression.rs
fairingrey: I think I could do this, but I'm not a *huge* fan of the way it looks (as I'd have to match the following `Encoding::Identity` arm with `unreachable!()`)... though I do agree, there is a lot of duplicate code here.

tirr-c: I don't mind leaving as is, it was just a suggestion 🙂

fairingrey: Oh, no problem :smile:

Yeah, it bugs me a little bit because if the client prefers no encoding (or our default encoding is set to `Identity`, then we're not supposed to touch the body at all (hence I can't do the swap until afterwards).

fairingrey:
One thing I should note, by the way -- Travis will likely have to be updated unless the feature gate is removed. Cargo doesn't check/test/clippy all the features if the `--all-features` flag isn't included.

fairingrey:
This PR is ready for final review, by the way. Feel free to make any final notes or suggestions. I'm looking to get it merged probably before Thursday (5/16, the next meeting).

src/middleware/compression.rs
pickfire: Or maybe `unwrap_or_default` if it implements `Default`.

src/middleware/compression.rs
```diff
@@ -0,0 +1,420 @@
...
+#[derive(Debug, Default)]
+pub struct Decompression {}
+
+impl Decompression {
+    /// Creates a new Decompression middleware.
+    pub fn new() -> Self {
+        Self {}
+    }
+
+    fn parse_encoding(s: &str) -> Result<Encoding, Error> {
```
pickfire: Should we use `TryInto<Encoding> for &str` instead of this?

src/middleware/compression.rs
```diff
@@ -0,0 +1,420 @@
...
+            hval.split(',')
+                .map(str::trim)
+                .rev() // apply decodings in reverse order
+                .map(Decompression::parse_encoding)
+                .collect::<Result<Vec<Encoding>, Error>>()?
+        } else {
+            return Ok(());
+        };
+
+        for encoding in encodings {
```
pickfire: Should we move this block to https://github.com/rustasync/tide/pull/194/files#diff-4069e2d3cfd22c835d307b45346dc68eR177 so we do not need to `collect` the encodings as vector and directly decompress after https://github.com/rustasync/tide/pull/194/files#diff-4069e2d3cfd22c835d307b45346dc68eR176 since the error rate there would be a lot lower.

src/middleware/compression.rs
```diff
@@ -0,0 +1,420 @@
...
+        cx.body_bytes().await.unwrap()
+    }
+
+    // Generates the app.
+    fn app() -> crate::App<()> {
+        let mut app = crate::App::new();
+        app.at("/").get(lorem_ipsum);
+        app.at("/echo").get(echo_bytes);
+        app.middleware(Compression::new());
+        app.middleware(Decompression::new());
```
pickfire: Should we make it possible for people to use only one of `Compression` and `Decompression`?

fairingrey: I hadn't checked if limiting users to only one use of Compression and Decompression middleware each was possible yet, actually. Though I don't believe it can be done in this file -- such a check would likely exist in `src/middleware/mod.rs` or something, I imagine...

src/middleware/compression.rs
fairingrey: That would necessitate implementing `TryFrom<&str>` for `accept_encoding::Encoding`, and I'm not sure if I'm ready to go through all that work when the header value should be a valid UTF-8 string anyway :sweat_smile:

src/middleware/compression.rs
fairingrey: I mean I suppose we *could*, but the reason I do this here is because I don't think it should do unnecessary work unless it has to -- it helps if we know we can decode the request body from the getgo given a valid `Content-Encoding` header, and the overhead should usually be really cheap to begin with.

Say, for example, a client sends over a header `Content-Encoding: exi, gzip`, it's technically valid, *but* we don't know how to decode `exi`. If we moved this block there, we'd already be performing unnecessary work on a body we can't fully decode -- that is, per the reversal, we'd have `gzip, exi`, and after finishing decoding gzip we'd run into exi and then suddenly have to return the 415 error that `Decompression::parse_encoding` returns, `Err(Error::from(StatusCode::UNSUPPORTED_MEDIA_TYPE)`.

prasannavl:
@fairingrey - This is great! Thank you for all the hard work! :) 

Some thoughts: 
- One tiny little thing I'd have done is do this as a separate PR: https://github.com/rustasync/tide/pull/194/files#diff-e0934bb6f70a4f8c3baf22a773c14ec7 
- The `git dependencies` - I think in the long run, whichever can make it into the main `Cargo.toml` as `patch-crate.io` as git master on `rustasync` org owned _reasonably stable_ libs should be fine. Other than that, I think a good way to move forward for early deps would be to use a __pinned__ branch (say `tide`), rather than master - so code doesn't inadvertently get changed -- I say this since git deps will pass through the tide CI/CD, and there would be no way catch these, since they can be updated separately and end up wreaking havoc depending on when the user actually downloads the cargo crates.

src/middleware/compression.rs
fairingrey: Oh, I will remark though -- including multiple `Decompression` middlewares will usually have no side effect (aside from going through an *extra* middleware that doesn't do anything, as the decompression middleware already strips out the `Content-Encoding` header on finishing the decompression.

Including multiple `Compression` middlewares... likely will. I don't think users will willingly do this, but if there is a way to warn them of duplicate middlewares I'd be for it.

fairingrey:
@prasannavl Thank you!

On the topic of `Cargo.toml`, I could have, probably -- but it was just a small thing and I was including new dependencies anyway :sweat_smile:

As for the git dependencies, I agree. The `async-compression-rs` and `accept-encoding` crates likely won't be changed for a while (aside from non-breaking changes that Nemo or I am making), but it would be in our best interests to pin a stable branch we *know* won't be changing without the necessary PRs to update them.

src/middleware/compression.rs
pickfire: I wonder if there are any ways to warn users when multiple `Compression` or `Decompression` is used.

Nemo157: It might be worth the `Compression` middleware checking for an existing `Content-Encoding` header on the body to avoid double encoding (e.g. it's relatively common to have static file servers that check for existing pre-compressed versions of files and serve those based on `Accept-Encoding`, you should be able to use both that and a generic `Compression` middleware to compress other responses that weren't pre-compressed).

fairingrey: Oh, good catch. I can probably throw that in real quick.

fairingrey:
Actually, lemme make one tiny new change to decrease compilation time (since this middleware doesn't use the bufread features in async-compression).

src/middleware/compression.rs
```diff
@@ -86,44 +86,48 @@ impl Compression {
...
-            Encoding::Zstd => {
-                let mut body = Body::empty();
-                std::mem::swap(&mut body, res.body_mut());
-                let stream = stream::ZstdEncoder::new(body, self.settings.zstd);
-                *res.body_mut() = Body::from_stream(stream);
-                res.headers_mut()
-                    .append(CONTENT_ENCODING, encoding.to_header_value());
-                res
+        if res.headers().get(CONTENT_ENCODING).is_some() {
+            res // avoid double-encoding a given response
```
pickfire: Should we do a quick return here so that the else branch below could have one indent lesser?

fairingrey: We could. I'll try it in a sec, it might look better.

Although I'm unsure if the borrow checker will complain without the else block anyway... Will try in a moment

src/middleware/compression.rs
fairingrey: @tirr-c I decided to go with reducing the boilerplate after all, it reads easier visually I think, especially with the new check in place for if a response was already encoded.

src/middleware/compression.rs
```diff
@@ -0,0 +1,412 @@
...
+    fn encode(&self, mut res: Response, encoding: Encoding) -> Response {
+        if res.headers().get(CONTENT_ENCODING).is_some() || encoding == Encoding::Identity {
+            return res; // avoid double-encoding a given response
+        }
+        let mut body = Body::empty();
+        std::mem::swap(&mut body, res.body_mut());
+        let mut res = match encoding {
+            Encoding::Gzip => {
+                let stream = stream::GzipEncoder::new(body, self.settings.gzip);
+                *res.body_mut() = Body::from_stream(stream);
```
tirr-c: Instead of mutating `res` directly in match arms, we can collect `Body` and mutate `res` after match!

src/middleware/compression.rs
```diff
@@ -0,0 +1,412 @@
...
+        Ok(encoding.unwrap_or(self.default_encoding))
+    }
+
+    /// Consumes the response and returns an encoded version of it.
+    fn encode(&self, mut res: Response, encoding: Encoding) -> Response {
+        if res.headers().get(CONTENT_ENCODING).is_some() || encoding == Encoding::Identity {
+            return res; // avoid double-encoding a given response
+        }
+        let mut body = Body::empty();
+        std::mem::swap(&mut body, res.body_mut());
```
tirr-c: [`std::mem::replace`][1] would look much nicer here, I think 😄 

[1]: https://doc.rust-lang.org/std/mem/fn.replace.html

fairingrey: Sure thing, I'll try it

fairingrey: Mmm... it looks like I can't, and I think I know why.

```
error[E0507]: cannot move out of borrowed content
  --> src/middleware/compression.rs:93:38
   |
93 |         std::mem::replace(&mut body, *res.body());
   |                                      ^^^^^^^^^^^ cannot move out of borrowed content
```

As I understand it, trying to replace the reference at `body` when an empty body is replaced moves `*res.body()` into body which isn't allowed... Versus using swap, where instead, we're allowed to take ownership of `res.body_mut()`, just replacing what's there with an empty `Body` temporarily while we create a new stream with the response body we now own...

Unless I can get around that some other way, then hmm... But I don't think replacing it outright with `std::mem::replace` will just work.

src/middleware/compression.rs
fairingrey: Good catch! (I'm not sure why I didn't see this earlier, probably because I was a bit quick to rewrite this code)

Will be pushing it in a sec, thanks :+1:

src/middleware/compression.rs
tirr-c: I was thinking like this:

```rust
let body = std::mem::replace(res.body_mut(), Body::empty());
```

Here we replace `res.body_mut()` with `Body::empty()` and get its previous value as `body`.

fairingrey: Oh *that's* what you meant, my bad!

Sure, lemme try that. That looks like it should work.

fairingrey:
Since #219 was merged, looks like I'll be doing a rebase now.

Thanks for all the review and feedback!

tide-compression/README.md
```diff
@@ -0,0 +1,33 @@
...
+    cx.body_bytes().await.unwrap()
+}
+
+fn main () {
+    let mut app = crate::App::new();
+    app.at("/").get(lorem_ipsum);
+    app.at("/echo").get(echo_bytes);
+    app.middleware(Compression::with_default(Encoding::Brotli));
+    app.middleware(Decompression::new());
+    app
```
tirr-c: Did you mean to serve this app? 😅

tide-compression/README.md
```diff
@@ -0,0 +1,33 @@
...
+## Examples
+
+__Basic usage__
+
+```rust
+use tide::App;
+use tide_compression::{Compression, Decompression, Encoding};
+
+// Returns a portion of the lorem ipsum text.
+async fn lorem_ipsum(_cx: Context<()>) -> String {
```
tirr-c: `Context` is not imported.

tide-compression/README.md
fairingrey: ...woops :sweat_smile: 

My copypasta-fu has failed me...

tide-compression/README.md
fairingrey: Yeah, copypasta-fu failure, *ahem*

fairingrey:
Actually, I'll probably move the example out to the examples folder, thanks for the suggestion :+1:

fairingrey:
@tirr-c Sorry about that! How about now?

tide-compression/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+Examples are in the `/examples` folder of this crate.
+
+__Simple Example__
+
+You can test the simple example by running `cargo run --example simple` while in this crate's directory, and then running either of the following commands:
+
+```console
+$ curl http://127.0.0.1:8000/ -v
+$ curl http://127.0.0.1:8000/echo -v -d "why hello there"
```
tirr-c: This command successfully tested compression and decompression.

```suggestion
$ echo 'why hello there' | gzip | curl -v --compressed -H 'Content-Encoding: gzip' 'http://127.0.0.1:8000/echo' --data-binary @-
```

fairingrey: Oh, I totally forgot about the existence of the *actual* gzip binary... Woopsiedaisy. Thanks for your help!

fairingrey:
Indeed... There's a lot of stuff here, and this is a fairly large component -- it's good to get lots of review in so we won't have to double over it so much when it's integrated :sweat_smile: 

But I think this is good now! It should be good to merge tomorrow, *hopefully* :sweat_drops:

tide-compression/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+categories = ["network-programming", "compression", "asynchronous"]
+edition = "2018"
+license = "MIT OR Apache-2.0"
+name = "tide-compression"
+readme = "README.md"
+repository = "https://github.com/rustasync/tide"
+version = "0.1.0"
+
+[dependencies]
+tide = { path = "../tide" }
```
tirr-c: We need a `version` field in order to publish this crate. Let's do this after structure revamp!

fairingrey:
Thanks for all the review, everyone! I'll get this merged now since it looks like we have not much else to change.

If anything else needs to be sharpened up, we can do it through a future PR, I imagine.

PR #198, opened 2019-05-03
Change App test code simply

k-nasa:
## Description

Change App test code simply

## Motivation and Context

The previous code may be misunderstood, Need to be nested

## How Has This Been Tested?

Run test code.

## Types of changes

refactor

## Checklist:

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -364,9 +364,7 @@ mod tests {
     #[test]
     fn multiple_methods() {
         let mut router = App::new(());
-        router.at("/a").nest(|router| {
-            router.at("/b").get(async move |_| "/a/b GET");
-        });
+        router.at("/a/b").get(async move |_| "/a/b GET");
         router.at("/a/b").post(async move |_| "/a/b POST");
 
         for (path, method) in &[("/a/b", http::Method::GET), ("/a/b", http::Method::POST)] {


secretfader:
The original implementation was also testing the functionality of `.nest`, a method that can save a few keystrokes (and probably reduce human errors) when dealing with nested paths.

I don't quite understand why flattening the code (and thereby, removing a feature we may have intended to test) is helpful.

k-nasa:
@secretfader I made this change because I thought it would mislead "Nesting allows multiple methods to be assigned to the same path". 
 The tests for nesting seemed to be done [here](https://github.com/rustasync/tide/blob/68b73fd3479ef2ade241401b291f3b363e1b9af2/src/app.rs#L324).

k-nasa:
@secretfader Should do it close ❓

secretfader:
@k-nasa I'd prefer to hear a few opinions from others on the topic. I think leaving this PR open is fine, for now.

Whether it was intentional or not, one scenario that is addressed by the previous code (and isn't in the tests covering nested routes specifically) is that `GET` and `POST` should return different values, even if **one path is nested while the other isn't.** Maybe that's a point in favor of the previous implementation?

yoshuawuyts:
@k-nasa thanks for this patch! I agree with @secretfader that this reduces the breadth of our tests, which doesn't seem like a desirable change. Is it okay if we close this?

k-nasa:
@yoshuawuyts Ok. I close it, And sorry for unnecessary PR.

yoshuawuyts:
@k-nasa no worries at all! -- thank you for all the PRs you've been writing!

PR #202, opened 2019-05-07
Fix doc links

k-nasa:
## Description

Currently method link is plain text, Because link format is invalid. I fixed it.

スクリーンショット 2019-05-07 10 54 34

スクリーンショット 2019-05-07 10 55 08

## How Has This Been Tested?

I see new documentation.

スクリーンショット 2019-05-07 10 57 15

スクリーンショット 2019-05-07 10 57 10

## Types of changes

## Checklist:

### DIFF
--- a/src/route.rs
+++ b/src/route.rs
@@ -2,10 +2,12 @@ use crate::{router::Router, Endpoint};
 
 /// A handle to a route.
 ///
-/// All HTTP requests are made against resources. After using [`App::at`](crate::app::App::new) (or
+/// All HTTP requests are made against resources. After using [`App::at`] (or
 /// [`Route::at`]) to establish a route, the `Route` type can be used to
 /// establish endpoints for various HTTP methods at that path. Also, using
 /// `nest`, it can be used to set up a subrouter.
+///
+/// [`App::at`]: ./struct.App.html#method.at
 pub struct Route<'a, State> {
     router: &'a mut Router<State>,
     path: String,
--- a/src/route.rs
+++ b/src/route.rs
@@ -2,7 +2,7 @@ use crate::{router::Router, Endpoint};
 
 /// A handle to a route.
 ///
-/// All HTTP requests are made against resources. After using [`App::at`] (or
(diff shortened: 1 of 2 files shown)


src/route.rs
```diff
@@ -2,7 +2,7 @@ use crate::{router::Router, Endpoint};
 
 /// A handle to a route.
 ///
-/// All HTTP requests are made against resources. After using [`App::at`] (or
+/// All HTTP requests are made against resources. After using [`App::at`](crate::app::App::new) (or
 /// [`Route::at`]) to establish a route, the `Route` type can be used to
 /// establish endpoints for various HTTP methods at that path. Also, using
 /// `nest`, it can be used to set up a subrouter.
```
yoshuawuyts:
```suggestion
 /// `nest`, it can be used to set up a subrouter.

[`App::at`]: ../struct.App.html#method.at
```

k-nasa: Why should I change the notation of the path?

src/route.rs
```diff
@@ -2,7 +2,7 @@ use crate::{router::Router, Endpoint};
 
 /// A handle to a route.
 ///
-/// All HTTP requests are made against resources. After using [`App::at`] (or
+/// All HTTP requests are made against resources. After using [`App::at`](crate::app::App::new) (or
```
yoshuawuyts:
```suggestion
/// All HTTP requests are made against resources. After using [`App::at`] (or
```

src/route.rs
yoshuawuyts: My understanding was that [`tide::Route`](https://docs.rs/tide/0.2.0/tide/struct.Route.html) wasn't rendering the link to `App::at` correctly.

Using link blocks at the bottom of a paragraph is pretty standard practice in Rust's stdlib, so I that's why I proposed making this change.

k-nasa: Ok.  I did not know that I could include a link block. Thanks 😄

k-nasa:
@yoshuawuyts I fixed it!  Please check when there is time

PR #206, opened 2019-05-10
Delete rustfmt.toml

nrc:
Both options used are defaults, and it is encouraged not to configure rustfmt if possible.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,2 +1,0 @@
-edition = "2018"
-tab_spaces = 4


JeanMertz:
Just a quick note: I actually prefer to add an empty `rustfmt.toml` file with something like

> `# defaults are great :)`

as its content.

Reason being that many projects aren't using rustfmt yet, so by adding a file, you signal to contributors that you'd like them to rustfmt their code. Obviously this can also be added to a CONTRIBUTING.md, but the added benefit of adding the `rustfmt.toml` file is that code editors can enable/disable rustfmt integration based on the existence of that file (which is what I do in my Vim setup).

That way, you can use rustfmt in projects like tide, without the risk of getting (accidentally committed) massive diffs in other project that don't yet use rustfmt, resulting your editor unintentionally changing all of a file's formatting.

Issue #187, opened 2019-04-24
Decide on extension trait convention


aturon:
Tide uses many "extension traits" (which extend an existing type with new methods, and have no other impls). We should standardize on a naming convention.

aturon:
One possibility would be `FooExt` where `Foo` is the name of the type being extended -- including the type in the name helps the reader quickly see what is being extended.

A downside is that we want to provide many extensions to `Context`. We'd end up with `cookies::ContextExt` and `forms::ContextExt` etc. But given that you only use these trait names on import, where they will be qualified by module, perhaps this is fine? The paths do contain the full set of relevant info: what is the extension, and what is being extended?

bIgBV:
I think using `ContextExt` would be a good idea. This would be the common type name across the entire ecosystem which would signal that this type extends something on tide's `Context`. I think the recognizability this achieves will be really useful to convey the idea of how it's supposed to be used. Overall it provides a consistent name for the core extension trait in a given crate across all such crates.

@aturon  I agree with you, given that the type itself will only be present at the top of the file for the most part and the type will not be used anywhere else, the import path will give enough context as to where this is coming from.

secretfader:
If I apply this specification to #175, I feel like we have several options:

- `QueryStringExt`
- `UrlQueryExt`
- `QueryExt`

I'm not entirely sure which is appropriate, but the method it defines is `url_query`. So `UrlQueryExt` is probably the most intuitive.

fairingrey:
I'm in support of using the `*Ext` naming scheme -- it's what I see a lot in the ecosystem already and feels the most familiar.

As for the comment above https://github.com/rustasync/tide/issues/187#issuecomment-486456855, yeah I'd also go with `UrlQueryExt`. I think `QueryExt` is a bit too unclear, and `QueryStringExt` doesn't quite feel right (as while it is a string deserialization it doesn't quite describe the entire nature of the extension trait).

secretfader:
@fairingrey Ignore my comment above. I completely misunderstood the convention. A more appropriate naming scheme for the `querystring` module is `tide::querystring::ContextExt`.

fairingrey:
Oh, sorry! Yeah, that makes sense.

Nemo157:
A downside of the chosen convention I recently ran into is not being able to actually bring multiple `ContextExt` into scope. A workaround is to use the relatively recent `use foo::Foo as _;` feature to import the traits _methods_ into scope without having the trait nameable:

```rust
[...]

PR #218, opened 2019-05-13
Rename extension traits for Cookies and QueryString

secretfader:
As discussed in #187, functionality of extension traits be identifiable by import path, and the trait itself should be named after the type it extends.

To satisfy those constraints, `querystring::ExtractQuery` becomes `querystring::ContextExt`. Similarly, `cookies::CookiesExt` is now `cookies::ContextExt`. 

Ideally, these changes will land before #162.

## Motivation and Context

To align with Rust best practices and the direction of current project discussions.

## How Has This Been Tested?

Current tests pass after being modified to import new types.

## Types of changes

## Checklist:

### DIFF
--- a/examples/cookies.rs
+++ b/examples/cookies.rs
@@ -1,7 +1,7 @@
 #![feature(async_await)]
 
 use cookie::Cookie;
-use tide::{cookies::CookiesExt, middleware::CookiesMiddleware, Context};
+use tide::{cookies::ContextExt, middleware::CookiesMiddleware, Context};
 
 /// Tide will use the the `Cookies`'s `Extract` implementation to build this parameter.
 ///
--- a/src/cookies.rs
+++ b/src/cookies.rs
@@ -28,7 +28,7 @@ impl CookieData {
 }
 
 /// An extension to `Context` that provides cached access to cookies
-pub trait CookiesExt {
+pub trait ContextExt {
     /// returns a `Cookie` by name of the cookie
     fn get_cookie(&mut self, name: &str) -> Result<Option<Cookie<'static>>, StringError>;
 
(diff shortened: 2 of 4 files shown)


prasannavl:
@secretfader This looks good to me! 

However, this is a breaking change. So, would rather mind holding off on this until #162. 

- Let's introduce the PR(s) for #162. Then since the cookies bit will get separated anyway, we should be free to change that. Perhaps, you could take the `tide-cookies` and do it as a part of that?

secretfader:
@prasannavl That's one way to handle this, and I'm certainly open to it. Discussions with @aturon in Discord seemed to indicate they hoped to land these changes alongside RFC #162 (which is also a breaking change).

I don't care too much when this PR goes in. That said, if we're making the decision to not merge because it has a breaking change (and we're discussing other breaking changes to be merged shortly afterwards), is there a significant difference between including it now to potentially reduce the surface area of other PRs and waiting until after they land?

I'd like to hear your reasoning on why merging later is the better option.

prasannavl:
@secretfader  - Just opened - https://github.com/rustasync/tide/pull/219. Hopefully, this helps clarify my motivations on to keep it separate. 

- #219 despite seemingly a complete revamp of the organisation is a non-breaking change. 
- Now the next step would be move bits out of tide into it's own crate.
- My rationale was that you could now use this extraction opportunity to make these changes as a part of it, since it's going to be a new crate anyway.

secretfader:
> My rationale was that you could now use this extraction opportunity to make these changes as a part of it, since it's going to be a new crate anyway.

`master` has already moved past a point where it would build with an `0.2` version dependency on Tide. Yes, I'm introducing a breaking change, but it isn't the first since our last release (which would make your opposition far more valid).

I don't see why this PR has to wait until #219 lands, other than for convenience. Integrating a smaller change like this into your roll-up isn't significant.

PR #219, opened 2019-05-13
Precursor: Tide core isolation revamp

prasannavl:
## Description
- Initial PR for complete revamp of organisation, as discussed in https://github.com/rustasync/tide/issues/162

### What's done

- [x] Move the main crate into `tide` and setup workspaces
- [x] Moves the examples into it's own 

The above is still a non-breaking change and as such can be merged without really isolating the core and as such is meant to be a precursor PR to the actual isolation.

### What's explicitly excluded from as a part of this PR
- Move each middleware into it's own crate from the `tide` crate

## Motivation and Context
https://github.com/rustasync/tide/issues/162 

## How Has This Been Tested?
All tests passed

## Types of changes

- [x] Refactoring and organisational changes 

## Checklist:


prasannavl:
@tirr-c - Ah, sounds good. I like it! However, that could be a separate PR? I don't think we need to mix it up with this.

examples/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+  "network-programming",
+  "asynchronous",
+  "web-programming::http-server"
+]
+edition = "2018"
+license = "MIT OR Apache-2.0"
+name = "tide-examples"
+readme = "README.md"
+repository = "https://github.com/rustasync/tide"
+version = "0.1.0"
```
Nemo157: Do you intend to publish these examples as a crate? If not it would be good to add `publish = false` to avoid accidental publishing (and could drop `keywords` and `categories` since they won't be used).

examples/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+documentation = "https://docs.rs/tide"
+keywords = ["tide", "http", "web", "framework", "async"]
+categories = [
+  "network-programming",
+  "asynchronous",
+  "web-programming::http-server"
+]
+edition = "2018"
+license = "MIT OR Apache-2.0"
+name = "tide-examples"
```
Nemo157: Also, if this is not going to be published then for ease of running it would be nice to drop the `tide-` prefix here. Then users will be able to `cargo run -p examples --bin body_types` to run a specific example.

examples/Cargo.toml
prasannavl: Done! Thanks!

examples/Cargo.toml
prasannavl: And done :)

wafflespeanut:
Does it make sense to move the examples to `tide` crate and enable `autoexamples = true` in `Cargo.toml` instead of maintaining a separate crate for it? (I did that in my attempt [here](https://github.com/wafflespeanut/tide/tree/code-split/tide)).

prasannavl:
@wafflespeanut - I think that would make sense for direct library based examples usually. Since we're going to have examples that span across middlewares and other deps, I think it'd be easier to maintain a separate crate, which will also double up as an integration test.   

Running should be easy once the bin splits happen.

wafflespeanut:
Okay cool. Is this good to go? Shall I branch from this and start working on splitting stuff?

prasannavl:
@wafflespeanut - I think so.  But I'd probably prefer to sit on it for a couple of hours before merging just to let the clock run to make sure there's a sufficient window for folks to catch up.  

Changes, if any should be minimal, so it's a good idea to base other work off this.

secretfader:
LGTM. That said, there's little reason to block other PRs until this one is completed. I elaborate more on that in #218.

Other PRs that were ready to go before this one was introduced should be merged.

prasannavl:
@secretfader __This is not at all cool__. I really wish you didn't just go and merge your PR esp, without any one else weighing in on it in such a hurry esp, as I had explicitly expressed my concerns. https://github.com/rustasync/tide/pull/218#issuecomment-491864964 .. Now all of the work on top, including ones which are ready to be merged will have to be rebased, even though I'm just waiting to let the clock run. 

What really disconcerts me is that this is for a simple rename that will be extracted out into it's own crate any way shortly, and I really don't understand the hurry in that, and I've been trying to do the whole revamp in a non-breaking way as well as much as possible, and you do this. 

If you really must, I wish you had atleast discussed it, or asked other maintainers' suggestion.

prasannavl:
Alright, we're good to go! Since this will unblock other work to move forward, going ahead and merging this! /cc @wafflespeanut @fairingrey

PR #220, opened 2019-05-14
Split tide into smaller crates

wafflespeanut:
The following changes have been made (so far):
 - Extracted core types and traits from `tide` into `tide-core`
 - Removed unnecessary deps
 - Extracted cookie `ContextExt` and middleware into `tide-cookies` and feature-gated.

## Description

This is a follow-up to #219 and so it's based on top of the commits introduced in the other PR.

## Motivation and Context

Based on the discussion from #162 

## How Has This Been Tested?

It's a refactor which only affects the existing structure. It doesn't break anything.

## Types of changes

- [x] Refactor

## Checklist:


tide-core/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+version = "0.2.0"
+edition = "2018"
+authors = [
+  "Tide Developers",
+]
+description = "Core types and traits for Tide web framework"
+documentation = "https://docs.rs/tide-core"
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/rustasync/tide"
+autoexamples = true
```
Nemo157: This is the default so doesn't need to be specified.

tide-core/src/app.rs
```diff
@@ -31,6 +31,7 @@ use crate::{
 ///
 /// ```rust, no_run
 /// #![feature(async_await)]
+/// # use tide_core as tide;
```
Nemo157: The way `futures` deals with this "writing doc examples from the perspective of the facade" is to add `futures` as a dev-dependency of `futures-utils` etc. You could do the same here, add `tide` as a dev-dependency of `tide-core` and the examples will work to reference the types through the facade.

tide-core/Cargo.toml
wafflespeanut: Yeah, this is a remnant from a previous attempt. I've removed it locally :smile:

tide-core/src/app.rs
wafflespeanut: Interesting. Will do.

tide/Cargo.toml
```diff
@@ -19,34 +19,27 @@ repository = "https://github.com/rustasync/tide"
...
 slog-async = "2.3.0"
 slog-term = "2.4.0"
-typemap = "0.3.3"
 serde_urlencoded = "0.5.5"
-
-[dependencies.http-service-hyper]
-optional = true
-version = "0.2.0"
+tide-cookies = { path = "../tide-cookies", optional = true }
+tide-core = { path = "../tide-core", optional = true }
```
Nemo157: Should this be optional? It looks like it's only included in the `hyper` feature so building with `--no-default-features` probably doesn't work.

Nemo157: Yep, just tested locally and this fails with

```console
> cargo build --no-default-features --manifest-path tide/Cargo.toml --verbose
```

CI will need some work to be able to test feature flags accurately, I'll open an issue and probably try and port some of the stuff we have setup in `futures` here for it.

wafflespeanut: Good catch! Fixed.

wafflespeanut:
We should discuss this in the relevant RFC, as we've already arrived at a consensus for `tide`, `tide-core` and how middlewares should be implemented. The goal of this PR is to restructure the existing codebase without affecting its usage/functionality.

I agree that `tide-core` should be as minimal as possible, but I don't think we should achieve that in a single PR, as it's getting complicated as we address deeper issues.

I also believe that chopping an epic issue into smaller tasks and landing small units of work is much more effective compared to blocking all tasks related to an RFC on one individual. If we land this now, then relatively smaller tasks like say, moving some type to `http-service` could be *mentored* work for newcomers. It doesn't necessarily have to be done by me (I will do it, just making a point).

One other (purely selfish) reason to land this is that restructuring PRs are fragile. It's already broken twice and it's analog work to go through the newly landed commits to ensure that a rebase was done properly. I'm pretty sure if we boot up another discussion now, then by the time we arrive at a consensus, it's gonna break a few more times again.

In addition to being minimal, `tide-core` should also do what its name implies - have the *core* types and traits. With that, I'll address the concerns:

 - It's sensible to move `Response`-convertible traits into `http-service`
 - While it makes sense to have an error type in `http-service`, I think tide should have its own error.
 - `Endpoint` is a core trait, so it should be in `tide-core`, whereas `EndpointResult` is a convenience type, so it can probably move to `tide`? (I don't wanna conclude anything based on that just yet).
 - We have a bunch of private entities (say, initializing `Context`, `Next`, etc.) - `App` and `Server` make use of such entities. That's why I moved them to core. Having `App` and `Server` in tide implies that we have to expose those entities publicly.
 - I feel `tide-cookies` as a separate crate is a nice segregation (same applies to future middlewares) as it isolates dependencies and aids semver.

prasannavl:
@wafflespeanut - I'm in a bit of time crunch, but I'll address some of the non technical bits really quickly.

> We should discuss this in the relevant RFC, as we've already arrived at a consensus for tide, tide-core and how middlewares should be implemented. The goal of this PR is to restructure the existing codebase without affecting its usage/functionality.
> 
> I agree that tide-core should be as minimal as possible, but I don't think we should achieve that in a single PR, as it's getting complicated as we address deeper issues.

I get the feeling you may have misunderstood.

- The key is to get the _insight_ so that we can be cognisant of where this split will leave us, and as such if there's a compelling reason to change specific bits about this PR, to do so before merging - as the consensus for tide, tide-core, and middleware are quite tied to this PR as well.
- Things like `IntoResponse`, `Error` are definitely RFC materials - __My question here is how do we go about handling them in the context of this split__ - "Leave it in tide_core" _could_ be valid answer too, as long as the reviewers agree.

> One other (purely selfish) reason to land this is that restructuring PRs are fragile. It's already broken twice and it's analog work to go through the newly landed commits to ensure that a rebase was done properly. I'm pretty sure if we boot up another discussion now, then by the time we arrive at a consensus, it's gonna break a few more times again.

1. First off, I'm sorry that you had ended up with this feeling in the first place. Early stage restructuring PRs are usually tricky to handle, and couple that with early stage open source project, and this becomes quite challenging to come up with a comprehensive design plan before hand to avoid this.

1. I'd ask that you not worry about this too much. It's neither fair nor needed for you alone to take the burden of keeping up with the changes. We're here to help (and while I can't speak for others, if you'd like, I'd be happy to help rebase and keep up with the changes if that helps us stay on the right track). But I personally think _this by itself a wrong reason to merge _any_ PRs_.

1. That being said, this is your PR and the decision of whether you'd be okay with someone else helping you rebase your PR is upto you :).

The ultimate goal here is to gather the reviews, and then make sure the PR is something that can be structured into the limitations of your interest and time, while not compromising on the quality -- if it isn't already. 

PS. Also, if you could, I think it'd be great for you bring this PR up, as the rustasync weekly meeting fortunately is in a couple of hours! (I'd guess it'll help speed this up).

wafflespeanut:
I'm sorry if I'm misunderstanding something.

If there's a compelling reason to make any changes in this PR, I'd be happy to do that! Re-iterating my argument, this PR aims to address *one thing* without introducing any breaking changes and if we all agree on what it does and how we can work on the changes introduced by it incrementally, then we can move on to the next task, instead of waiting on the PR to address *everything* related to the split as it's inefficient IMO.

In short, I'm only worried about making this PR address tasks that could otherwise be done incrementally. Rebasing and keeping up with the changes isn't tricky (it's just boring analog work), I'm not worried about it and I'm definitely not suggesting that as a reason to merge this PR. :)

README.md
```diff
@@ -36,7 +36,7 @@
       API Docs
     </a>
     <span> | </span>
-    <a href="https://github.com/rustasync/tide/blob/master/.github/CONTRIBUTING.md">
+    <a href="./.github/CONTRIBUTING.md">
```
yoshuawuyts: The URLs were intentionally long-form so they correctly link from e.g. crates.io

tide/src/lib.rs
```diff
@@ -28,27 +25,19 @@ macro_rules! box_async {
...
-mod endpoint;
+pub mod error;
 pub mod forms;
 pub mod middleware;
 pub mod querystring;
-pub mod response;
-mod route;
-mod router;
+
+pub use tide_core::response;
```
yoshuawuyts: Can we add `#[doc(inline)]` above all `tide_core` re-exports?

yoshuawuyts: Or better: this could probably be part of line 44.

tide/src/lib.rs
```diff
@@ -28,27 +25,19 @@ macro_rules! box_async {
 }
 
 #[macro_use]
-pub mod error;
+extern crate tide_core;
+
+#[cfg(feature = "cookies")]
+pub use tide_cookies as cookies;
```
yoshuawuyts: This should probably have `#[doc(inline)]` above it.

README.md
wafflespeanut: Oh. Oops.

tide/src/lib.rs
wafflespeanut: I've somehow missed this.

tide/src/lib.rs
wafflespeanut: Added.

wafflespeanut:
Fixed and rebased!

tide/src/error.rs
```diff
@@ -1,105 +1,6 @@
 use core::pin::Pin;
 use futures::future::Future;
-use http::{HttpTryFrom, Response, StatusCode};
-use http_service::Body;
 
-use crate::response::IntoResponse;
+pub use tide_core::error::*;
```
fairingrey: Personal nitpick, but I don't like using glob imports outside of test modules. Would prefer all the related error types be exported explicitly.

wafflespeanut: Updated.

fairingrey:
Looks good to me, thanks for your work!

wafflespeanut:
Thank you all for reviewing!

prasannavl:
I'm a bit busy this weekend, and I've not been able to make time to think through more clearly some of potential directions of what I expressed above just yet. 

If everyone else is cool with the PR, I'll jump back into the extraction bits post merge :) 

One thing I'd like to __strongly insist__ in that case, would be to have `tide_core` as a private dependency for now until we can answer most of those questions, and do the refactoring needed, since @yoshuawuyts is planning on cutting a new release soon.

Nemo157:
`tide-core` appears to be as much of a private dependency as it can be at this stage, it's not mentioned anywhere I can find in the documentation of `tide`, and I doubt users will just start depending on it without understanding what it is for.

PR #221, opened 2019-05-15
Fix example link in readme

daleione:


prasannavl:
I missed this! Thanks for the catch! :)

PR #222, opened 2019-05-15
add tide-log

prasannavl:
## Description
- A basic request logger that only uses the `log` crate and nothing else.
- Resolves https://github.com/rustasync/tide/issues/152
- Resolves https://github.com/rustasync/tide/issues/107

Depends on #220 

## Motivation and Context

- https://github.com/rustasync/tide/issues/162
- Progress https://github.com/rustasync/tide/issues/8

## How Has This Been Tested?

Running on several different environment successfully, and merely hands off to `log` crate.  

## Types of changes

## Checklist:


tide-log/src/lib.rs
```diff
@@ -0,0 +1,65 @@
+#![feature(async_await)]
+#![deny(
+    nonstandard_style,
+    rust_2018_idioms,
+    future_incompatible,
+    missing_debug_implementations
+)]
```
Nemo157: While this does seem to be the standard in this repo, I would suggest moving away from using `deny` in the crate as that just makes developers lives harder when they have temporary warnings being emitted during development. It's better to turn on extra lints via `warn` and then have CI include `--deny=warnings` to force developers to ensure code is warning free before it makes it to master (I know this is more of a policy question than specific to this PR, but the introduction of a new crate seems like a good time to bring up cross-crate standards like this).

prasannavl: I agree! It followed along from the core. I think it makes perfect sense to do this only for the core, and let other crates have warnings instead. Is that what you had in mind?

PS. Could we leave `nonstandard_idioms` at deny though? Or would you prefer to leave everything as just warnings?

cc/ @yoshuawuyts @tirr-c

Nemo157: Personally I do not believe _any_ crate should _ever_ use `deny` _in code_. I'm a very big proponent of using Clippy and enabling extra warnings that are `allow`-by-default, but those should be denied _in CI only_. From what I have seen the Rust community as a whole seems to be onboard with treating all warnings as issues that need fixing, so it's highly likely they will fix them before even opening a PR. If a developer doesn't then CI will fail and I have yet to see a developer that has complained about being forced to fix warnings in their PRs.

Denying warnings in code has some significant downsides:

 * When you are mid development it causes extra compile cycles because of trivial issues e.g. missing putting `#[derive(Debug)]` on a new struct. It's not uncommon to add code while developing that you later decide is not necessary and gets deleted, being forced to fix warnings in that code just so you can run your tests is a real pain.

 * It causes incompatibility with future Rust versions. The set of code that a particular lint applies to is not guaranteed to stay the same across Rust versions, if it expands to cover something in the current codebase then users that have a `path` dependency on Tide or are in some other way compiling it without `--cap-lints` will no longer be able to build their project.

prasannavl: Thanks @Nemo157! This is an incredibly valuable insight! Personally, I tend to use Makefiles which fails on publish/release and handle all the complexity there so it's consistent between CI and developers. So, I'm already convinced on this. 

Considering the `forbid` pattern was in tide since early days, should we wait until everyone pitches in or possibly today's meeting?

yoshuawuyts: @prasannavl I'm empathetic to Nemo's point here, and :+1: on loosening the constraints here if it makes contributing easier (:

prasannavl: Awesome! Let me do this as a separate PR and do a full sweep for all crates once #220 merges.

tide-log/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+authors = [
+    "Tide Developers"
+]
+description = "Tide middleware for logging"
+documentation = "https://docs.rs/tide"
```
fairingrey: When this crate is published, you might want to change the URL here :sweat_smile: 

Otherwise, looks good I think!

tide-log/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+keywords = ["tide", "http", "web", "framework", "async", "middleware"]
```
fairingrey: Crates.io is limited to 5 keywords here, I believe. You should remove one of these and perhaps replace another with "logging", as I think that describes this middleware crate better.

My personal suggestion is:

```toml
keywords = ["tide", "web", "async", "middleware", "logging"]
categories = ["network-programming", "logging", "asynchronous"]
```

Similar to https://github.com/rustasync/tide/blob/master/tide-compression/Cargo.toml

tide-log/Cargo.toml
prasannavl: Done. We also might want to add version to `tide` before release for all subcrates. I'll add them to #162

tide-log/Cargo.toml
prasannavl: Thanks @fairingrey! This pasted your suggestion directly for `keywords`, but categories, I used 
```
categories = [
  "logging",
  "network-programming",
  "web-programming::http-server",
]
```
instead, since while the middleware is inherently async, the logging particularly isn't! And I think the web bit is probably more relevant? :)

prasannavl:
Also rebased over #220 while tweaking the manifest.

tide-log/src/lib.rs
prasannavl: Converted to issue: https://github.com/rustasync/tide/issues/232

README.md
```diff
@@ -36,7 +36,7 @@
       API Docs
     </a>
     <span> | </span>
-    <a href="https://github.com/rustasync/tide/blob/master/.github/CONTRIBUTING.md">
+    <a href="./.github/CONTRIBUTING.md">
```
fairingrey: The changes here are remnants of old commits from #220, you may want to keep this in mind later when fixing the merge conflicts.

(long paths are required so the URLs on crates.io can be fully resolved)

prasannavl: Yeah, will likely have to rebase with master again once it gets merged.

yoshuawuyts:
#220 was merged, this should be ready for a rebase!

Cargo.toml
```diff
@@ -1,7 +1,10 @@
 [workspace]
 members = [
     "tide",
+    "tide-cookies",
+    "tide-core",
     "examples",
+    "tide-log",
```
fairingrey: Could we move this above "examples"?

prasannavl: Should be resolved with latest rebase

tide-log/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+            status.as_str(),
+            start.elapsed().as_millis()
+        );
+        res
+    }
+}
+
+impl<Data: Send + Sync + 'static> Middleware<Data> for RequestLogger {
+    fn handle<'a>(&'a self, ctx: Context<Data>, next: Next<'a, Data>) -> BoxFuture<'a, Response> {
+        box_async! { self.log_basic(ctx, next).await }
```
yoshuawuyts: Nit: we could probably do without this `box_async` here.

tide-log/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+)]
+
+use futures::future::BoxFuture;
+use log::{info, trace};
+use tide::{
+    middleware::{Middleware, Next},
+    Context, Response,
+};
+
+macro_rules! box_async {
```
fairingrey: Following the merging of #245, the `box_async` macro should now be avoided, instead bringing `FutureExt` into scope instead and using it that way. As said here also https://github.com/rustasync/tide/pull/222#discussion_r285783972

Given it's a relatively recent change, but I think with this change we can finish merging it!

prasannavl: Would someone like to pick this up at the moment? (my branch is editable to maintainers)
 (I'm a bit of a rush today, but this and #241 are otherwise ready to merge, I think)

fairingrey: I can wrap it up for you real quick, sure!

prasannavl: @fairingrey awesome! thank you :smiley:

PR #241, opened 2019-05-20
Adjust code quality enforcement

prasannavl:
## Description
- Switch all crate level forbid and deny into warn
- Make code enforcement consistent across all crates

Based on #222 

## Motivation and Context
https://github.com/rustasync/tide/issues/232

## How Has This Been Tested?
All tests passed

## Types of changes

## Checklist:


tide-log/src/lib.rs
```diff
@@ -1,5 +1,7 @@
+#![cfg_attr(feature = "nightly", deny(missing_docs))]
```
Nemo157: I noticed that none of the `Cargo.toml` actually have a `nightly` feature in them. I think this being written like this in other `lib.rs` is just an artifact from an older setup, and it should probably move into the general `warn` section.

tide-log/src/lib.rs
```diff
@@ -1,5 +1,7 @@
...
+#![cfg_attr(test, deny(warnings))]
```
Nemo157: This will still stop developers from being able to run tests locally when their code has warnings in it. I think having the warnings denied when running clippy in CI is enough: https://github.com/rustasync/tide/blob/2878abad4695a9fa85a90818c31eb844363bce35/.travis.yml#L21

prasannavl: Oh, yikes, yeah, this makes no sense. Mechanically copied over the entire header part from tide. Fixing it up

tide-log/src/lib.rs
prasannavl: Removed this. However, adding this on to warn opens up a lot of warning for all the middleware crates. Middleware structs are documented, however not the crates. Suggestions?

prasannavl: @Nemo157 - would you like me to add `missing_docs` to warn, or leave it out for the moment? (Adding it will fail all crates except for `tide` currently)

Nemo157: They should probably have some documentation added, maybe open an issue about filling out the documentation and adding it to each of the crates.

fairingrey:
@prasannavl Fixed the merge conflict for you, merge it whenever you'd like :smile:

PR #246, opened 2019-05-20
Sweep the cache before uploading it

### LINKED ISSUE #243: Travis CI caching for tests is timing out
See for example https://travis-ci.org/rustasync/tide/jobs/534672877#L942-L944

Taking more than 3 minutes probably makes the cache worthless. It's likely that it has accumulated old versions of dependencies that should not be there anymore, I think there are some tools that can cleanup the `target` directory based on what is actually used in the recent builds so that the cache doesn't keep storing unused data.

Nemo157:
## Description
Use [`cargo-sweep`](https://github.com/holmgr/cargo-sweep) along with [`-Zmtime-on-use`](https://github.com/rust-lang/cargo/pull/6573) for removing unused data from the Travis cache.

`cargo-sweep` will remove all files that were not used during the current build, so for example old versions of dependencies. To be able to track this accurately it needs `cargo`/`rustc` to mark the files it uses during the current build, the `-Zmtime-on-use` flag is an unstable flag to tell it to do so.

With this implemented the files uploaded to Travis should only be the relevant ones, keeping the cache minimal and actually usable. This will then mean that future builds will only have to build what has changed and run much faster.

## Motivation and Context
Fixes #243

## Types of changes

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,29 +2,57 @@ language: rust
 rust: nightly-2019-05-09
 cache: cargo
 
+before_script:
+- >
+  [[ "$(cargo-sweep --version)" == "cargo-sweep 0.4.1" ]]
+  || cargo install cargo-sweep
+- cargo sweep --stamp
+
+before_cache:
+- cargo sweep --file
+
 matrix:
   include:
   - name: cargo doc
     env: [CACHE_NAME=docs]
     script:
-    - RUSTDOCFLAGS=-Dwarnings
-      cargo doc --all --all-features --no-deps --exclude tide
+    - RUSTDOCFLAGS=-Dwarnings cargo doc
+        -Zmtime-on-use
+        --all --all-features
+        --exclude tide
+        --no-deps
 
   - name: cargo fmt
     cache: false
-    before_script: rustup component add rustfmt
-    script: cargo fmt --all -- --check
+    before_script: []
+    install:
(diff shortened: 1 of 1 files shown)


Nemo157:
I wish there were an easy way to look at the cache content, the ones from this branch are still half a gig larger than my testing on travis-ci.com and the test job is taking twice as long. But at least comparing https://travis-ci.org/rustasync/tide/builds/535025048 and https://travis-ci.org/rustasync/tide/builds/535036746 the latter used about half the time for the test job.

Once this is merged I might hit the "delete all caches" button on Travis and see if it creates a smaller cache inline with what I was seeing in the future.

.travis.yml
```diff
@@ -2,29 +2,57 @@ language: rust
...
-    - cargo build --manifest-path tide/Cargo.toml --no-default-features
-    - cargo build --manifest-path tide-core/Cargo.toml --no-default-features
+    - cargo build
+        -Zmtime-on-use
+        --manifest-path tide-core/Cargo.toml
+        --no-default-features
+    - cargo build
+        -Zmtime-on-use
+        --manifest-path tide/Cargo.toml
+        --no-default-features
```
Nemo157: https://github.com/rust-lang/cargo/issues/6972 means that this does some extra work, but it's still faster than the `test` build so doesn't really matter at this point.

PR #227, opened 2019-05-15
Reorganize examples and make runnable

prasannavl:
## Description

Make examples runnable with:
`cargo run -p examples --bin <example_name>` (or) 
`cargo run --bin <example_name>` (from examples dir)

Resolves https://github.com/rustasync/tide/issues/223

(Wait until to avoid conflicts: https://github.com/rustasync/tide/pull/220)

## Motivation and Context

https://github.com/rustasync/tide/pull/219#pullrequestreview-236996850
https://github.com/rustasync/tide/pull/219#discussion_r283687631

## How Has This Been Tested?

All examples functional

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Refactor

## Checklist:

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the [CONTRIBUTING](https://github.com/rustasync/tide/blob/master/.github/CONTRIBUTING.md) document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #247, opened 2019-05-21
Bring tide to root, and allow `cargo run --example`

yoshuawuyts:
## Description
Moves the tide submodule to the root of the crate, allows running examples using:
```sh
$ cargo run --example <example>
```
Also makes the examples to be run standalone without a need for `lib.rs`. Ref https://github.com/rustasync/tide/issues/223.

## Motivation and Context
This patch is an alternative to #227. As explained in https://github.com/rustasync/tide/issues/223#issuecomment-494184444, I think it's preferable if we keep Cargo's default behavior of being able to run examples using `cargo run --example`.

I spent some time today trying to figure out how to run our examples, only to find out that they weren't runnable, which was unexpected, and admittedly rather frustrating.

## Types of changes

## Checklist:


prasannavl:
Awesome! I do think this is much nicer, and also keeps things consistent with `runtime`. My initial motivation to not go this route, was that we may have to tweak tooling scaffolds and other experience if porting over from `futures`, `tokio`, etc. 

But yeah, we can take the new path and figure things out as we go :)

fairingrey:
This approach is neat, reminds me of the way actix-web is laid out (with the main crate at root, so we no longer rely on a workspace manifest alone). I like it :+1: 

PS: Thanks for fixing my example!

prasannavl:
If the travis bit can be fixed, I think we can merge this and handle the example clippy warnings post merge, for this particular instance.

fairingrey:
@prasannavl Yup, after these Travis runs finish, clippy should only be yelling at the examples (and we can just get those fixed up in a later PR)

fairingrey:
Clippy freaks out on the examples, but as mentioned before we can fix it later -- merging!

PR #252, opened 2019-05-21
Core revamp part 2

prasannavl:
## Description

- Depend on `tide-core` for `tide-log` instead of full `tide`
- `DefaultHeaders` to new crate `tide-headers`
  - Add trace
  - Remove `inline` attr that's counter intuitive.
- `RootLogger` -> `RequestLogger` in new crate `tide-slog`
  - Minor refactors
  - Add `with_logger` constructor
  - Make behaviour consistent with `tide-log`
- `tide-log`
  - Add `with_target` constructor to configure log `target`
  - Make default `target` as `requests`
- [BREAKING] Make `RequestLogger` from tide be exposed from `tide::middleware`
- [BREAKING] Remove `tide-slog` based logger from `tide::middleware`. 

### Other

- Add examples: `hello_envlog`, `hello_logrs`
- Add `tide` as `dev-dependency` to all crates, since they'll all require documentation at some point
- Minor additional documentation throughout touched surface

### TODO - Core revamp - Part 3

- Extract `form.rs` into `tide-forms`
- Extract `querystring.rs` into `tide-querystring`

### TODO - Core revamp - Part 4

- Move non essentials from `tide-core` into `tide` 

## Motivation and Context
#162 

## How Has This Been Tested?
All tests passed

## Types of changes

- [x] Refactor
## Checklist:


tide-slog/src/lib.rs
```diff
@@ -1,50 +1,70 @@
...
 
-        let log = slog::Logger::root(drain, o!());
-        RootLogger { inner_logger: log }
+    pub fn with_logger(logger: slog::Logger) -> Self {
+        Self { inner: logger }
     }
 }
 
-impl Default for RootLogger {
+impl Default for RequestLogger {
```
Nemo157: I don't think this should be provided at all, IMO if someone is going to the effort of using `slog` that's because they care strongly about their logging and will want full control over the root logger.

It also unconditionally introduces the `slog_term` and `slog_async` dependencies even if a user is going to use a different root logger.

Having a doc example show how to construct this together with `slog_term` etc. seems enough.

prasannavl: Sure, I'm not even sure if `tide-slog` should be a part of tide repo, since it's a highly opinionated option. I just decided to make in a tad bit more configurable than it was previously extracting this out. 

(Hence, removed this from `tide::middleware`)

prasannavl: Also, I personally don't use slog - now that it's out in it's own crate, think it'd be better with someone who actually uses it to pick this up and continue. /cc @bIgBV

tide-log/Cargo.toml
```diff
@@ -18,7 +18,7 @@ repository = "https://github.com/rustasync/tide"
 version = "0.1.0"
 
  [dependencies]
-tide = { path = "../" }
+tide-core = { path = "../tide-core" }
```
Nemo157: `tide-log` will need `tide` added to its `dev-dependencies` for the examples to compile (https://travis-ci.org/rustasync/tide/jobs/535258952#L872-L876)

prasannavl: Added to all crates, since pretty much all of them are going to require it for documentation at some point anyway.

tide-headers/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+authors = [
+    "Tide Developers"
+]
+description = "Header related middleware for tide"
```
fairingrey: I think Tide should be capitalized here.

tide-log/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 authors = [
     "Tide Developers"
 ]
-description = "Tide middleware for logging"
+description = "Logging middleware for tide"
```
fairingrey: Same here.

tide-slog/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+authors = [
+    "Tide Developers"
+]
+description = "Logging middleware for tide based on slog"
```
fairingrey: And here.

prasannavl:
Thanks @fairingrey for fixing up the names. 

Part 3 sitting in here, though not ready yet: https://github.com/rustasync/tide/compare/master...prasannavl:core_revamp_3?expand=1

(Have the macro bit to resolve. Should be able to make some time a little later tonight)

tide-slog/src/lib.rs
bIgBV: I originally added slog for the structured logging it provides. But given that the `log` crate is incorporating structured logging as well, I don't think we need to depend on slog again. We could look into providing support for `tokio-trace` once it stabilizes a bit.

Nemo157: `slog`s main value add for something like a web server is _scoped_ logging. Being able to inject a request id onto the logger context at the start of the request and have that automatically emitted with every log line during that request is super useful.

I have an idea of a project I'm tempted to start writing with Tide, if I do get around to doing that I'll definitely be using `tide-slog` and would be happy to help with core maintainership of it.

Issue #51, opened 2018-11-19
Implement OPTIONS http method


tzilist:
This is a pretty common http method when using javascript's `fetch` api with CORS enabled. I think Tide should probably be able to automatically be able to handle this based on what routes are set and what corresponding methods are available.

fairingrey:
Just a remark on this (and also to the related PR at https://github.com/rustasync/tide/pull/104), it seems that Tide currently stores routes as a HashMap keyed on HTTP method, with values being the routers that contain endpoints with the given HTTP method (and accessed through the URL path).

This is just my idea, but one naive way of adding automatic support for `OPTIONS` is probably to just key on *every* available HTTP endpoint (`GET`/`POST`/`PUT`/`DELETE`/etc.) and see if the route exists at that particular resource. It seems a little inefficient, since in the end we're making a total of at least [9 calls](https://docs.rs/http/0.1.17/http/method/struct.Method.html) as per all of them, but maybe there's a more efficient way of doing this?

Gowee:
Another similar problem involving with the same design choice of routing aforementioned. :

From the perspective of the semantics of HTTP methods, `405 Method Not Allowed` is a better choice over `404 Not Found` when there is an endpoint on a specific route path but request methods are not matched. 

For example, https://github.com/rustasync/tide/blob/279e0f05a0a24fba9fc192f1d7956ad60fa0cf24/examples/messages.rs#L69
registers `POST` method on `/message`. It is supposed to response with `405 Method Not Allowed` HTTP status which clearly indicates that the request method is unexpected but the resource (path) is valid.

This behaviour is adhered by most matured Web framework in other programming languages. But it seems that many of the famous used Web frameworks in Rust prefer to take the contrary way. There would be some performance penalty to be compliant, but I think it is worth discussing.

malte-v:
Any progress on this? It's on the 0.1 milestone, but we're already at 0.6... I'm willing to help out if you point me in the right direction.

yoshuawuyts:
Note from triage: We handle CORS through the [`CorsMiddleware`](https://docs.rs/tide/0.14.0/tide/security/struct.CorsMiddleware.html); `OPTIONS` is handled inside the Tide router. We should add docs for this though.

We're unsure however if simplify specifying the middleware also setups the options endpoint; it's been a while. We should make it so that we can enable that. One way to get there is by enabling middleware to run _before_ the router, which is something we may want to do for other reasons as well.

PR #254, opened 2019-05-21
Fix the documentation for wildcards for path definitions to match implementation

gameldar:
Fixed the documentation for wildcards for path definitions for routing and added a bunch of tests that demonstrate the implemented behavior.

## Description
The documentation referred to syntax like :name* but this isn't actually supported by the route recognizer, instead it is looking for *name for named path matches. So I've fixed the explanation and given more examples.

Additionally I added a number of tests for these different path options.

## Motivation and Context
I was starting out with Tide and was trying to get some routes working - in particular capturing the rest of the path and so headed to the documentation to find the suggested wildcard of :path* which should match. Trying to use this caused issues with not being able to find the parameter back (as it turns out it is because the * gets added to the name so you can use context.param("path*") to get the value back - but it only matches the next segment).

Looking through blame I tracked it back to the underlying issue - #12  and I can see that isn't resolved/finalized yet. Looking at some of the behaviour that is possible (partly why the tests are there) - I can see that there might be some undesirable routing behaviour (e.g. just using `:` to namelessly match) might be an issue.  

## How Has This Been Tested?
I've run the tests after these changes and the extra tests I'ved added. I'm testing on Debian Testing using the nightly version of Rust from 2019-05-21. 

## Types of changes

## Checklist:

### DIFF
--- a/tests/wildcard.rs
+++ b/tests/wildcard.rs
@@ -10,6 +10,22 @@ async fn add_one(cx: Context<()>) -> Result<String, tide::Error> {
     Ok((num + 1).to_string())
 }
 
+async fn add_two(cx: Context<()>) -> Result<String, tide::Error> {
+    let one: i64 = cx.param("one").client_err()?;
+    let two: i64 = cx.param("two").client_err()?;
+    Ok((one + two).to_string())
+}
+
+async fn echo_path(cx: Context<()>) -> Result<String, tide::Error> {
+    let path: String = cx.param("path").client_err()?;
+    Ok(path)
+}
+
+async fn echo_empty(cx: Context<()>) -> Result<String, tide::Error> {
+    let path: String = cx.param("").client_err()?;
+    Ok(path)
+}
+
 #[test]
 fn wildcard() {
     let mut app = tide::App::new();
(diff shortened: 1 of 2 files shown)


fairingrey:
We ought to resolve #12 sometime in the future if and or when we acquire ownership of the route-recognizer crate or update it with something more flexible (and that we have control over, currently) like [path-table](https://github.com/rustasync/path-table). I'll mention it in that issue too, but wildcard behavior over dynamic paths/segments is something that's better resolved sooner than later, + taking another look at how we can design the router will also help with issues like #51.

Thanks for your contribution! I'll let one more person review it before a potential merge, but I like the changes.

technetos:
There are a lot of use cases for tide and IIRC route-recognizer provides a mechanism for changing the priority of routes, though it may not be exposed publicly.  My point is that there isnt a one size fits all route match priority and it would make sense to expose a mechanism in tide to change the priority so the most possible use cases can be supported.

prasannavl:
LGTM! Resolving the complete routing story will probably take a bit.
Thank you for the PR, @gameldar!

Issue #263, opened 2019-05-27
Dealing with panics


Nemo157:
Currently Tide doesn't provide any way to deal with panics in handlers, these just unwind up to the executor and (at least in the case of Tokio) kill the executors thread. During the unwinding the TCP connection is dropped and the client gets something like `The connection was reset`.

My first thought on dealing with this is having a middleware that uses `catch_unwind` to catch the panic and return a minimal 500 to the client. (I have prototyped this ad it works very easily).

The other part of this is that when I encountered it I had locked a `Mutex` stored in the `State`, this subsequently started returning `PoisonError` on access so my application was effectively dead. Maybe when a panic is encountered it should be proxied through to the future returned by `serve` and resumed on the main thread to try and kill the entire application. Alternatively this could be dealt with at the application level using "health checks" with support from a service coordinator (like Kubernetes). Since it depends on usecase it may make sense to provide middleware supporting different options and allow the user to choose.

yoshuawuyts:
@Nemo157 I like the idea of doing a basic thing where we terminate all connections with a 500. I'm a big fan of the panic = abort model, where if we're panicking it means something has gone terminally wrong, and we need to reboot.

I'm a bit hesitant for any more elaborate schemes though. Especially if there's a risk we might hang while we try and return the `500`s (e.g. network timeouts). When a panic occurs we should want to terminate, and let the supervisor restart the application. If panics occur frequently supervisors (e.g. kubernetes) can issue an auto-rollback to give developers time to investigate the failure without it affecting production.

Re: poison errors -- [parking-lot](https://github.com/Amanieu/parking_lot) provides a way around poison errors specifically, but I'm not sure what to do about {dead,live}locks in general. I think your intuition is right, and we should have a way to provide status checks to any supervisor process so they can judge the status. But this indeed feels like a different topic than "dealing with panics". Perhaps we should open a separate issue on this?

Nemo157:
I brought up the poison error because that is a knockon effect from panic unwinding. I don't know if there are other commonly used data structures that you might store in `State` which become unusable during unwinding, but if there are then it seems like a good argument for the "proxy panics through to `App::{serve,run}`" approach where a panic in any handler kills off the entire application (which can't be dealt with just as a middleware currently). It sounds like you're in favour of this approach rather than the basic "panic fails the current request, but the rest keep running" approach I prototyped.

(Opened https://github.com/rustasync/tide/issues/264 for more specifics on {dead,live}locked applications).

[...]

PR #265, opened 2019-05-27
Add middleware for catching panics

Nemo157:
## Motivation and Context
cc #263, @prasannavl, @yoshuawuyts 

## Types of changes

## Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,6 +60,7 @@ members = [
     "tide-log",
     "tide-querystring",
     "tide-slog",
+    "tide-panic",
 ]
 
 [patch.crates-io]
--- a/tide-panic/Cargo.toml
+++ b/tide-panic/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "tide-panic"
+version = "0.1.0"
+edition = "2018"
+authors = ["Tide Developers"]
+description = "Advanced panic support for Tide"
+documentation = "https://docs.rs/tide-panic"
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/rustasync/tide"
+
+[dependencies]
+futures-preview = "0.3.0-alpha.16"
+http = "0.1"
+http-service = "0.2.0"
+tide-core = { path = "../tide-core" }
--- a/tide-panic/LICENSE-APACHE
+++ b/tide-panic/LICENSE-APACHE
@@ -1,0 +1,1 @@
+../LICENSE-APACHE
--- a/tide-panic/LICENSE-MIT
+++ b/tide-panic/LICENSE-MIT
@@ -1,0 +1,1 @@
+../LICENSE-MIT
--- a/tide-panic/src/catch_unwind.rs
+++ b/tide-panic/src/catch_unwind.rs
@@ -1,0 +1,65 @@
+use futures::future::{BoxFuture, FutureExt, TryFutureExt};
+use http::status::StatusCode;
+use std::{
+    any::Any,
+    panic::{AssertUnwindSafe, RefUnwindSafe},
+};
+use tide_core::{
+    middleware::{Middleware, Next},
+    response::IntoResponse,
+    Context, Response,
+};
+
+/// A [`Middleware`] that will catch any panics from later middleware or handlers and return a
+/// response to the client.
+///
+/// It is **not** recommended to use this middleware for a general try/catch mechanism. The
+/// [`Result`] type is more appropriate to use for middleware/handlers that can fail on a regular
+/// basis.  Additionally, this middleware is not guaranteed to catch all panics, see the "Notes"
+/// section in the [`std::panic::catch_unwind`] docs.
+pub struct CatchUnwind {
+    f: Box<dyn Fn(Box<dyn Any + Send + 'static>) -> Response + Send + Sync>,
+}
+
+impl CatchUnwind {
+    /// Create a [`CatchUnwind`] which will respond with [`StatusCode::INTERNAL_SERVER_ERROR`] when
+    /// any panic is caught.
+    pub fn new() -> Self {
+        Self::with_response(|_| {
+            "Internal server error"
+                .with_status(StatusCode::INTERNAL_SERVER_ERROR)
+                .into_response()
+        })
+    }
+
+    /// Create a [`CatchUnwind`] with a custom function to generate the response, the function will
+    /// be passed the caught panic.
+    pub fn with_response(
+        response: impl Fn(Box<dyn Any + Send + 'static>) -> Response + Send + Sync + 'static,
+    ) -> Self {
+        Self {
+            f: Box::new(response),
+        }
+    }
+}
+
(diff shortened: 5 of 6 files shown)


Nemo157:
One thing I'm unsure about post restructure is where relatively trivial middleware like this should go. Is it big enough to be worth its own crate? If not, should it be in the main `tide` crate like it is currently, or somewhere else?

prasannavl:
@Nemo157 Nice! I was also just discussing this with @yoshuawuyts and I'm wondering if there's any common case that actually cause a _non-unwindable_ panic outside of interop situations. 

If this is sufficient, there's probably no need to think about future executors or resurrecting worker threads for now. 

Another question: If it isn't desired for this to be caught, should we still catch them and propagate them to tide somehow as an error and initiate a (clean) shutdown instead of just crashing? (This could be a another middleware in the same crate or a config, may be?)

---

Thinking out loud here, on the `where it goes` bit:

- I think we're probably going to have some 10-20 middlewares. And most of trivial ones will likely fall in some category - `log`, `headers`, `auth`, `cookies`, `errors` etc.
- I think over time we can probably logically group them into one of the existing category. For instance today we have `DefaultHeaders` in `tide-headers`, we probably could have other trivial bits that have to do with headers in `tide-headers` -- say `ForwardHeaders` for adding support for proxy forwarded headers - just separated by feature flags. (or does this affect discoverability?)
- Personally, I think we generally should avoid putting things in `tide` directly, as that defeats the purpose of splitting things up. 
- Anything that doesn't fit in, we probably could just let it be outside in personal repos.
- If absolutely needed, may be add `tide-extras` and keep things in there?

That said, I personally think, error handling could be in it's own logical category and hence go into a crate.

Nemo157:
> I'm wondering if there're common cases that cause non-unwindable panic outside of alloc and interop situations.
>
> If this is sufficient, there's probably no need to think about future executors or resurrecting worker threads for now.

As far as I'm aware running out of memory and most panic-inside-FFI situations abort the _process_, so there's no way that Tide/the executor could recover. 

> Another question: If it isn't desired for this to be caught, should we still catch them and propagate them to tide somehow as an error and initiate a (partially clean?) shutdown instead of just crashing? (This could be a another middleware in the same crate or a config, may be?)

I have another branch adding a middleware for something like that along with examples for both this middleware and the one it adds: https://github.com/Nemo157/tide/compare/catch-unwind...propagate-unwind (I haven't included it in this PR yet because I'm not certain whether it should `resume_unwind` like it does, or return the panic as an error, and I've noticed some weird behaviour of Tokio when panicking on the "main" task I want to investigate for the example).

> That said, I personally think, error handling could be in it's own logical category and hence go into a crate.

How does `tide-panic` sound?

prasannavl:
> As far as I'm aware running out of memory and most panic-inside-FFI situations abort the process, so there's no way that Tide/the executor could recover.

I'm wondering if there are any other common cases which might abort the process directly beyond these two. However, alloc can also be recovered by this: https://github.com/rust-lang/rust/issues/43596

Currently, this works:

```rust
#![feature(alloc_error_hook)]

use std::alloc::set_alloc_error_hook;
use std::alloc::Layout;

fn handle_alloc_error(layout: Layout) {
    panic!("alloc: failed to allocate {} bytes", layout.size());
}

fn main() {
    set_alloc_error_hook(handle_alloc_error);
    let result = std::panic::catch_unwind(|| {
        let v = Vec::<u8>::with_capacity(100000000000000);
        println!("{:p}", &v[..]);
    });
    println!("{:?}", result);
}
```

But interestingly, looks like drops don't work properly with release optimisations:

```rust
#![feature(alloc_error_hook)]

use std::alloc::set_alloc_error_hook;
use std::alloc::Layout;

fn handle_alloc_error(layout: Layout) {
    panic!("alloc: failed to allocate {} bytes", layout.size());
}

fn main() {
    set_alloc_error_hook(handle_alloc_error);
    let result = std::panic::catch_unwind(|| {
        // Let's do some non sense to try to make sure the compiler
        // doesn't optimize these away.
        let mut r = Resource { ptr: 0 as *mut usize };
        let v = Vec::<u8>::with_capacity(100000000000000);
        println!("{:p}", &v[..]);
        println!("{:p}", r.ptr);
        r.ptr = 1 as *mut usize;
        println!("{:p}", r.ptr);
        
    });
    println!("{:?}", result);
}

struct Resource { ptr: *mut usize }

impl Drop for Resource {
    fn drop(&mut self) {
        println!("dropping resource: {:p}", self.ptr);
    }
}
```

- So, likely not a viable solution at this stage.

And `tide-panic` sounds great! - though I'll let others chime in on this.

src/catch_unwind.rs
```diff
@@ -0,0 +1,57 @@
...
+        }
+    }
+}
+
+impl<State: 'static> Middleware<State> for CatchUnwind {
+    fn handle<'a>(&'a self, cx: Context<State>, next: Next<'a, State>) -> BoxFuture<'a, Response> {
+        std::panic::AssertUnwindSafe(next.run(cx))
+            .catch_unwind()
+            .unwrap_or_else(move |err| (self.f)(err))
+            .boxed()
```
yoshuawuyts: This only returns 500 for a single request, right? My initial understanding was that _all_ requests would be terminated with 500, and then the `panic` would continue. But I guess the intended behavior here is different?

prasannavl: Is there a need to terminate all requests? panic on the request path doesn't need to prevent other working code paths, as long as it's safely unwindable?

Nemo157: Yes, this is the “I think my application will continue working if a single request panics” middleware. See the link in my other comment to the “fail-fast” version I haven’t finished yet.

yoshuawuyts: I was talking with the other Runtime maintainers yesterday, and if a future breaks inside of runtime somewhere (e.g. `runtime::spawn`), it'll default to terminating its thread, and we have no good way to restore that thread. There might not even be one tbh. I believe Tokio adopted a similar model recently too (on request of the Deno project).

If Tide starts catching panics rather than propagating them, we will start losing threads without having a way to spin them back up; which will severely limit throughput without failing the process entirely. Which seems like not a great situation to be in. I feel this would be particularly hard to debug when not intimately familiar with all moving parts up and down the stack (and even then it's not _easy_).

I guess what I'm saying is that this approach is making me feel quite nervous, and perhaps we should start with https://github.com/Nemo157/tide/compare/catch-unwind...propagate-unwind first?

Nemo157:
Note from chat: make documentation more explicit that using this can lead to other issues and that you really need to understand how unwinding could affect the rest of your code, and investigate whether it should require a bound like `State: (Ref)UnwindSafe` to make it more explicit.

Nemo157:
I've moved this to a new `tide-panic` crate and extended the documentation. I've added a `State: RefUnwindSafe` bound for now, since the panic may occur while there is an `&State` accessible, and all other data from the context should be request local.

Issue #207, opened 2019-05-10
Is it possible to return an HTML response?


k-nasa:
## Question

How do I implement it to return an HTML response?
I checked about response, but "content-type: text / html"
 There is no applicable place in. Also, there is no processing to return HTML in example.
https://github.com/rustasync/tide/blob/7e87e4e0699a3fff4453c8a25d3f8f5fa186fdbb/src/response.rs

However, I think there are many such use cases.

Here is a suggestion, but I think we should implement a response function that returns HTML, and I think it would be nice if there was something to easily change the response header.
And should add an example to return an HTML response

yoshuawuyts:
@k-nasa we've talked about HTML templates before, building a module similar to what we're envisioning for authentication and DB connections.

Right now it's possible to create your own HTML responses with a bit of glue code. I made an example of this late last year: https://github.com/yoshuawuyts/playground-html-rust

Hope this is helpful!

k-nasa:
@yoshuawuyts Thanks! 
I will try to add an example if this [PR](#227 ) is merged.

I wonder what kind of things you are trying to add. Where is the information?

k-nasa:
Closed by #267

PR #267, opened 2019-05-31
Add templating example with tera

milesgranger:
Presently, there are no examples demonstrating rendering HTML; this PR addresses that.

## Description

Added an example using [tera](https://tera.netlify.com/) which renders some variables into a template.

## Motivation and Context

Conveys a concise way to render a HTML response with Tide.

Relates to #207, and might even close it? 

## How Has This Been Tested?

Verified functionality locally, and passes tests

## Types of changes

## Checklist:

### DIFF
--- a/examples/templating_tera.rs
+++ b/examples/templating_tera.rs
@@ -1,0 +1,53 @@
+#![feature(async_await)]
+
+use tera::{self, compile_templates};
+use tide::{self, App, Context, EndpointResult, Error};
+
+// AppState to pass with context and will hold
+// the interface to the tera rendering engine
+struct AppState {
+    template: tera::Tera,
+}
+
+// Render some data into the 'tera-hello-world.html template in examples/templates directory
+async fn index(ctx: Context<AppState>) -> EndpointResult {
+    // Create the context for the template
+    let mut context = tera::Context::new();
+    context.insert("page_title", "Hello from Tera templating!");
+    context.insert("points", &vec!["point1", "point2"]);
+
+    // Render the variables into the template
+    let s = ctx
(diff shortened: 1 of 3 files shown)


examples/templating_tera.rs
```diff
@@ -0,0 +1,44 @@
...
+    let mut context = tera::Context::new();
+    context.insert("page_title", "Hello from Tera templating!");
+    context.insert("points", &vec!["point1", "point2"]);
+
+    // Render the variables into the template
+    let s = ctx
+        .state()
+        .template
+        .render("tera-hello-world.html", &context)
+        .expect("Failed to render!");
```
yoshuawuyts: Could we perhaps use `?` here instead? -- `.expect` is unlikely to be used for production applications.

examples/templating_tera.rs
```diff
@@ -0,0 +1,44 @@
...
+
+fn main() {
+    let template_dir = env!("CARGO_MANIFEST_DIR");
+    let state = AppState {
+        template: compile_templates!(&format!("{}/examples/templates/*", template_dir)),
+    };
+
+    let mut app = App::with_state(state);
+    app.at("/").get(index);
+    app.run("127.0.0.1:8000").unwrap();
```
yoshuawuyts: Similar here -- could we return `Result` from main?

milesgranger:
My pleasure! I've made the updates, per your request. Guess I've gotten into the habit of doing `unwrap`s and `expect`s for demo / examples, but I like your way of thinking; should set a _good_ example. :tophat: 

Please let me know if there is anything else I should update.

PR #273, opened 2019-06-07
tide-test

yoshuawuyts:
## Description
Adds a testing framework for Tide, with an API similar to Node's supertest.

## Motivation and Context
This should make it significantly easier to test Tide applications, allowing the authoring of request-like APIs that should be both easy to read, and convenient to write.

## How Has This Been Tested?
It hasn't. WIP.

## Types of changes

## Checklist:

## Example
```rust
let mut app = tide::App::new();
app.at("/hello").get(async move |_| "Hello, world!");

let mut app = app.test();
let res = app.get("/hello")
    .send()
    .await?;

let body = res
    .assert_status(200)
    .body_string()
    .await?;

assert_eq!(body, String::from("Hello, world!"));
```


src/test.rs
```diff
@@ -0,0 +1,163 @@
...
+        req.method(method);
+        req.uri(uri);
+        Self { app, req }
+    }
+
+    /// Submit the request.
+    pub async fn send(
+        mut self,
+    ) -> Result<HttpTestResponse, Box<dyn Error + Send + Sync + 'static>> {
+        let req = self.req.body(http_service::Body::empty())?;
```
bIgBV: I'm guessing in the next iteration, you would be able to pass in your own request body, like a JSON blob?

yoshuawuyts: Yeah, that's the idea!

Issue #278, opened 2019-06-13
Tide-cookies fails to compile with Rust Nightly newer than 2019-06-01


gameldar:
Largely this is here for a tracking bug of the issue within the rustc - see https://github.com/rust-lang/rust/issues/61579

but it affects building master currently.

## Bug Report

Building the master branch of Tide with a nightly newer than 2019-06-01 results in the following error:
```
error: internal compiler error: src/librustc_mir/transform/generator.rs:540: Broken MIR: generator contains type std::option::Option<data::CookieData> in MIR, but typeck only knows about for<'r> {data::CookieData, std::sync::Arc<std::sync::RwLock<cookie::jar::CookieJar>>, std::pin::Pin<std::boxed::Box<(dyn core::future::future::Future<Output = http::response::Response<http_service::Body>> + std::marker::Send + 'r)>>, ()}
  --> tide-cookies/src/middleware.rs:39:37
   |
39 |           FutureExt::boxed(async move {
   |  _____________________________________^
40 | |             let cookie_data = cx
41 | |                 .extensions_mut()
42 | |                 .remove()
...  |
63 | |             res
64 | |         })
   | |_________^

thread 'rustc' panicked at 'Box<Any>', src/librustc_errors/lib.rs:572:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
error: aborting due to previous error


note: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/rust-lang/rust/blob/master/CONTRIBUTING.md#bug-reports

note: rustc 1.37.0-nightly (400b409ef 2019-06-09) running on x86_64-unknown-linux-gnu

note: compiler flags: -C debuginfo=2 -C incremental --crate-type lib

note: some of the compiler flags provided by cargo are hidden

error: Could not compile `tide-cookies`
```
#### Environment
- Rust toolchain version(s): [e.g. nightly-]
- OS: Linux 4.19.0-5-amd64 SMP Debian 4.19.37-3 (2019-05-15) x86_64 GNU/Linux

tzsword:
i have same issue

```
[...]

PR #279, opened 2019-06-19
Workaround for issue #278: works around Rust internal compiler error.

realcr:
## Description

Solves problem mentioned in issue #278. 
Solution according to: https://github.com/rust-lang/rust/issues/61579#issuecomment-500436524

## Motivation and Context

`tide` can not be compiled after nightly 2019-06-01. However, another rustc bug doesn't allow me to compile my project's code for rustc of version lower than 2019-06-15. Therefore I currently have no possible way of compiling my project.

## How Has This Been Tested?

`cargo test`

## Types of changes

## Checklist:

### DIFF
--- a/tide-cookies/src/middleware.rs
+++ b/tide-cookies/src/middleware.rs
@@ -44,7 +44,6 @@ impl<State: Send + Sync + 'static> Middleware<State> for CookiesMiddleware {
 
             let cookie_jar = cookie_data.content.clone();
 
-
             // The `let _ = ...` is a workaround for issue: https://github.com/rustasync/tide/issues/278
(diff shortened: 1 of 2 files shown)


tide-cookies/src/middleware.rs
```diff
@@ -44,7 +44,7 @@ impl<State: Send + Sync + 'static> Middleware<State> for CookiesMiddleware {
 
             let cookie_jar = cookie_data.content.clone();
 
-            cx.extensions_mut().insert(cookie_data);
+            let _ = cx.extensions_mut().insert(cookie_data);
```
tirr-c: It would be nice if we have a comment here saying this is a workaround.

realcr: Done, waiting for the CI.

PR #288, opened 2019-07-09
Add async_closure feature gate where needed

spastorino:
This PR adds async_closure feature gate to the crates that use the feature, in tests and example to make everything compile and work properly again.

This is related to the recent change done in https://github.com/rust-lang/rust/pull/62292

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: rust
-rust: nightly-2019-06-02
+rust: nightly-2019-07-08
 cache: cargo
 
 before_script:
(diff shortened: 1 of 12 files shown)


fitzgen:
Care to publish a new version when you have a chance? Happy to use a git dependency in the meantime, so no rush :)

taiki-e:
I think we probably need to release a new version of `http-service*` and update version of them (and get rid of the `patch` field from `Cargo.toml`) before releasing a new version of `tide`. because the current tide's master branch depends on `http-service`'s master branch, not `http-service`, which has been released.

https://github.com/rustasync/tide/blob/36c15fd4ca67fcb269e7bc87d2ceb6eeb8826483/Cargo.toml#L77-L80

PR #291, opened 2019-07-18
Add doc example showing query parameter extraction

phansch:
## Description

This adds a doc test that shows how to parse the URL query into a `HashMap`.

## Motivation and Context

I found this not very well documented and hope the docs will save some time.

## Types of changes

## Checklist:

### DIFF
--- a/tide-querystring/src/lib.rs
+++ b/tide-querystring/src/lib.rs
@@ -14,6 +14,23 @@ use serde::Deserialize;
 use tide_core::{error::Error, Context};
 
 /// An extension trait for `Context`, providing query string deserialization.
+///
+/// # Example
+///
+/// Turning the query parameters into a `HashMap`:
+///
+/// ```
+/// #![feature(async_await)]
+///
+/// # use std::collections::HashMap;
+/// use tide::querystring::ContextExt;
+///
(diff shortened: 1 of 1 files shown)


phansch:
Thinking a bit more about this, I believe my main problem was initially that `ContextExt`/`querystring` were not easily discoverable in the docs; This PR doesn't really address that.

Would you be fine with also adding this example to the docs in `App`?

secretfader:
Thanks for including an example. I wrote this crate as Tide was just beginning to stabilize, with the goal of coming back and adding doctests later. 

As you noticed, that never happened! 

I agree that it's reasonable to include a link to this example in docs for App, but I'm unsure about how to best proceed, since I'm not an expert on rustdoc. 

As is, this PR already adds significant value. Maybe we open an issue on the topic of docs and interlinking? I'm willing to bet we have a bunch of cases similar to this one, across the Tide ecosystem.

phansch:
> Maybe we open an issue on the topic of docs and interlinking? 

Sounds like the best approach to me :+1:

secretfader:
Cool. I'll go ahead and merge this PR. Thanks for your help! :+1:

PR #292, opened 2019-07-18
Create a new method 'html' on response

gameldar:
## Description
Utility method for creating a HTML response from anything that can be converted into a Vec<u8>

## Motivation and Context
Making use of tide for quick little projects I realised it didn't have an easy way to return HTML which is a common use case for web servers (although perhaps diminishing these days...). Given there was a utility method for json it made sense to add a simple utility method for html as well.

## How Has This Been Tested?
I've added an example hello_html that makes use of the method. 

## Types of changes

## Checklist:

- [x ] I have read the [CONTRIBUTING](https://github.com/rustasync/tide/blob/master/.github/CONTRIBUTING.md) document.
- [x ] All new and existing tests passed.


secretfader:
I have a very similar helper in my Tide applications. :+1:

mendelt:
I think this can be closed. The code this pr changes has moved to `http-types`.

Do we want a `from_html` on the `http-types::Body` type? I could add a feature request + a pr there.

Fishrock123:
It strikes me as odd that `.json` would take a structure but that `.html` would take essentially a string. I'm not really sure what that structure would be either though, nor am I familiar with how html is most often used in Rust.

mendelt:
There are some libraries that provide html representations (mostly on the wasm side) for rust. But I don't think there really is something that is so generally used as serde + serde_json on the json side. Probably also because there is no single well defined standard (.xhtml?) and browsers are very accepting in what they will parse.
Html you'll find in server-side rust is probably output from a template engine and those mostly deal with strings. So I agree that stringly typed html is unfortunate but it's probably the most practical way of dealing with it.

PR #315, opened 2019-08-22
export structure Route

### LINKED ISSUE #304: Route struct not exported?
## Question
Where is the Route struct exported in master? On v0.2.0 its at tide::Route.

kamyuentse:
## Motivation and Context

Export `Route` allow mounting endpoint in function, not just in the closure.

## How Has This Been Tested?

Test case added.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

Close #304


src/router/route.rs
```diff
@@ -17,7 +17,7 @@ pub struct Route<'a, State> {
 }
 
 impl<'a, State: 'static> Route<'a, State> {
-    pub fn new(router: &'a mut Router<State>, path: String) -> Self {
+    pub(crate) fn new(router: &'a mut Router<State>, path: String) -> Self {
```
k-nasa: Can I ask why we made this change?

kamyuentse: IIRC, crate user can not create or get a ref of a router, expose `fn new(...)` here is meanless.

k-nasa: thanks

Issue #316, opened 2019-08-23
Why get_cookie take a mutable reference ?


langbamit:
## Question
https://github.com/rustasync/tide/blob/c41ed8407da13bd2db77af9357befb0b334f344a/tide-cookies/src/data.rs#L32

My English is bad. Sorry for this
## Additional context

I want extract authentication token from header or cookie and I suprise when get_cookie take a `&mut self` instead of `&self` like `headers()` function

kamyuentse:
Change to `get_cookie(&self, name: &str)` is reasonable. Indeed, `set_cookie` and `remove_cookie` are not require `&mut self` for current implementation.

secretfader:
Closed by #317.

PR #317, opened 2019-08-23
cookies: is &mut self required?

secretfader:
`&mut self` doesn't seem to be required, as mentioned in #316.

## Description

Change cookie trait function signature on `get_cookie` from `&mut self` to `&self`.

## Motivation and Context

Tide seems to compile fine without requiring a mutable reference. Unsure why the previous behavior may have been required, but I believe there's little need to require mutability when reading cookie values.

## How Has This Been Tested?

Existing tests passed.

## Types of changes

## Checklist:

### DIFF
--- a/examples/cookies.rs
+++ b/examples/cookies.rs
@@ -2,7 +2,7 @@ use cookie::Cookie;
 use tide::{cookies::ContextExt, middleware::CookiesMiddleware, Context};
 
 /// Tide will use the the `Cookies`'s `Extract` implementation to build this parameter.
-async fn retrieve_cookie(mut cx: Context<()>) -> String {
+async fn retrieve_cookie(cx: Context<()>) -> String {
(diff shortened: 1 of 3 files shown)


kamyuentse:
I think we should keep`&mut self` for `set|remove_cookie`.

secretfader:
Ping @rustasync/maintainers

secretfader:
This would be a breaking change, but I agree with the assertion in #316: that `&mut self` doesn't make sense here.

secretfader:
I'm slightly concerned with merging this breaking API change, even though I believe it to be technically more correct. As a consumer of Tide and owner of several applications that will be broken when this change is released, and believe the update is worth applying.

I'll go ahead and merge since I'm no longer the only one frustrated (see #316, where the concern is similar to mine: building authentication middleware). Feel free to comment (or revert) if you feel strongly about requiring `mut`. I think dropping the requirement will make extracting data much cleaner.

PR #321, opened 2019-09-05
feat: if json Serialize failure return 500

fundon:
## Description
If JSON Serialize failure return 500 and Error message.

## Types of changes

## Checklist:

### DIFF
--- a/tide-core/src/response.rs
+++ b/tide-core/src/response.rs
@@ -7,12 +7,19 @@ pub type Response = http_service::Response;
 
 /// Serialize `t` into a JSON-encoded response.
 pub fn json<T: serde::Serialize>(t: T) -> Response {
-    // TODO: remove the `unwrap`
-    http::Response::builder()
-        .status(http::status::StatusCode::OK)
-        .header("Content-Type", "application/json")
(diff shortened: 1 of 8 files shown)


tide-core/src/response.rs
```diff
@@ -7,12 +7,17 @@ pub type Response = http_service::Response;
...
-        .unwrap()
+    let mut res = http::Response::builder();
+
+    match serde_json::to_vec(&t) {
+        Ok(v) => res
+            .header("Content-Type", "application/json")
+            .body(Body::from(v)),
+        Err(e) => res
+            .status(http::status::StatusCode::INTERNAL_SERVER_ERROR)
+            .body(Body::from(e.to_string())),
```
yoshuawuyts: We probably shouldn't leak details to the client here; instead logging using `log::error` and returning an empty body seems like a better option (:

fundon:
CI's network problem.

secretfader:
Looks to me like the requested changes were made (I agree, we don't want to expose the reason serialization failed, by default). I'd like to see this merged into `tide-core`, perhaps with a plan to allow custom error handling in a later PR.

PR #331, opened 2019-10-24
Nested router example

ibaryshnikov:
Added nested router example. Similar example was here before: https://github.com/rustasync/tide/commit/4bb3d3f19861d567ad319372d4ce93a7c04a7970, but disappeared during refactoring.


secretfader:
LGTM. :+1:

PR #334, opened 2019-10-28
Remove dev dependency on the runtime crate

### LINKED ISSUE #333: Remove example that relies on runtime crate
Tide's repository currently [hosts an example](https://github.com/rustasync/tide/blob/master/examples/runtime.rs) that utilizes the [`runtime`](https://github.com/rustasync/runtime) crate. While `runtime` presents a easy-to-understand async programming environment, I don't think it should be a required dependency of Tide (either in development or production).

I believe Tide's dependency on `runtime` is slowing completion of PRs like #307, which updates `tokio` to version `0.2`. Having stale PRs on a project that is owned by a working group isn't the best look; I understand why [users and contributors are beginning to pose questions concerning Tide's future](https://github.com/rustasync/tide/issues/325).

I plan on opening a PR to support this issue, which can be merged by any maintainers who agree with this proposal. cc: @rustasync/maintainers

secretfader:
Resolves #333 by removing the development dependency on `runtime`.


secretfader:
> a separate PR to add the `http-service-hyper` example would be great.

I'll PR a new example shortly.

Issue #336, opened 2019-10-30
Tide development plan


yoshuawuyts:
As a follow-up to https://github.com/http-rs/tide/issues/325, I'd like to share the plan of how we can get Tide to a publishable state. This is not so much an issue to ask for input, but more to share what the plan is in case people are confused by what's going on.

The core problem I think we currently have is one of my design: we have too many different crates, which makes it hard to publish & manage. In contrast: `async-std` has all functionality as part of a single repo, and adds new features behind the `"unstable"` feature flag. I think we __should consolidate Tide into a single crate, and introduce an `"unstable"` feature flag__.

I think we should very much do the same, and the easiest way to get there is by resetting the repo by a few dozen commits to 4a538908f928aa1108473d1832073fc5cb799cc6, the commit right before all the "split" PRs landed.

_Note that this was fantastic work, implemented by @prasannavl. But I think this wasn't the right direction, which is entirely on me._

From there we can move back through the commit log and start applying patches that have landed before. I estimate this to be a few uninterrupted days worth of work.

---

From there the plan is less clear. But I have some idea of API changes I'd like to make. But before we do that we should probably create a new "snapshot" release so people can work off the latest master branch.

I'd like to spend some effort during all this to write docs. We don't really have much right now, and I think that's holding us back. Unsure about the exact timeline, but it's something I think needs to be prioritized.

## This all sounds cool, how can I help?

Unfortunately I think there's not much that can be done right now. The only separate issue that's coming up is https://github.com/http-rs/http-service/issues/31; so if you want to pick this up and help port `http-service` that would be fantastic. Please note though that I won't have any bandwidth to mentor this, and it won't be easy.

Other than that this is mostly me effectively taking a lock on the project to move things forward. Once that's over I should have a better idea of which steps to take after, and tracking issues can be written. That's all further out still though. More updates in the future!

## Conclusion

This is the current work plan I'm thinking of executing on. This means some big changes, but the path seems clear.

secretfader:
> Other than that this is mostly me effectively taking a lock on the project to move things forward. Once that's over I should have a better idea of which steps to take after, and tracking issues can be written. That's all further out still though. More updates in the future!

^ Why Tide is broken, in a nutshell. A maintainer who has done little work in the past few months returns to save the project!

yoshuawuyts:
>  A maintainer who has done little work in the past few months returns to save the project!

[...]

PR #339, opened 2019-10-31
Prepare 0.3.0 release

yoshuawuyts:
This prepares the 0.3.0 release, releasing all the work since the `0.2.0` release up until we started work on splitting the repo into multiple components. This is a snapshot release, and is intended to set us up to rewind the work and improve APIs as outline in #336. Thanks!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,9 +1,11 @@
 [package]
+name = "tide"
+version = "0.4.0"
+description = "Serve the web – HTTP server framework"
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -1,9 +1,11 @@
 [package]
+name = "tide"
+version = "0.4.0"
```
taiki-e:
```suggestion
version = "0.3.0"
```

PR #345, opened 2019-11-03
Rewind page 1

yoshuawuyts:
Rewinds page 1 of https://github.com/http-rs/tide/issues/336. Thanks!


src/cookies.rs
```diff
@@ -30,7 +30,7 @@ impl CookieData {
 /// An extension to `Context` that provides cached access to cookies
 pub trait ContextExt {
     /// returns a `Cookie` by name of the cookie
-    fn get_cookie(&mut self, name: &str) -> Result<Option<Cookie<'static>>, StringError>;
+    fn get_cookie(self, name: &str) -> Result<Option<Cookie<'static>>, StringError>;
```
langbamit: I think it should be `&self` instead

yoshuawuyts: Good catch, thank you!

PR #347, opened 2019-11-05
Don't print anything by default

acasajus:
println! should be avoided so users can decide what to print :)

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -239,8 +239,6 @@ impl<State: Send + Sync + 'static> App<State> {
             .to_socket_addrs()?
             .next()
             .ok_or(std::io::ErrorKind::InvalidInput)?;
-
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
@acasajus I don't disagree with removing the println here, though I do think we may want to replace it with a `log::info` statement instead.

acasajus:
If you're already working on a branch that includes this I'll gladly remove my PR. I'm really eager to try tide in my project in stable when 1.39 lands tomorrow. How long do you reckon it will take to stabilise tide?

yoshuawuyts:
@acasajus I'm glad you're excited! If we don't hit any snags I think we should have a new release within a week? Very excited for this too!

dzrtc:
@yoshuawuyts @acasajus Any updates on this?

PR #350, opened 2019-11-06
Refactor API

yoshuawuyts:
This is a whole set of related changes, addressing most of what was laid out in #336. This disables a lot of work, but sets us up to carefully reintroduce later.

The biggest changes are that we have several core constructs:
- `Server` is the main application struct (formerly called `App`).
- `Endpoint` (trait) is the endpoint people work with.
- `Request` is what is is passed to `Endpoint` and represents an incoming request. (formerly called `Context`).
- `Response` is what is returned from `Endpoint` and represents an outgoing response.

Both `Request` and `Response` operate with `AsyncRead`, and have shorthand methods to send and receive json, forms, querystrings, etc.

I believe this should dramatically streamline the experience of using Tide, and overall make writing HTTP services a lot nicer!

## Examples
__echo server__
```rust
let app = tide::new();
app.at("/echo").get(|req| async move { req });
```
__hello world__
```rust
let app = tide::new();
app.at("/echo").get(|_| async move { "hello world" });
```

## Screenshot
Screenshot_2019-11-04 tide - Rust(2)


CHANGELOG.md
```diff
@@ -34,7 +34,7 @@ were right before splitting up the crate. This release is mostly similar to
 - Added keywords and categories to Cargo.toml.
 - Implemented `Default` for `App`.
 - Added `App::with_state` constructor method.
-- Added `Context::state` (replacing `Context::app_data`)
+- Added `Request::state` (replacing `Request::app_data`)
```
tirr-c: I don't think we need to edit previous changelogs.

LICENSE-APACHE
```diff
@@ -35,7 +35,7 @@
       "Work" shall mean the work of authorship, whether in Source or
       Object form, made available under the License, as indicated by a
       copyright notice that is included in or attached to the work
-      (an example is provided in the Appendix below).
+      (an example is provided in the Serverendix below).
```
tirr-c: This is definitely an artifact of find-and-replace 😄

src/request.rs
```diff
@@ -128,20 +130,34 @@ impl<State> Context<State> {
...
+        let res = serde_qs::from_bytes(&body)
+            .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, format!("could not decode form: {}", e)))?;
+        Ok(res)
     }
+}
 
-    /// Mutably access the extensions to the context.
-    pub fn extensions_mut(&mut self) -> &mut http::Extensions {
-        self.request.extensions_mut()
+impl<State> Read for Request<State> {
```
tirr-c: Well, I think libraries should be runtime-independent and we should impl `futures_io::AsyncRead` instead. I know that `async-std` doesn't have a blanket impl now, but looking at the code, having one seems *technically* possible if we ignore breaking changes.

tirr-c: nvm, it's just a re-export from `futures-io` -- I personally prefer pulling `futures-io` though, but `async-std` with only `std` feature is okay for me too!

yoshuawuyts:
Tests pass; has also been reviewed by a few folks. Going to go ahead and merge!

PR #360, opened 2019-11-27
Redirect

### LINKED ISSUE #348: redirects
I was browsing the new [warp](https://docs.rs/warp) docs, and spotted a fun module called [redirects](https://docs.rs/warp/0.1.20/warp/redirect/index.html). In it there's a small route handler to setup redirects.

I think Tide would benefit from having something similar, as doing redirects are fairly common, and having to write your own can be a bit of a hassle. Making this easy is def something we should want to do.

Importantly, we should be using [`307, temporary redirect`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307) here instead of `302, permanent redirect`. This is to prevent a whole category of bugs that can come with the use of `302`.

## Example
```rust
let app = tide::new();
app.at("/hello").get(tide::redirect("/my_other_url"));
```

yoshuawuyts:
Adds redirects to Tide. Closes #348. Thanks!

## Examples

Redirect from `/nori` to `/chashu`:

```rust
#[async_std::main]
    let mut app = tide::new();
    app.at("/chashu").get(|_| async move { "meow" });
    app.at("/nori").get(tide::redirect("/chashu"));
    Ok(())
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,16 +24,27 @@ easier to write APIs with Tide out of the box.
 
 ## Example
 
+Create a "hello world" app:
 ```rust
-use async_std::task;
+#[async_std::main]
+async fn main() -> Result<(), std::io::Error> {
+    let mut app = tide::new();
+    app.at("/").get(|_| async move { "Hello, world!" });
+    app.listen("127.0.0.1:8080").await?;
+    Ok(())
+}
+```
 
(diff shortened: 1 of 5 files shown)


yoshuawuyts:
Merging as this was a just to move @hoodie's PR onto master from the `redirect` branch; already reviewed it a while ago. CI now passes, which is great!

PR #361, opened 2019-11-27
Json example

yoshuawuyts:
Adds a json example. Thanks!

### DIFF
--- a/examples/json.rs
+++ b/examples/json.rs
@@ -1,22 +1,26 @@
+use async_std::io;
 use async_std::task;
-use serde::{Serialize, Deserialize};
+use serde::{Deserialize, Serialize};
 
 #[derive(Deserialize, Serialize)]
 struct Cat {
-    name: String
+    name: String,
 }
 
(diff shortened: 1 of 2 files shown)


examples/json.rs
```diff
@@ -0,0 +1,29 @@
...
+    name: String,
+}
+
+fn main() -> io::Result<()> {
+    task::block_on(async {
+        let mut app = tide::new();
+
+        app.at("/submit").post(|mut req: tide::Request<()>| {
+            async move {
+                let cat: Cat = req.body_json().await.unwrap();
```
Kerollmops: You changed the tide error handling but is it always possible to `.client_error()?;` with it ?

yoshuawuyts: I'm unsure. Directly using ? definitely isn't. If you could try it out I'd love to hear about it!

Kerollmops: I just tried and it doesn't seems possible, at least with just an `?` at the end of a line :)

PR #364, opened 2019-12-06
impl Endpoint for Service, for service nesting

tirr-c:
~~This is a half-baked version of nesting. Need to strip the path prefix, but I feel this is a good first step!~~ Nevermind, I managed to write necessary mechanics for nesting.

- `HttpService` impl is refactored to use `Endpoint` impl internally.
- Added `Route::strip_prefix`. When used, this will strip the route path prefix from request before passing it to the endpoint.

With this PR, we can finally nest Tide services!

```rust
let mut inner = tide::new();
inner.at("/").get(|_| async { "root" });
inner.at("/foo").get(|_| async { "foo" });
inner.at("/bar").get(|_| async { "bar" });

let mut outer = tide::new();
outer
    .at("/nested")
    .strip_prefix() // catch /nested and /nested/*
    .get(inner.into_http_service()); // the prefix /nested will be stripped here
```

Might need a better name for `strip_prefix`, like `prefixed`?

### DIFF
--- a/src/server/route.rs
+++ b/src/server/route.rs
@@ -12,6 +12,10 @@ use crate::{router::Router, Endpoint};
 pub struct Route<'a, State> {
     router: &'a mut Router<State>,
     path: String,
+    /// Indicates whether the path of current route is treated as a prefix. Set by
+    /// [`strip_prefix`].
+    ///
+    /// [`strip_prefix`]: #method.strip_prefix
     prefix: bool,
 }
 
--- a/src/server/route.rs
+++ b/src/server/route.rs
@@ -12,11 +12,16 @@ use crate::{router::Router, Endpoint};
 pub struct Route<'a, State> {
     router: &'a mut Router<State>,
     path: String,
+    prefix: bool,
 }
 
 impl<'a, State: 'static> Route<'a, State> {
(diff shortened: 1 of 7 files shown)


eignnx:
I wonder if something like `nested_service` would be clearer than `strip_prefix`?

src/server/route.rs
```diff
@@ -12,11 +12,16 @@ use crate::{router::Router, Endpoint};
 pub struct Route<'a, State> {
     router: &'a mut Router<State>,
     path: String,
+    prefix: bool,
```
yoshuawuyts: We could probably use a doc comment here to help share what this is for.

yoshuawuyts:
@tirr-c oh, also another question: on Discord someone asked whether we supported `middleware` on sub-routes, and you mentioned this patch would add support for that. So it looks like this patch specifically doesn't add a `middleware` method on `Route`; but it does seem like that would be possible to add as a follow-up?

src/server/route.rs
tirr-c: Done!

tirr-c:
@yoshuawuyts I think this is good to merge now! Marked `strip_prefix` as unstable, as its name might change in the near future.

And yeah, per-route middleware shouldn't be hard to implement. I'll add it as a follow-up.

PR #369, opened 2019-12-15
fix: context moved

fundon:
If we want to execute multiple times `cx.set_local(value)`.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -79,9 +79,8 @@ impl<State> Request<State> {
     }
 
     /// Set a local value.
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
It should already be possible to do:
```rust
cx.set_local(val).set_local(val);
```

I'm open to making methods not chain at all, but we'd have to port *all* methods over. Unsure whether that's the right API though.

fundon:
Maybe in general they will not be set at the same time.

yoshuawuyts:
This patch hasn't seen activity in a while. Going to go ahead and close it. Thanks!

PR #379, opened 2019-12-22
Add Route::all and Route::nest

tirr-c:
Follow-up of #364.

`Route::all` catches all HTTP methods. This is mandatory for nesting story, but I missed this somehow.

This PR also adds a convenience method `nest_service`, which turns on nesting flag temporarily and calls `all`.

### DIFF
--- a/src/server/route.rs
+++ b/src/server/route.rs
@@ -54,11 +54,26 @@ impl<'a, State: 'static> Route<'a, State> {
 
     /// Treat the current path as a prefix, and strip prefixes from requests.
     ///
+    /// This method is marked unstable as its name might change in the near future.
+    ///
     /// Endpoints will be given a path with the prefix removed.
     #[cfg(any(feature = "unstable", feature = "docs"))]
     #[cfg_attr(feature = "docs", doc(cfg(unstable)))]
     pub fn strip_prefix(&mut self) -> &mut Self {
         self.prefix = true;
+        self
+    }
+
+    /// Nest a [`Server`] at the current path.
+    ///
+    /// [`Server`]: struct.Server.html
+    pub fn nest_service(&mut self, service: crate::Server<State>) -> &mut Self
+    where
+        State: Send + Sync + 'static,
+    {
+        self.prefix = true;
+        self.all(service.into_http_service());
+        self.prefix = false;
         self
     }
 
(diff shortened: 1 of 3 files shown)


src/server/route.rs
```diff
@@ -62,6 +64,19 @@ impl<'a, State: 'static> Route<'a, State> {
         self
     }
 
+    /// Nest a [`Server`] at the current path.
+    ///
+    /// [`Server`]: struct.Server.html
+    pub fn nest_service(&mut self, service: crate::Server<State>) -> &mut Self
```
yoshuawuyts: I wonder if we could shorten this to `nest`. I'm still wondering if we could implement `Service` on `Server` directly, at which point the word `service` will likely disappear from most places. `nest` still feels quite clear to me.

tests/nested.rs
```diff
@@ -53,7 +50,7 @@ fn nested_middleware() {
         app.middleware(test_middleware);
         app.at("/echo").get(echo_path);
         app.at("/:foo/bar").strip_prefix().get(echo_path);
-        route.strip_prefix().get(app.into_http_service());
```
yoshuawuyts: Should we remove the `strip_prefix` method too? Or are there any other uses left for it still?

src/server/route.rs
tirr-c: Well, I think it's okay to leave this as is, because `tide::Server` is more like a builder type for `tide::Service`. This seems natural for me, how do you think?

And yeah, `nest` seems clear enough; maybe we can remove existing method and rename this to `nest`?

tests/nested.rs
tirr-c: I left the method for `HttpService`s other than Tide. Users might want to nest some service and call `route.strip_prefix().all(my_http_service)`.

yoshuawuyts: Alright, fair enough!

src/server/route.rs
yoshuawuyts: Renaming to `nest` sounds good.

Also I've been wondering if we need a distinction between `Server` and `Service`. It seems that if they could be the same struct that would allow us to consolidate the API a bit. But that's something for the future.

tirr-c:
Renamed the method, and merging!

PR #380, opened 2019-12-22
Ported Cookies implementation from archive-split

nhellwig:
This is a port of the cookies implementation from the time when the tide was split into multiple crates.
Changes are:

- ContextExt into RequestExt
- StringError was needed again
- appendHeader also needed

Tests have been ported as well. I was unsure how the strategy will be regarding Error handling so feel free to comment and suggest alternatives. So far I have tried to stay as close to the original version.

Cheers,
Niko


src/middleware/cookies.rs
```diff
@@ -0,0 +1,274 @@
...
+
+        let mut locked_jar = cookie_data
+            .content
+            .write()
+            .map_err(|e| StringError(format!("Failed to get read lock: {}", e)))
+            .with_err_status(StatusCode::INTERNAL_SERVER_ERROR)?;
+        locked_jar.remove(cookie);
+        Ok(())
+    }
+}
```
yoshuawuyts: These should probably be inherent methods on `Request`. We've moved away from providing `Ext` traits on `Request` / `Response` in Tide since we've defined our own.

src/middleware/cookies.rs
```diff
@@ -0,0 +1,274 @@
...
+            assert_eq!(cookie1, "C1=V1");
+            assert_eq!(cookie2, "C2=V2");
+        } else {
+            assert_eq!(cookie2, "C1=V1");
+            assert_eq!(cookie1, "C2=V2");
+        }
+
+        assert!(iter.next().is_none());
+    }
+}
```
yoshuawuyts: Could we move these tests into their own file?

Cargo.toml
```diff
@@ -25,7 +25,8 @@ rustdoc-args = ["--cfg", "feature=\"docs\""]
 [features]
 default = ["hyper-server"]
 hyper-server = ["http-service-hyper"]
-docs = ["unstable"]
+docs = ["unstable", "cookies"]
+cookies = ["cookie"]
```
yoshuawuyts: We shouldn't feature gate `cookies`; the dependency itself is small, and using cookies is common for many applications.

nhellwig:
Thanks for the comments. Have some time probably next week and will work on it.

yoshuawuyts:
Something else I'm wondering about now is whether we could make it so parsing the incoming request cookies into the `CookieJar` could be done on the first call to `cookie`/`set_cookie`/`remove_cookie`. That way APIs that don't use cookies at all wouldn't ever pay the parsing cost, and it would be effectively be zero-cost.

src/request.rs
```diff
@@ -293,6 +300,55 @@ impl<State> Request<State> {
...
+            .with_err_status(StatusCode::INTERNAL_SERVER_ERROR)?;
+
+        let mut locked_jar = cookie_data
+            .content
+            .write()
+            .map_err(|e| StringError(format!("Failed to get read lock: {}", e)))
+            .with_err_status(StatusCode::INTERNAL_SERVER_ERROR)?;
+        locked_jar.add(cookie);
+        Ok(())
+    }
```
yoshuawuyts: Actually, I think it would make more sense if these methods lived on `Response` instead. Before when we still had `Context` there was essentially a single shared mutable state. But that's no longer true, and it's a bit odd we're setting / removing values on the `Request`.

The way this could be implemented is if `Response` would hold its own `CookieJar`, and when the middleware runs we extract all values from the request's `CookieJar` and add them to the `Response`'s `CookieJar` using [`set_original`](https://docs.rs/cookie/0.12.0/cookie/struct.CookieJar.html#method.add_original).

## Example
That way the end-user API would be:

```rust
async fn set_cookie(_req: Request<()>) -> Response {
    let mut res = tide::Response::new(200);
    res.set_cookie(Cookie::new("hello", "world")).unwrap();
    res
}

File: src/request.rs [hunk 5]
```diff
@@ -9,6 +10,12 @@ use async_std::task::{Context, Poll};
 use std::pin::Pin;
 use std::{str::FromStr, sync::Arc};
 
+use crate::error::{Error, ResultExt, StringError};
+use crate::middleware::cookies::CookieData;
+
+const MIDDLEWARE_MISSING_MSG: &str =
+    "CookiesMiddleware must be used to populate request and response cookies";
```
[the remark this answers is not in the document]
yoshuawuyts: Though as mentioned in https://github.com/http-rs/tide/pull/380#issuecomment-576795696 we might not need a feature flag at all if we can make the parsing occur on first access of the cookies: https://github.com/http-rs/tide/pull/380#issuecomment-576795696.

Unsure how well this would work with https://github.com/http-rs/tide/pull/380#discussion_r369169063 though.

yoshuawuyts:
@nhellwig phew, that was a large review. I hadn't thought too much about cookie handling until now, but I feel we're really on track for something good here! The comments so far are part architectural, and part implementation specific. Perhaps best would be to start with the architectural ones:

- [move set/get methods to `Response`](https://github.com/http-rs/tide/pull/380#discussion_r369169063)
- [enable the cookie middleware by default](https://github.com/http-rs/tide/pull/380#discussion_r369120267)
- [parse cookies on first access](https://github.com/http-rs/tide/pull/380#issuecomment-576795696)

When put together I'm unsure how well these ideas go. But I think they're the right direction for these APIs. The other comments are mostly implementation-specific.

Thanks so much for putting the work in! I'm very excited about the direction this is headed!

tests/cookies.rs
```diff
@@ -0,0 +1,118 @@
...
+    if cookie1.to_str().unwrap().starts_with("C1") {
+        assert_eq!(cookie1, "C1=V1");
+        assert_eq!(cookie2, "C2=V2");
+    } else {
+        assert_eq!(cookie2, "C1=V1");
+        assert_eq!(cookie1, "C2=V2");
+    }
+
+    assert!(iter.next().is_none());
+}
```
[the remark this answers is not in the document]
nhellwig: agreed!

src/request.rs
nhellwig: never thought about it, but makes actually much more sense.

yoshuawuyts:
@nhellwig ohh, yeah this is getting pretty close now!

The biggest thing that still seems to be missing is enabling the middleware by default. That would allow us to get rid of the `Result` from all cookie methods as well (replacing it with `unwrap` instead), which would help simplify the API even further!

src/request.rs
nhellwig: working on a solution here. I think could be done but not with `CookieJar`. The problem we need to fill the jar before the user can interact with it. Otherwise if the uses removes a cookie, that was not there before the Jar won't recognize is. This means the Set-Cookie won't be sent with empty value and thus not removed.

A workaround could be to track actions in the response rather than manipulation a cookie jar and later replay in cookie middleware. Should actually work fine for "normal" user behaviour. In the end, I guess the overhead should be minimal considering a user will not set and remove cookies with the same name in the `Response`.

nhellwig: @yoshuawuyts what do you think?

nhellwig:
Yeah that is worth looking at ^^. 

Just moved the `set_cookie` and `remove_cookie` to `Response`.
Next I will add doc tests as well.

src/request.rs
```diff
@@ -9,6 +10,11 @@ use async_std::task::{Context, Poll};
 use std::pin::Pin;
 use std::{str::FromStr, sync::Arc};
 
+use crate::error::{Error, ResultExt, StringError};
+use crate::middleware::cookies::CookieData;
+
+const MIDDLEWARE_MISSING_MSG: &str = "`CookiesMiddleware` has not been enabled";
```
yoshuawuyts: We should enable the cookies middleware by default; removing the need for this string and the corresponding logic.

PR #382, opened 2019-12-22
IntoResponse implementation for Result<A: IntoResponse, B: IntoResponse>

### LINKED ISSUE #371: Implementing IntoResponse for tide::Result<tide::Response>
Hey!

I have started to port https://github.com/colinbankier/realworld-tide from `v0.3.0` to `v0.4.0` - so far so good, no major issues.

I noticed that `cargo build` fails if I try to mount a route with following signature:
```rust
pub async fn list_articles(cx: Request<Repo>) -> tide::Result<Response> {
    [...]
}
```
with the following error
```
error[E0277]: the trait bound `std::result::Result<tide::response::Response, tide::error::Error>: tide::response::into_response::IntoResponse` is not satisfied
  --> src/main.rs:28:29
   |
28 |         api.at("/articles").get(web::articles::list_articles);
   |                             ^^^ the trait `tide::response::into_response::IntoResponse` is not implemented for `std::result::Result<tide::response::Response, tide::error::Error>`
   |
   = note: required because of the requirements on the impl of `tide::endpoint::Endpoint<db::Repo<diesel::PgConnection>>` for `fn(tide::request::Request<db::Repo<diesel::PgConnection>>) -> impl core::future::future::Future {web::articles::list_articles}`
```

I'd be happy to submit a PR to implement `IntoResponse` for `tide::Result<tide::Response>`, but I'd like to understand first if this implementation is missing on purpose due to some pending discussions around error handling in `tide` :grin:

LukeMathWalker:
This closes #371 - it turns out there was already an implementation, it was just commented out :grin:

### DIFF
--- a/src/response/into_response.rs
+++ b/src/response/into_response.rs
@@ -73,16 +73,7 @@ impl<T: IntoResponse, U: IntoResponse> IntoResponse for Result<T, U> {
     fn into_response(self) -> Response {
         match self {
             Ok(r) => r.into_response(),
-            Err(r) => {
-                let res = r.into_response();
-                if res.status().is_success() {
-                    panic!(
-                        "Attempted to yield error response with success code {:?}",
-                        res.status()
-                    )
-                }
-                res
-            }
+            Err(r) => r.into_response(),
         }
     }
 }
--- a/src/response/into_response.rs
+++ b/src/response/into_response.rs
@@ -69,23 +69,23 @@ impl IntoResponse for &'_ str {
 //     }
 // }
 
-// impl<T: IntoResponse, U: IntoResponse> IntoResponse for Result<T, U> {
-//     fn into_response(self) -> Response {
-//         match self {
-//             Ok(r) => r.into_response(),
-//             Err(r) => {
-//                 let res = r.into_response();
-//                 if res.status().is_success() {
-//                     panic!(
-//                         "Attempted to yield error response with success code {:?}",
-//                         res.status()
-//                     )
-//                 }
-//                 res
-//             }
-//         }
-//     }
-// }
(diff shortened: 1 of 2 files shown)


src/response/into_response.rs
```diff
@@ -69,23 +69,23 @@ impl IntoResponse for &'_ str {
...
-//         }
-//     }
-// }
+impl<T: IntoResponse, U: IntoResponse> IntoResponse for Result<T, U> {
+    fn into_response(self) -> Response {
+        match self {
+            Ok(r) => r.into_response(),
+            Err(r) => {
+                let res = r.into_response();
+                if res.status().is_success() {
```
dignifiedquire: this seems like an unnecessary limitation, maybe the user wants to return a different kind of success response on certain errors, so I wouldn’t force this, especially not with a panic

LukeMathWalker: I am conflicted about this - I think it's semantically confusing to return something that will map to a successful status code in the `Err` variant of a `Result`.
Should it be a `panic`? I guess that's up for discussion.

dignifiedquire:
what if I want to return a redirect?

dvic:
Redirects are one thing but I think it can also be perfectly fine to return HTTP 200's for errors. For example, often in GraphQL HTTP implementations one wants to return HTTP 200 for error responses as there is a distinction between success and failure responses in the serialization format.

LukeMathWalker:
If there is consensus on the possibility of returning successes from the `Err` variant, I am happy to change the PR and simply cast both variants to `Response` without further checks.

LukeMathWalker:
Updated the PR to avoid `panic` on the `Err` variant if it returns a success.

PR #385, opened 2020-01-07
Add support for nested application with different application state type

alex-berger:
This PR enables nested applications to have a different application state type than the enclosing application has. This small change allows to compose applications with different state types.

### DIFF
--- a/tests/nested.rs
+++ b/tests/nested.rs
@@ -93,9 +93,11 @@ fn nested_middleware() {
 fn nested_with_different_state() {
     let mut outer = tide::new();
     let mut inner = tide::with_state(42);
-    inner.at("/").get(|req: tide::Request<i32>| async move {
-        let num = req.state();
-        format!("the number is {}", num)
+    inner.at("/").get(|req: tide::Request<i32>| {
+        async move {
+            let num = req.state();
+            format!("the number is {}", num)
+        }
     });
     outer.at("/").get(|_| async move { "Hello, world!" });
     outer.at("/foo").nest(inner);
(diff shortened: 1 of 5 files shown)


felipesere:
👋  Can you possibly give a bigger example how this would be used?

I tried the following but it didn't work:
```
use async_std::task;

    task::block_on(async {
        let mut app = tide::new();

        let mut inner = tide::with_state(42);
        inner.at("/").get(|req: tide::Request<i32>| async move {
            let num = req.state();

            format!("the number is {}", num)
        });

        app.at("/").get(|_| async move { "Hello, world!" });
        app.at("/foo").get(inner.into_http_service());
        Ok(())
    })
}
```
Doing `http GET http://127.0.0.1/foo` just gives me a 404.

I am not really sure how this change would pass through and convert the request from `Request<S>` to `Request<I>` without some kind of mapping function?

I am curious ot see where this goes! 💙

alex-berger:
@felipesere, you have to use `.strip_prefix()` then it will work as expected:

```rust
app.at("/foo").strip_prefix().get(inner.into_http_service());
```

tests/nested.rs
```diff
@@ -91,3 +91,34 @@ fn nested_middleware() {
...
+    let mut inner = tide::with_state(42);
+    inner.at("/").get(|req: tide::Request<i32>| async move {
+        let num = req.state();
+        format!("the number is {}", num)
+    });
+    outer.at("/").get(|_| async move { "Hello, world!" });
+    outer
+        .at("/foo")
+        .strip_prefix()
+        .get(inner.into_http_service());
```
tirr-c: I've added a simpler method to nest services. This can be simplified to `.at().nest(inner)` now; it would be good to rebase and update!

PR #388, opened 2020-01-19
Add `Method Not Allowed` in the Router

isgj:
When the same path is registered but with another method respond with `405` instead of `404`

### DIFF
--- a/src/router.rs
+++ b/src/router.rs
@@ -61,9 +61,16 @@ impl<State: 'static> Router<State> {
             // if not then fallback to the behavior of HTTP GET else proceed as usual
 
             self.route(path, http::Method::GET)
-        } else if self.method_map.values().any(|r| r.recognize(path).is_ok()) {
+        } else if self
+            .method_map
+            .iter()
+            .filter(|(k, _)| *k != method)
+            .any(|(_, r)| r.recognize(path).is_ok())
+        {
+            // If this `path` can be handled by a callback registered with a different HTTP method
+            // should return 405 Method Not Allowed
             Selection {
-                endpoint: &method_not_allawed,
+                endpoint: &method_not_allowed,
                 params: Params::new(),
             }
         } else {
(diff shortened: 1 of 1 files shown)


src/router.rs
```diff
@@ -61,6 +61,11 @@ impl<State: 'static> Router<State> {
             // if not then fallback to the behavior of HTTP GET else proceed as usual
 
             self.route(path, http::Method::GET)
+        } else if self.method_map.values().any(|r| r.recognize(path).is_ok()) {
+            Selection {
+                endpoint: &method_not_allawed,
```
yoshuawuyts:
```suggestion
                endpoint: &method_not_allowed,
```

src/router.rs
```diff
@@ -73,3 +78,7 @@ impl<State: 'static> Router<State> {
 fn not_found_endpoint<State>(_cx: Request<State>) -> BoxFuture<'static, Response> {
     Box::pin(async move { Response::new(http::StatusCode::NOT_FOUND.as_u16()) })
 }
+
+fn method_not_allawed<State>(_cx: Request<State>) -> BoxFuture<'static, Response> {
```
yoshuawuyts:
```suggestion
fn method_not_allowed<State>(_cx: Request<State>) -> BoxFuture<'static, Response> {
```

src/router.rs
```diff
@@ -61,6 +61,11 @@ impl<State: 'static> Router<State> {
             // if not then fallback to the behavior of HTTP GET else proceed as usual
 
             self.route(path, http::Method::GET)
+        } else if self.method_map.values().any(|r| r.recognize(path).is_ok()) {
```
yoshuawuyts: This code contains quite a few branches which can be hard to follow; could you maybe add a comment here explaining what this does?

yoshuawuyts (approved):
This patch is really good; all for better diagnostics!

Left a few nits, should be good to merge when addressed. Thanks so much for this!

isgj:
Happy to hear that.

PR #389, opened 2020-01-20
A re-work of the error handling system

kbillings:
This is my idea of error system that should work with the ideas mentioned by @yoshuawuyts
in https://github.com/http-rs/tide/issues/371#issuecomment-574554804 .

This system should both handle errors that have a custom `IntoResponse` implementation
as well as more generic errors using methods on `ResultExt` to create an `Error` type.

Some changes:

* Rework methods on `ResultExt`
* Add `From<E>` for `Error` where `E: Display`
* Add back `impl<T, E> IntoResponse for Result<T, E>`
* Changed `Result` alias from `Result<T>` to `Result<T, E = Error>`

How it should look in endpoints:

```rust
struct CustomError;
impl IntoResponse for CustomError {
    fn into_response(self) -> Response { Response::new(418) }
}

async fn endpoint_a(_req: tide::Request<()>) -> Result<String, CustomError> {
    Err(CustomError)
}

async fn endpoint_b(_req: tide::Request<()>) -> tide::Result<String> {
    use tide::ResultExt;

    example_fn().await?;                         // returns 500 with body of "error"
    example_fn().await.with_empty(400)?;         // returns 400 with empty body
    example_fn().await.with_status(400)?;        // returns 400 with body of "error"
    example_fn().await.with_err_res(|_| "hello") // returns 200 with body of "hello"

    Ok(String::from("ok"))
}

// do stuff that can return errors
async fn example_fn() -> Result<(), String> {
    Err(String::from("error"))
}
```

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,16 +1,29 @@
 //! Tide error types.
 use http::{HttpTryFrom, StatusCode};
-use http_service::Body;
 
 use crate::response::{IntoResponse, Response};
 
 /// A specialized Result type for Tide.
-pub type Result<T = Response> = std::result::Result<T, Error>;
+pub type Result<T, E = Error> = std::result::Result<T, E>;
 
-/// A generic error.
+/// An error which holds a response.
 #[derive(Debug)]
 pub struct Error {
     resp: Response,
+}
+
+impl Error {
+    /// Create an `Error` with the given response.
+    pub fn from_response(resp: Response) -> Error {
+        Error { resp }
+    }
+
+    /// Create an `Error` with an empty response and the given status code.
+    pub fn from_status(status: StatusCode) -> Error {
+        Error {
+            resp: Response::new(status.as_u16()),
+        }
+    }
 }
 
 impl IntoResponse for Error {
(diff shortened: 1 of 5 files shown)


PeteJodo:
I hope this is a useful question/thought but what could be a way a user determines the environment their app is running and if it's development, return useful debugging information?

Would it require the user having to do something explicit for each endpoint function that can potentially return an error, for example referring to app state every time i.e. `CustomError::from_state(reg.state())`

isgj:
It will be handy to have a method to map also a **serializable**  error, not only the ones that implement the `Display` trait.

```rust
new_data.validate().map_to_json(400)?;
// the next line doesn't work either
// .map_err(|e| Response::new(400).body_json(&e))?;
```

where `validate()` return `Result<NewData, E:Serialize>`

yoshuawuyts:
@kbillings thanks so much for this PR! Wrote about this more last week: https://blog.yoshuawuyts.com/async-http/#error-handling. We now have a full error implementation as as part of `http-types`, which we'll likely merge in the near future.

Some of the ideas in this patch have been integrated into `http-types` as well, especially the `status` method that allows setting a status as a shorthand.

This is one issue of about 8? that touch on error handling in Tide, and wanted to share an update on progress that has been made since.

PR #397, opened 2020-01-31
Make Endpoint::call generic over lifetime

tirr-c:
This is a breaking change.

Make `Endpoint::call` generic over lifetime like `Middleware::handle`.

### DIFF
--- a/src/redirect.rs
+++ b/src/redirect.rs
@@ -1,8 +1,4 @@
-use async_std::future;
-use async_std::task::{Context, Poll};
-
-use std::pin::Pin;
-
+use crate::utils::BoxFuture;
 use crate::{Endpoint, Request, Response};
 
 /// Redirect a route to another route.
@@ -21,7 +17,7 @@ use crate::{Endpoint, Request, Response};
 /// app.listen("127.0.0.1:8080").await?;
 /// #
 /// # Ok(()) }) }
-/// ````
+/// ```
 pub fn redirect<State>(location: impl AsRef<str>) -> impl Endpoint<State> {
     let location = location.as_ref().to_owned();
     Redirect { location }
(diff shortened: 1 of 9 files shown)


src/router.rs
```diff
@@ -33,12 +33,11 @@ impl<State: 'static> Router<State> {
         self.method_map
             .entry(method)
             .or_insert_with(MethodRouter::new)
-            .add(path, Box::new(move |cx| Box::pin(ep.call(cx))))
+            .add(path, Box::new(ep))
     }
 
     pub(crate) fn add_all(&mut self, path: &str, ep: impl Endpoint<State>) {
-        self.all_method_router
-            .add(path, Box::new(move |cx| Box::pin(ep.call(cx))))
```
yoshuawuyts: Oh, getting rid of this is nice!

src/server/mod.rs
```diff
@@ -321,14 +321,23 @@ impl<State: Send + Sync + 'static> Server<State> {
...
+impl<State> Clone for Service<State> {
+    fn clone(&self) -> Self {
+        Self {
+            router: self.router.clone(),
+            state: self.state.clone(),
+            middleware: self.middleware.clone(),
+        }
+    }
+}
+
```
yoshuawuyts: What's the difference with the `Derive` block? It looks like there should be no difference?

src/redirect.rs
```diff
@@ -21,7 +17,7 @@ use crate::{Endpoint, Request, Response};
 /// app.listen("127.0.0.1:8080").await?;
 /// #
 /// # Ok(()) }) }
-/// ````
```
yoshuawuyts: ha, nice catch

src/server/mod.rs
tirr-c: `derive` block adds `State: Clone` bound, which is undesirable.

yoshuawuyts: Oh I thought that bound was intentional. How else do we guarantee that the State can always be cloned for each route?

I feel like I'm missing something here 😅

PR #398, opened 2020-02-01
Create tera.rs

eribol:
An example for tide using tera template engine .


jbr:
Closing this, as there is now a `tide-tera` crate with an example: https://github.com/jbr/tide-tera

PR #399, opened 2020-02-01
Per-route middleware

### LINKED ISSUE #320: Support "scoped" middlewares
## Feature Request
In practices, we often want enable some middlewares on the subtree of the routes.

## Detailed Description

For example, we have URI `/v1/:user_id/posts/:post_id`, and we need to check whether the post belongs to that user. Or we have an `OAuth2Middleware`, but the only subset of our APIs need to be protected.

Currently, we need to intrusive modify the middleware if we want to archive this purpose. We have the maintain the router but also the middleware implementation.

## Context

```rust
root.at("/:user_id/posts").nest(|parent| {
    parent.middleware(access_guard); // <----
    parent.post(crate::service::post::new_post);
    parent.at("/:post_id")
        .get(crate::service::post::get_post)
        .put(crate::service::post::mod_post)
        .delete(crate::service::post::del_post);
});
```

## Possible Implementation

1. Add associated middlewares information to the NFA states of `route-recognizer`.(Global middlewares associated with the initial state??)
2. When perform the matching, we need to trace the path from the initial state to the terminal state. After the matching process, we can collect the associated middlewares on the path and invoke them in order.

tirr-c:
This PR adds per-route middleware support. Current design is as follows:

* `Route::middleware` applies given middleware to the route.
* Following `Route::method` (and `Route::all`) will apply middleware given before.
* Middleware list for the route can be reset using `Route::reset_middleware`.
* When `Route::at` is called, per-route middleware applied to the parent will be applied to the child route by default. This can be reset using `.reset_middleware()`.

Feel free to comment if there is better design!

---

Closes #320.

### DIFF
--- a/src/endpoint.rs
+++ b/src/endpoint.rs
@@ -1,7 +1,10 @@
+use std::sync::Arc;
+
 use async_std::future::Future;
 
+use crate::middleware::Next;
 use crate::utils::BoxFuture;
-use crate::{response::IntoResponse, Request, Response};
+use crate::{response::IntoResponse, Middleware, Request, Response};
 
 /// An HTTP request handler.
 ///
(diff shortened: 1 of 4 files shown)


yoshuawuyts:
> Subroutes returned by Route::at won't have any middleware applied.

I would expect the following to work:

```rust
let mut app = tide::new();
app.middleware(middleware_a());

app.at("/").get(|req| "hi"); // middleware "a" applies here

app.at("/foo")
    .middleware(middleware_b())
    .get(|req| "hello");  // middleware "a" + b" apply here

```

This is also so that top-level middleware such as logging, cookie handling and others will always be able to run. Being able to define per-route, and per-app middleware is generally desirable. Does it work like this?

yoshuawuyts:
Also the following:
```rust
let mut outer = tide::new();
outer.middleware(middleware_b());
outer.at("/foo").get(|req| "hello");  // middleware "b" applies here

let mut inner = tide::new();
inner.middleware(middleware_a());
inner.at("/").get(|req| "hi"); // middleware "b" + "a" apply here

outer.nest(inner);
outer.listen("localhost:8080").await?;
```

tirr-c:
@yoshuawuyts Ah, I meant per-route middleware is reset -- yeah, per-app middleware will remain applied. I'll add some tests to cover those!

yoshuawuyts:
@tirr Can you share an example of what you mean? I don't have a clear understanding of the rules quite yet.

tirr-c:
@yoshuawuyts I meant like this:

```rust
let mut app = tide::new();
app.middleware(MiddlewareGlobal); // This is app-global and won't be reset
app.at("/a")
    .middleware(MiddlewareA) // This is per-route and subject to be reset
    .get(echo_path) // Global + A
    .at("/b") // /a/b, A is removed here
    .middleware(MiddlewareB)
    .get(echo_path); // Global + B
```

(and you mentioned the wrong person 😅)

yoshuawuyts:
Lol oops, yeah typing from the mobile app.

Hmm, yeah I think I would expect that to work still, even if I probably wouldn't write it like that myself. Route "b" is nested under "a", so hierarchically I'd expect all of the middleware from "a" to apply to "b".

tirr-c:
@yoshuawuyts That's the alternative approach I wrote. If you feel it's better, I'll change the behavior!

yoshuawuyts:
Yess, thank you — that'd be great!

tirr-c:
@yoshuawuyts Ooh I misunderstood your first code! It's quite tough to apply middleware of the parent if the subroute is created *without* using `Route::at`, e.g. calling `app.at("/parent")` and `app.at("/parent/child")` separately. I'm going to make subroutes middleware work only when it's called like `app.at("/parent").at("/child")`. Is it okay to do so?

yoshuawuyts:
I think that's ok! As long as the global middleware still works. I think using `at` as a means of nesting works well!

tirr-c:
@yoshuawuyts Done! Tests are added to ensure what works and what doesn't.

Issue #396, opened 2020-01-31
Cookies are set multiply times when using *server.nest*


isgj:
When setting a cookie in a nested service it is set multiply times depending on the level of the nested service. For example:

```rust
use async_std::task;
use cookie::Cookie;
use tide::{self, Response};

fn main() {
    task::block_on(async {
        let mut inner = tide::new();
        inner.at("/3").get(|_| async {
            let mut r = Response::new(200);
            r.set_cookie(Cookie::new("snack", "tuna"));
            r.body_string(String::from("ok"))
        });

        let mut second = tide::new();
        second.at("/2").nest(inner);

        let mut outer = tide::new();
        outer.at("/1").nest(second);

        outer
            .listen("localhost:8080")
            .await
            .expect("Error starting the server");
    })
}
```
When `GET localhost:8080/1/2/3` the response is 
```
HTTP/1.1 200 OK
content-type: text/plain; charset=utf-8
set-cookie: snack=tuna,snack=tuna,snack=tuna
transfer-encoding: chunked
date: Fri, 31 Jan 2020 10:39:54 GMT

ok
```

The cookie is set 3 times.

yoshuawuyts:
@isgj thanks for the bug report! I suspect this is because each instance of `Server` has its own instance of `CookieMiddleware` enabled, it looks at the `CookieJar` struct three times in a row, and sets the headers accordingly.

I see two ways of fixing this:
1. Make the `CookieJar` value in `Response` and `Request` an `Option`, so that running it multiple times becomes a no-op.
2. Make it so nesting applications disables default middleware for the nested applications.

I'm thinking the better solution here would be the second one. Because it would allow us to prevent future issues when we enable other middleware by default (e.g. `logging` comes to mind).

## Implementation

I think the way to fix this is to remove this:

https://github.com/http-rs/tide/blob/4e2857a2bc46726ffb6a00c5a83c33420ca67e7c/src/server/mod.rs#L206-L208

And instead add it back inside `Server::listen`, so that the default middleware is only enabled once:

https://github.com/http-rs/tide/blob/4e2857a2bc46726ffb6a00c5a83c33420ca67e7c/src/server/mod.rs#L308-L310

The default middleware should probably be pushed to the start of the middleware vector though, so that logging, cookies, etc. all run in a predetermined order before any userland middleware.

cc/ @tirr-c

PR #400, opened 2020-02-01
Fix nested cookies

yoshuawuyts:
Fixes https://github.com/http-rs/tide/issues/396 by ensuring the cookies middleware is only created when we actually call `.listen` and not when nested.

This also adds a `into_mock` method so that the cookies middleware can still be constructed when running tests. This seemed like a good setup to start improving our test workflow, by at least starting to remove some of the boilerplate involved now.

Thanks!


isgj:
Just throwing it out there. Should you provide a function to listen without any default middleware?

src/server/mod.rs
```diff
@@ -279,6 +277,11 @@ impl<State: Send + Sync + 'static> Server<State> {
     ///
     /// This lower-level method lets you host a Tide application within an HTTP
     /// server of your choice, via the `http_service` interface crate.
+    ///
+    /// # Middleware
+    ///
+    /// This method does not enable default middleware. See `into_mock` for
+    /// converting Tide into a server that can be used for testing purposes.
     pub fn into_http_service(self) -> Service<State> {
```
tirr-c: How about leaving this as is, and provide a new method that doesn't apply any default middleware? One rarely calls `.into_http_service()` directly for nesting since `Route::nest` handles it instead, so I think it should be okay?

src/server/mod.rs
```diff
@@ -315,6 +329,20 @@ impl<State: Send + Sync + 'static> Server<State> {
...
+    ///
+    /// # Stability
+    ///
+    /// This method will be expanded in the future to allow mocking requests
+    /// in a more direct manner.
+    pub fn into_mock(mut self) -> io::Result<http_service_mock::TestBackend<Service<State>>> {
+        self.enable_default_middleware();
+        let server = http_service_mock::make_server(self.into_http_service())?;
+        Ok(server)
+    }
```
tirr-c: By providing separate method for nesting, we don't need to provide a "special" wrapper method like this.

yoshuawuyts:
This is too far out of sync with master; we should try again in a new patch.

PR #403, opened 2020-02-03
Don't wipe cookie events when using cors middleware

callym:
The cors middleware was going from a `tide::Response` to a `http_service::Response`, which was wiping the newly added cookie events vec. This PR changes it - it also changes some sort of unrelated header code (the old code was getting the header map from the response, which this crates response type doesn't expose, and I wasn't sure whether doing it this way was better than adding a `pub(crate)` method or something just to add some headers?)

### DIFF
--- a/src/middleware/cors.rs
+++ b/src/middleware/cors.rs
@@ -168,22 +168,31 @@ impl<State: Send + Sync + 'static> Middleware<State> for Cors {
                 return self.build_preflight_response(&origin).into();
             }
 
-            let mut response: http_service::Response = next.run(req).await.into();
-            let headers = response.headers_mut();
+            let access_control_allow_origin = header::ACCESS_CONTROL_ALLOW_ORIGIN.as_ref();
+            let access_control_allow_credentials =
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
@callym Thanks for opening this issue! I've got two requests:

1. Could you perhaps add a test that catches the regression? I think I understand what you're doing here, but seeing a test would help ensure we don't regress.
2. Could you revert the cookie header name changes? I understand why this is desirable, but it seems like a separate issue, and we also have a different solution for that in the works.

Thanks!

callym:
@yoshuawuyts thanks for looking at this! I've reverted the header change stuff, agree that it's a separate issue :smile: 

re tests - where should I add tests for this? I think we'd want something that somehow goes through every middleware and makes sure that the request object we put in is the same one we get out at the end, but not sure whether that's right or how that'd be done?

(Because this issue is caused by Tide's Request type including extra data, so going from tide::Request -> http::Request -> tide::Request is lossy - because of cookie events and maybe in the future even more data)

callym:
I'm going to close this because #599 looks like it's a better fix :smile:

PR #404, opened 2020-02-04
Update async-std and bytes dependencies

jeff-hiner:
We can tackle hyper 0.13 next, but let's get the low hanging fruit first.


yoshuawuyts:
Hey @jeff-hiner; thanks for opening this PR. Apologies for not getting around to it sooner; we're trying to become better at addressing issues on time. It seems there merge conflicts have been introduced in the interim. I'm going to go ahead and close this patch since we can't land it as-is, but if you'd like to re-submit a patch to bump our deps, that'd be much appreciated still!

PR #405, opened 2020-02-07
Replace utils::BoxFuture with futures::future::BoxFuture

ar3s3ru:
Currently, `utils` module only exists for `BoxFuture` type alias declaration.

However, since:

* this type alias exists already in `futures` crate
* `futures` crate is already a dependency of `tide`

all references to `utils::BoxFuture` can be replaced with `futures::future::BoxFuture` and the `utils` module deleted.

### DIFF
--- a/src/endpoint.rs
+++ b/src/endpoint.rs
@@ -1,6 +1,7 @@
 use async_std::future::Future;
 
-use crate::utils::BoxFuture;
+use futures::future::BoxFuture;
+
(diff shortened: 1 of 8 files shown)


yoshuawuyts:
Thanks for opening this issue! We want to only depend on futures-core in the future; as such having it be part of utils was intentional. This patch moves in the opposite direction we intend to go.

ar3s3ru:
Oh, I see.  
Sorry for the useless PR then :sweat_smile:

PR #409, opened 2020-02-20
Make set_header and append_header more flexible

isgj:
Almost same change as http-rs/surf#123.


isgj:
Since `tide` will make use of `http_types` these changes won't be necessary any more.
Feel free to close this PR. Thank you.

PR #411, opened 2020-02-23
Uncomment impl IntoResponse for Result<T: IntoResponse, U: IntoResponse>

strohel:
Was commented-out in 3aea1d9baba1a13b by @yoshuawuyts.

Not sure shy it was commented out. Maybe a different approach was to be pursued?

My motivation to uncomment was that I want my endpoint handler to have Result<T, E> return type, so that I can use `?` operator directly in it. I was not able to implement this in my crate due to Rust orphan rules.

If this is a desired change, I can add tests. If not, I'd be glad for any advice how to achieve the goal differently.

### DIFF
--- a/src/response/into_response.rs
+++ b/src/response/into_response.rs
@@ -69,23 +69,23 @@ impl IntoResponse for &'_ str {
 //     }
 // }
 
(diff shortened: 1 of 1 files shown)


strohel:
Ah, silly me, this is a duplicate of #382. Closing, should have checked open issues/prs first.

PR #414, opened 2020-03-03
feat: add implementation based on async-h1 and http-service-h1

dignifiedquire:


src/response/mod.rs
```diff
@@ -41,49 +41,60 @@ impl Response {
...
+        self.res.set_status(status);
         self
     }
 
     /// Insert an HTTP header.
-    pub fn set_header(mut self, key: &'static str, value: impl AsRef<str>) -> Self {
+    pub fn set_header(
+        mut self,
+        key: http_types::headers::HeaderName,
+        value: impl AsRef<str>,
```
yoshuawuyts: Ideally this could forward https://docs.rs/http-types/1.0.1/http_types/struct.Response.html?search=#method.insert_header so that we can get rid of all the `parse().unwrap()` calls.

```suggestion
        key: http_types::headers::HeaderName,
        values: impl http_types::ToHeaderValues
```

dignifiedquire: almost makes things better:

```
the trait `http_types::headers::to_header_values::ToHeaderValues` is not implemented for `&std::string::String`
```

Cargo.toml
```diff
@@ -23,39 +23,46 @@ features = ["docs"]
...
 unstable = []
 
 [dependencies]
-futures = "0.3.1"
-http = "0.1.19"
-http-service = "0.4.0"
-http-service-hyper = { version = "0.4.1", optional = true }
+http-types = "1.0.1"
+http-service = { git = "https://github.com/dignifiedquire/http-service-1", branch = "h1" }
+http-service-h1 = { git = "https://github.com/dignifiedquire/http-service-1", branch = "h1", optional = true }
```
pickfire: Why both of these is the same git repository?

dignifiedquire: because they are both in the same repository?

PR #415, opened 2020-03-05
Static files

yoshuawuyts:
Init static file serving support. This depends on https://github.com/http-rs/tide/pull/414.

```rust
#[async_std::main]
    let mut app = tide::new();
    app.at("/").serve_dir("/public");
    Ok(())
}
```

### DIFF
--- a/src/server/route.rs
+++ b/src/server/route.rs
@@ -1,5 +1,9 @@
+use std::fmt::Debug;
+use std::io;
+use std::path::Path;
 use std::sync::Arc;
 
+use super::serve_dir::ServeDir;
 use crate::endpoint::MiddlewareEndpoint;
 use crate::utils::BoxFuture;
 use crate::{router::Router, Endpoint, Middleware, Response};
@@ -54,6 +58,11 @@ impl<'a, State: 'static> Route<'a, State> {
         }
     }
 
+    /// Get the current path.
+    pub fn path(&self) -> &str {
+        &self.path
+    }
+
     /// Treat the current path as a prefix, and strip prefixes from requests.
     ///
     /// This method is marked unstable as its name might change in the near future.
@@ -67,7 +76,15 @@ impl<'a, State: 'static> Route<'a, State> {
     }
 
     /// Apply the given middleware to the current route.
-    pub fn middleware(&mut self, middleware: impl Middleware<State>) -> &mut Self {
+    pub fn middleware<M>(&mut self, middleware: M) -> &mut Self
+    where
+        M: Middleware<State> + Debug,
+    {
+        log::trace!(
+            "Adding middleware {:?} to route {:?}",
+            middleware,
+            self.path
+        );
         self.middleware.push(Arc::new(middleware));
         self
     }
@@ -90,6 +107,33 @@ impl<'a, State: 'static> Route<'a, State> {
         self.all(service);
         self.prefix = false;
         self
+    }
+
+    /// Serve a directory statically.
+    ///
+    /// Each file will be streamed from disk, and a mime type will be determined
+    /// based on magic bytes.
+    ///
+    /// # Examples
+    ///
+    /// Serve the contents of the local directory `./public/images/*` from
+    /// `localhost:8080/images/*`.
+    ///
+    /// ```no_run
+    /// #[async_std::main]
+    /// async fn main() -> Result<(), std::io::Error> {
+    ///     let mut app = tide::new();
+    ///     app.at("/public/images").serve_dir("images/")?;
+    ///     app.listen("127.0.0.1:8080").await?;
+    ///     Ok(())
+    /// }
+    /// ```
+    pub fn serve_dir(&mut self, dir: impl AsRef<Path>) -> io::Result<()> {
+        // Verify path exists, return error if it doesn't.
+        let dir = dir.as_ref().to_owned().canonicalize()?;
+        let prefix = self.path().to_string();
+        self.at("*").get(ServeDir::new(prefix, dir));
+        Ok(())
     }
 
     /// Add an endpoint for the given HTTP method
--- a/src/server/serve_dir.rs
+++ b/src/server/serve_dir.rs
@@ -1,0 +1,79 @@
+use async_std::fs::File;
+use async_std::io::BufReader;
+use http_types::{Body, StatusCode};
+
+use crate::{Endpoint, Request, Response, Result};
+
+use std::path::{Path, PathBuf};
+
+type BoxFuture<'a, T> = std::pin::Pin<Box<dyn std::future::Future<Output = T> + 'a + Send>>;
+pub struct ServeDir {
+    prefix: String,
+    dir: PathBuf,
+}
+
+impl ServeDir {
+    /// Create a new instance of `ServeDir`.
+    pub(crate) fn new(prefix: String, dir: PathBuf) -> Self {
+        Self { prefix, dir }
+    }
+}
+
+impl<State> Endpoint<State> for ServeDir {
+    fn call<'a>(&'a self, req: Request<State>) -> BoxFuture<'a, Result<Response>> {
+        let path = req.uri().path();
+        let path = path.replacen(&self.prefix, "", 1);
+        let path = path.trim_start_matches('/');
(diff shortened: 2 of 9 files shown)


src/server/serve_dir.rs
```diff
@@ -34,9 +34,23 @@ impl<State> Endpoint<State> for ServeDir {
...
-                    // TODO: verify this is a sub-path of the original dir.
-                    File::open(file).await.unwrap() // TODO: remove unwrap
+                Ok(mut file_path) => {
+                    // Verify this is a sub-path of the original dir.
+                    let mut file_iter = (&mut file_path).iter();
+                    for lhs in &dir {
+                        if Some(lhs) != file_iter.next() {
+                            return Response::new(StatusCode::Forbidden);
+                        }
+                    }
```
pickfire: Is this the same as?
```rust
                    if !(&mut file_path).iter().eq(&dir) {
                        return Response::new(StatusCode::Forbidden);
                    }
```

yoshuawuyts: @pickfire not quite: this will never resolve to true since the iterators are of different length. I made sure to try `eq` before doing it this way. If `file_path` implemented `ExactSizeIterator` we could use `len` with `take` to make sure the lengths are the same. But alas.

pickfire: @yoshuawuyts How about `all`?

yoshuawuyts: Oh yeah that might work. We should try it out!

pickfire:
@yoshuawuyts Is there any reason why the original `body` builder method is changed to `set_body` method separately? I personally don't think it looks nice but probably there is a reason behind it.

yoshuawuyts:
@pickfire this is more of a temporary name than anything else. Ideally we could get closer to http-types' way of constructing requests and responses but this is a stopgap in the meantime.

pickfire:
@yoshuawuyts It seemed similar to the one in http-types. Is there any reason why it uses `set_body` rather than `body_mut` or `body`? Currently, http-types have quite a few `*_body` and it does seemed a bit weird for me.

src/server/serve_dir.rs
pickfire: We could use this instead. It compiles.
```rust
                    if !dir.iter().all(|lhs| Some(lhs) == file_iter.next()) {
                        return Response::new(StatusCode::Forbidden);
                    }
```

src/server/route.rs
```diff
@@ -54,6 +58,24 @@ impl<'a, State: 'static> Route<'a, State> {
...
+    //         let entry = entry?;
+    //         if !entry.file_type().is_dir() {
+    //             let p = entry.into_path();
+    //             continue;
+    //         }
+    //         // self.at(entry.name()).get(/* serve file */)
+    //     }
+    //     todo!();
+    // }
+
```
goto-bus-stop, line 78: I guess this can be removed?

yoshuawuyts:
@pickfire `http-types` has the following body methods: [docs.rs/http-types?search=body](https://docs.rs/http-types/1.2.0/http_types/?search=body). There are currently four methods:

- `set_body` to set a body.
- `swap_body` to replace a body with a different body through a reference.
- `replace_body` to replace body with a different body through ownership.
- `body_string` to read the body as a string (we plan to make this respect encodings as well).

I suspect we'll be adding more methods to request and response for reading the body as various types. Likely candidates include: `json`, `form`, and `bytes`. But setting the body will almost exclusively go through `set_body` which takes `Into<Body>` which has defined conversions from many types. This should make setting a body overall a smooth process, and is the reason why I chose to go with `set_body` in this patch as well (also note: none of the existing methods worked; I actually spent a good amount of time debugging why filed with a known length were getting chunk-encoded).

pickfire:
@yoshuawuyts But why do we need so many different methods? Couldn't that be just `body` (shared reference) and `body_mut` (unique reference)?

yoshuawuyts:
@pickfire in the http-types model both `Request` and `Response` provide direct access to the body and headers. This has some similarities to Ruby's net::http model.

Personally I really dislike the `body/body_mut` distinction since it leads to patterns like these:

```rust
*req.body_mut() = MyBody;
```

Std doesn't use this pattern anywhere, and instead prefers the distinction between `body` and `set_body`. But where possible (e.g. in `io::Cursor`) it may just forward trait impls such as `Read/Write` directly and only provide access to the inner value through an `into_inner` method. This is the model we've decided to follow instead since it provides a reduced surface area overall.

The convenience methods I mentioned earlier are really only to move beyond the basics. Being able to automatically deserialize from JSON while respecting encodings is pretty hard to do, and providing a shorthand for this is useful. This goes for many of the other methods as well.

src/server/route.rs
```diff
@@ -92,6 +122,33 @@ impl<'a, State: 'static> Route<'a, State> {
...
+    ///     app.at("/public/images").serve_dir("images/")?;
+    ///     app.listen("127.0.0.1:8080").await?;
+    ///     Ok(())
+    /// }
+    /// ```
+    pub fn serve_dir(&mut self, dir: impl AsRef<Path>) -> io::Result<()> {
+        // Verify path exists, return error if it doesn't.
+        let dir = dir.as_ref().to_owned().canonicalize()?;
+        let prefix = self.path().to_string();
+        self.at("*").get(ServeDir::new(prefix, dir));
```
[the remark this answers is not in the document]
yoshuawuyts: The value of `prefix` here is the route that's been used so far. E.g. prefix equates to `/foo` in the following invocation:

```rust
let mut app = tide::new();
app.at("/foo").serve_dir(...);
```
The code you're proposing would mean we'd handle routes on `/foo/foo`, while we need to handle for `/foo/*` instead.

src/server/serve_dir.rs
```diff
@@ -0,0 +1,79 @@
...
+    /// Create a new instance of `ServeDir`.
+    pub(crate) fn new(prefix: String, dir: PathBuf) -> Self {
+        Self { prefix, dir }
+    }
+}
+
+impl<State> Endpoint<State> for ServeDir {
+    fn call<'a>(&'a self, req: Request<State>) -> BoxFuture<'a, Result<Response>> {
+        let path = req.uri().path();
+        let path = path.replace(&self.prefix, "");
```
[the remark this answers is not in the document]
yoshuawuyts: Good catch; looks like we can use `replacen` here instead.

src/response/into_response.rs
```diff
@@ -43,18 +43,22 @@ pub trait IntoResponse: Send + Sized {
...
     }
 }
 
 impl<State: Send + Sync + 'static> IntoResponse for Request<State> {
     fn into_response(self) -> Response {
-        Response::new(StatusCode::Ok).body(BufReader::new(self))
+        let mut res = Response::new(StatusCode::Ok);
+        let len = self.len();
+        let body = Body::from_reader(BufReader::new(self), len);
+        res.set_body(body);
```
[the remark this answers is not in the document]
yoshuawuyts: I've reverted this change instead. We may want to get rid of this whole submodule.

src/response/into_response.rs
```diff
@@ -43,18 +43,22 @@ pub trait IntoResponse: Send + Sized {
...
-            .set_header(
-                http_types::headers::CONTENT_TYPE,
-                "text/plain; charset=utf-8",
-            )
-            .body_string(self)
+        let mut res = Response::new(StatusCode::Ok).set_header(
+            http_types::headers::CONTENT_TYPE,
+            "text/plain; charset=utf-8",
+        );
+        res.set_body(self);
```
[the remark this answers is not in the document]
yoshuawuyts: I've reverted this change instead. We may want to get rid of this whole submodule.

src/server/route.rs
yoshuawuyts: Good catch; thanks!

src/server/route.rs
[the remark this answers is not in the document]
rkusa: Gotcha, I missed that the route for serve_dir is added relative to its parent and not the root. The following seems to work though, which re-uses another prefix-removal feature that I saw in tide.

```
self.at("/").strip_prefix().get(ServeDir::new(dir));
```

PR #419, opened 2020-03-15
Middleware and logging docs changes

drewish:
I'd like to offer up some docs changes from the perspective of an experienced programmer that is very new to rust. I'm trying to use tide for a simple project to webhooks and got tripped up by a few things:
* I wanted to enable some basic logging and it seemed like an oversight that the crate's overview didn't specifically show how to use the middleware. I added a code sample to try to demonstrate that. 
* I didn't realize the `RequestLogger` needed additional crates to actually log output so I added a couple comments there.
* I tried using the middleware example to write my own logger... and then realized that "simplified example" meant "won't compile". I found an [example in the tests](https://github.com/http-rs/tide/blob/master/tests/nested.rs#L37-L45) and realized it wasn't that much more complicated to show a working example.


drewish:
Looks like something timed out in that test. Not seeing a way to restart it so going to see if closing and reopening the PR will trigger it.

Issue #365, opened 2019-12-09
fn into_response call set_header twice


gogobook:
one time by `set_header()`, one time by `set_mime(mime::TEXT_HTML_UTF_8) in `fn body_string`

yoshuawuyts:
Oh, yeah we should probably check if a header has been set already before trying to set it. Good catch!

PR #420, opened 2020-03-16
fix(response): do not set the same header twice with body_string #365

bnjjj:
I tried to solve the issue #365 let me know if I should implement this in another way.

Signed-off-by: Benjamin Coenen <5719034+bnjjj@users.noreply.github.com>


drewish:
Seems like you'd want to address that inside set_mime so it would apply everywhere.

PR #423, opened 2020-03-23
Add optional simd-json support

Licenser:
This adds the simd-json as a feature flag to use SIMD accelerated JSON parsing.

This might warrant a note in the README about cpu features required (AVX2 or SSE4.1) to use the flag but I didn't find a section in the README on features and wasn't sure if you want one added so I'll just open this for suggestions :).

( same as https://github.com/http-rs/surf/pull/154 )

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,7 @@ futures = { version = "0.3.1", optional = true }
 http = { version = "0.2.0", optional = true }
 tokio = { version = "0.2.13", optional = true }
 url = "2.1.1"
+simd-json = { version = "0.3.2", optional = true }
 
 [dev-dependencies]
 async-std = { version = "1.4.0", features = ["unstable", "attributes"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ futures = { version = "0.3.1", optional = true }
 http = { version = "0.2.0", optional = true }
 tokio = { version = "0.2.13", optional = true }
 url = "2.1.1"
-simd-json = { version = "0.3.2", optional = true }
+simd-json = { version = "0.3.7", optional = true }
 
 [dev-dependencies]
 async-std = { version = "1.4.0", features = ["unstable", "attributes"] }
--- a/src/response/mod.rs
+++ b/src/response/mod.rs
@@ -142,8 +142,15 @@ impl Response {
     /// # Mime
     ///
     /// The encoding is set to `application/json`.
+    #[cfg(not(feature = "simd-json"))]
     pub fn body_json(mut self, json: &impl Serialize) -> serde_json::Result<Self> {
         self.res.set_body(serde_json::to_vec(json)?);
+        Ok(self.set_mime(mime::APPLICATION_JSON))
+    }
+
+    #[cfg(feature = "simd-json")]
+    pub fn body_json(mut self, json: &impl Serialize) -> simd_json::Result<Self> {
+        *self.res.body_mut() = simd_json::to_vec(json)?.into();
         Ok(self.set_mime(mime::APPLICATION_JSON))
     }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -166,6 +166,9 @@
 //! changes we make. Also as time goes on you may find that fewer and fewer changes occur, until we
 //! eventually remove this notice entirely. The goal of Tide is to build a premier HTTP experience
 //! for Async Rust. We have a long journey ahead of us. But we're excited you're here with us!
+//!
+//! # Features
+//! - __`simd-json`:__ uses simd-json instead of serde for json parsing. Note: you got to pass `-C target-feature=+avx,+avx2,+sse4.2` to `RUSTFLAGS` if this is enabled.
 
 #![cfg_attr(feature = "docs", feature(doc_cfg))]
 // #![warn(missing_docs)]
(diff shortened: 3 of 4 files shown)


Licenser:
those look like nightly related errors

Cargo.toml
```diff
@@ -41,6 +41,7 @@ serde_qs = "0.5.0"
 async-std = { version = "1.0.1", features = ["unstable"] }
 mime = "0.3.14"
 cookie = { version="0.12.0", features = ["percent-encode"]}
+simd-json = { version = "0.3.2", optional = true }
```
pickfire, line 44: If `simd-json` is enabled, shouldn`t `serde_json` be disabled?

Licenser: serde_json is still used for encoding data, given the performance gains will be minimal compared to decoding we've not gotten around to that part :)

Licenser: I took your inspiration here and added a serializer to simd-json, I stand corrected there is a decent performance benefit in that :)

dignifiedquire:
any reason not to enable this by default?

src/response/mod.rs
```diff
@@ -130,11 +130,18 @@ impl Response {
     /// # Mime
     ///
     /// The encoding is set to `application/json`.
+    #[cfg(not(feature = "simd-json"))]
     pub fn body_json(mut self, json: &impl Serialize) -> serde_json::Result<Self> {
         *self.res.body_mut() = serde_json::to_vec(json)?.into();
         Ok(self.set_mime(mime::APPLICATION_JSON))
     }
```
pickfire, line 137: Why not just
```rust
    pub fn body_json(mut self, json: &impl Serialize) -> serde_json::Result<Self> {
        #[cfg(not(feature = "simd-json"))]
        *self.res.body_mut() = serde_json::to_vec(json)?.into();
        #[cfg(feature = "simd-json")]
        *self.res.body_mut() = simd_json::to_vec(json)?.into();
        Ok(self.set_mime(mime::APPLICATION_JSON))
    }
```

Cargo.toml
pickfire: Oh, so even for serializing there is a decent performance benefit over `serde-json` too?

Licenser: Ja here is a run of json-benchmark w/ simd-json serialisation:
```
                                DOM                  STRUCT
======= serde_json ======= parse|stringify ===== parse|stringify ====
data/canada.json         250 MB/s   380 MB/s   510 MB/s   310 MB/s
data/citm_catalog.json   400 MB/s   360 MB/s   930 MB/s   430 MB/s
data/twitter.json        290 MB/s   530 MB/s   590 MB/s   600 MB/s
data/log.json            360 MB/s   720 MB/s   720 MB/s   720 MB/s

======= simd_json ======== parse|stringify ===== parse|stringify ====
data/canada.json         510 MB/s   440 MB/s   680 MB/s   420 MB/s
data/citm_catalog.json  1030 MB/s   550 MB/s  1470 MB/s   460 MB/s
data/twitter.json        910 MB/s   750 MB/s  1050 MB/s   640 MB/s
data/log.json           1080 MB/s  1080 MB/s  1080 MB/s  1080 MB/s

===== simd_json_tape ===== parse|stringify ===== parse|stringify ====
data/canada.json         750 MB/s                    
data/citm_catalog.json  1790 MB/s                    
data/twitter.json       1620 MB/s                    
data/log.json           2170 MB/s                    
```

src/response/mod.rs
Licenser: That's not legal rust at the moment, attributes on expressions are not supported :)

Cargo.toml
Licenser: And looking at the results I really want to make a custom derive version for `simd-json`, it seems that `serde-derive::Serialize` does degrade performance things down given that serializing a DOM object is faster.

pickfire: Ah, that's interesting. It seemed like DOM is faster than struct for some of the `simd-json` benchmark.

Licenser: Yap the DOM serialization bypasses the traits and serialized directly.

pickfire: By the way, why not add it to actix too or the techempower benchmarks, I believe it can improve the JSON serialization ranks. IIRC last time actix was the first in JSON serialization now it dropped to 10th.

https://www.techempower.com/benchmarks/#section=test&runid=c7013ea3-22c9-44c1-864f-949509d78e0c&hw=ph&test=json
https://github.com/TechEmpower/FrameworkBenchmarks/blob/cfbbf5d11143c96851620eb1aeaf3c2893e862c9/frameworks/Rust/actix/src/main.rs#L23

Licenser: Looking at that code I don't think it'd make a big difference it'd make a bigger difference not to use struct serialization altogether, the message is tiny so it'd not get much of a benefit from SIMD instructions. I don't want to completely generalize it but at least for simd-json, it'd never get to the SIMD part of the code as the key and value are shorter than a single register. Might Still worth looking into from the perspective of avoiding the Serialize train by using a DOM object instead.

Licenser: I created a tracking issue on the serializer: https://github.com/simd-lite/simd-json/issues/121

pickfire: How about the other link for the techempower benchmark?

Licenser: Since it's not really related and to not spam this PR with it I created https://github.com/simd-lite/simd-json/issues/122 for that topic

PR #424, opened 2020-03-27
Fixed typos in documentation.

humb1t:
In several methods `request` were used, but `response` should be.


src/response/mod.rs
```diff
@@ -158,7 +158,8 @@ impl Response {
         self.set_mime(mime::TEXT_PLAIN_UTF_8)
     }
 
-    /// Pass raw bytes as the request body.
+
+    /// Pass raw bytes as the response body.
```
yoshuawuyts, line 162:
```suggestion
    /// Pass raw bytes as the response body.
```

Fishrock123:
Thanks but unfortunately the code here has changed and these changes are no longer relevant.

PR #428, opened 2020-04-09
update Response::body documentation

jbr:
This updates a few typos in Response (referring to the response as a request), clarifies that any reader which is BufRead can be provided to Response::body, and provides an example using a File.  I was especially unsure if the File example was appropriate, but wanted something that was closer to real-world usage than building an arbitrary Vec<u8>. Possible additional examples might include the use of a Cursor or passing a &[u8] to Response::body.
Thanks!

### DIFF
--- a/src/response/mod.rs
+++ b/src/response/mod.rs
@@ -38,7 +38,8 @@ impl Response {
         }
     }
 
-    /// Create a new instance from a reader.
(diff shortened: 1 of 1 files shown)


jbr:
closing because #424 addresses the only remaining documentation concern

PR #436, opened 2020-04-18
Example: How to use `.bytes` with static and non-static data

phansch:
It wasn't entirely clear to me how to build a response from dynamically
generated data. It's easy with a `&'static [u8]` but it seems for
non-static data a `Cursor` is needed?

### DIFF
--- a/examples/bytes-response.rs
+++ b/examples/bytes-response.rs
@@ -1,0 +1,21 @@
+use async_std::{io, task};
+use http_types::StatusCode;
+
+static STATIC_DATA: &[u8] = &[4, 2];
+
+fn main() -> Result<(), std::io::Error> {
+    task::block_on(async {
+        let mut app = tide::new();
+
+        app.at("/bytes").get(|_req: tide::Request<()>| async move {
+            // For static data you can pass it directly:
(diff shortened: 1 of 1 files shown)


phansch:
I'm not sure if that failure is related?

```
test echo_server ... FAILED

failures:

---- echo_server stdout ----
Error: Curl("Unsupported protocol"): Unsupported protocol
thread 'echo_server' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', <::std::macros::panic macros>:5:6
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

yoshuawuyts:
>  but it seems for non-static data a Cursor is needed?

Yeah, that's right. I think the right way to go about it is to add a `from_bytes` method to [`http_types::Body`](https://docs.rs/http-types/1.2.0/http_types/struct.Body.html).

phansch:
> Would you be willing to file a PR for that?

Sure! I don't have much experience with writing library APIs, so I might ask some questions =)

phansch:
Going to open a new PR once https://github.com/http-rs/http-types/pull/100 is merged and usable in tide :ocean:

PR #439, opened 2020-04-18
Fix CORS behaviour if there is no origin header

rkusa:
The CORS origin header handling fell back to an empty header value. This value is considered valid for `Origin::Any` (wildcard origin). However, if the CORS middleware is restricted to a whitelist of origins, the `is_valid_origin` check correctly fails for an empty origin header
value. Though, since we fell back to an empty value when there is no origin header at all, we fail the validation and consequentially respond with a 403 (forbidden) to such requests.

The [CORS specifications](https://www.w3.org/TR/cors/#resource-processing-model) stats the following for simple cross-origin requests, actual requests and preflight requests:

> If the Origin header is not present terminate this set of steps. The request is outside the scope of this specification.

This PR therefore updates the CORS middleware to ignore requests that have no origin header set.

### DIFF
--- a/src/middleware/cors.rs
+++ b/src/middleware/cors.rs
@@ -149,13 +149,16 @@ impl Cors {
 impl<State: Send + Sync + 'static> Middleware<State> for Cors {
     fn handle<'a>(&'a self, req: Request<State>, next: Next<'a, State>) -> BoxFuture<'a, Response> {
         Box::pin(async move {
-            let origins = req
-                .header(&headers::ORIGIN)
-                .cloned()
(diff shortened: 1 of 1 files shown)


src/middleware/cors.rs
```diff
@@ -395,7 +398,7 @@ mod test {
     #[test]
     fn not_set_origin_header() {
         let mut app = app();
-        app.middleware(Cors::new());
+        app.middleware(Cors::new().allow_origin(ALLOW_ORIGIN));
```
rkusa, line 401: The test succeeded before because the default setting is `Origin::Any` which is successful for empty origin header values (as described in the PR message). Adding a specific header was necessary to make the test fail (before this PR).

PR #442, opened 2020-04-20
Remove intermediate Service struct

yoshuawuyts:
This implements `HttpService` directly on `Server` removing the need for the intermediate `Service` struct. This allows us to remove `into_http_service` and the `server` submodule making e.g. nesting apps really easy.

## Examples

```rust
let mut app = tide::new();
let mut inner_app = tide::new();
inner_app.at("/").get(|_| async { Ok("hello world") });
app.at("/foo").nest(inner_app);
```

## Future directions

In https://github.com/http-rs/http-types/pull/94 we've introduced the (experimental) Client and Server traits. The hope is that these can replace `http-client` and `http-service` including on Tide directly. Because these traits don't carry a connection this would allow us to bypass `http_service_mock::make_server` and `http_mock_service::TestBackend::simulate` calls entirely by simply calling `app.recv_req` instead, making it really easy to test applications:

```rust
let mut app = tide::new();
app.at("/").get(|_| async { Ok("hello world") });

let req = Request::new(Method::Get, "http://localhost:8080");
let res = app.recv_req(req).await?;
```

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -274,19 +274,6 @@ impl<State: Send + Sync + 'static> Server<State> {
         self
     }
 
-    /// Make this app into an `HttpService`.
-    ///
-    /// This lower-level method lets you host a Tide application within an HTTP
-    /// server of your choice, via the `http_service` interface crate.
-    pub fn into_http_service(self) -> Self {
-        self
-        // Service {
-        //     router: Arc::new(self.router),
-        //     state: Arc::new(self.state),
-        //     middleware: Arc::new(self.middleware),
-        // }
-    }
-
     /// Asynchronously serve the app at the given address.
     #[cfg(feature = "h1-server")]
     pub async fn listen(self, addr: impl ToSocketAddrs) -> std::io::Result<()> {
@@ -294,9 +281,7 @@ impl<State: Send + Sync + 'static> Server<State> {
 
         let addr = format!("http://{}", listener.local_addr()?);
         log::info!("Server is listening on: {}", addr);
-        let http_service = self.into_http_service();
-
-        let mut server = http_service_h1::Server::new(addr, listener.incoming(), http_service);
+        let mut server = http_service_h1::Server::new(addr, listener.incoming(), self);
 
         server.run().await
     }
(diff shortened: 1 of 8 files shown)


src/server/mod.rs
```diff
@@ -270,49 +269,25 @@ impl<State: Send + Sync + 'static> Server<State> {
     /// Middleware can only be added at the "top level" of an application,
     /// and is processed in the order in which it is applied.
     pub fn middleware(&mut self, m: impl Middleware<State>) -> &mut Self {
-        self.middleware.push(Arc::new(m));
+        let middleware = Arc::get_mut(&mut self.middleware).unwrap();
```
dignifiedquire, line 272: Uhm why are these `Arc::get_mut` `unwrap`s safe? Pretty sure you are not guranteed to be the only holder here.

yoshuawuyts: > Pretty sure you are not guranteed to be the only holder here.

Notice how we're not cloning the Arc, we're only giving a reference to the inner value. This is the same semantics we currently have. The `Arc` is only cloned once we start listening for requests, after which `self` is taken by value and becomes unavailable.

Maybe we can turn this into an `expect` so in the off-chance someone tries something bizarre it fails with a better error. But this is effectively the same semantics we had before.

yoshuawuyts: Okay, added error messages for the failures. I think this should do it!

yoshuawuyts: Oh also for reference: this is how we used to go between `Server` and `Service` before:

```rust
    pub fn into_http_service(self) -> Service<State> {
        Service {
            router: Arc::new(self.router),
            state: Arc::new(self.state),
            middleware: Arc::new(self.middleware),
        }
    }
```

All it does is put the extra fields in an `Arc` and make certain methods available. Those methods currently still take `self` / `&mut self` which means we're protected by the borrow checker. The main difference is that we now start by putting the values in an `Arc` in the first place instead of moving them into one later. This is semantically the same.

dignifiedquire: I see, thank you

PR #443, opened 2020-04-20
make Request::rest public outside of the crate

### LINKED ISSUE #426: Request::rest / *--tide-path-rest
Does tide support unnamed star routes, like `"/static/*"`? It [seems like](https://github.com/http-rs/route-recognizer/blob/master/src/lib.rs#L393-L405) route-recognizer does, and like tide [saves them in --tide-path-rest](https://github.com/http-rs/tide/blob/master/src/server/route.rs#L145-L146), but [tide::Request::rest](https://github.com/http-rs/tide/blob/master/src/request.rs#L181-L185) is only pub(crate). As far as I can tell, `--tide-path-rest` and `Request::rest` are not used anywhere in the crate. Should Request::rest be public?

jbr:
closes #426

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -178,7 +178,7 @@ impl<State> Request<State> {
             .parse()
     }
 
-    pub(crate) fn rest(&self) -> Option<&str> {
+    pub fn rest(&self) -> Option<&str> {
         self.route_params
             .last()
             .and_then(|params| params.find("--tide-path-rest"))


yoshuawuyts:
You can use [`app.all`](https://docs.rs/tide/0.7.0/tide/struct.Route.html#method.all) to match everything:

```rust
app.at("*").all(endpoint);
```

Is there still something missing?

jbr:
This is a different concern.  Sorry I was unclear. Within a route handler with a "/nested/*" route, how do I get access to the entirety of the globbed part?

vladan:
Haven't quite understood your problem, but maybe you're aiming for `/nested/*path` which will give you the rest of the request path under `/nested/` in a `path` param. You can then get it as any other param with `req.param("path")...`.

jbr:
As I mentioned in the issue, the param parser supports unnamed star globs, and they’re also mentioned in the tide docs, and as far as I can tell, tide stores them and has a method that is unused within the crate which does exactly this, but is pub(crate).

jbr:
Perhaps the solution is just to remove mention of unnamed globs from the docs, also the rest method and `--tide-path-rest`. It seems quite reasonable not to support them

PR #445, opened 2020-04-20
Documentation: Add "Getting Started" section to Readme and lib.rs.

Walther:
Mentions `async-std` and its `features = ["attributes"]` clearly.
Fixes https://github.com/http-rs/tide/issues/431

Let me know if there are any wording changes or something else you'd want to improve in this PR!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,6 +12,15 @@
 //! - __Minimal:__ With only a few concepts to learn, Tide is easy to pick up and become productive
 //!     with.
 //!
+//! # Getting started
+//!
+//! Add two dependencies to your project's `Cargo.toml` file: `tide` itself, and `async-std` with the feature `attributes` enabled:
+//! ```text
+//!# Example, use the version numbers you need
+//!tide = "0.7.0"
+//!async-std = { version = "1.5.0", features = ["attributes"] }
+//!```
+//!
 //! # Examples
 //!
 //! __hello world__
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -12,6 +12,15 @@
 //! - __Minimal:__ With only a few concepts to learn, Tide is easy to pick up and become productive
 //!     with.
 //!
+//! # Getting started
+//!
+//! Add two dependencies to your project's `Cargo.toml` file: `tide` itself, and `async-std` with the feature `attributes` enabled:
+//! ```text
```
yoshuawuyts, line 18:
```suggestion
//! ```toml
```

Walther: Thanks! Wasn't sure if the docs supported toml - when running `cargo doc` I had a warning of
```
warning: doc comment contains an invalid Rust code block
    |
    = help: mark blocks that do not contain Rust code as text: ```text
```
but on a second thought that might have been from some other location or even from some other dependency's docs as `cargo doc` builds all of them by default.

It was kinda amusing trying to look for what the docs support - didn't find docs.rs itself in docs.rs, the about page and the repository README seemed to be mostly about running the tool/service and not the doc language specification. A search in the Rust book for ```` ```text ```` didn't get many results. Lastly, I checked a couple of random libraries and they seemed to use ```` ```text ```` for their toml-y bits, so I just used that to be safe.

Glad to hear ```` ```toml ```` works though!

PR #446, opened 2020-04-21
bind to an unused port in integration tests

jbr:
Prior to this PR, if a machine has a server running on any of the ports used for testing (808x), tests would fail in unclear ways.  Because socket reuse flags are platform dependent, this PR finds an unused port by temporarily binding to port 0 and using the assigned port.

### DIFF
--- a/tests/chunked-encode-large.rs
+++ b/tests/chunked-encode-large.rs
@@ -1,3 +1,4 @@
+mod test_utils;
 use async_std::io::Cursor;
 use async_std::prelude::*;
 use async_std::task;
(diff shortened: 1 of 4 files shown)


tests/chunked-encode-large.rs
```diff
@@ -67,22 +68,23 @@ const TEXT: &'static str = concat![
 
 #[async_std::test]
 async fn chunked_large() -> Result<(), http_types::Error> {
-    let server = task::spawn(async {
+    let bind = test_utils::determine_port_to_bind().await;
```
yoshuawuyts, line 71: Can we rename the `bind` variable to `port`?

tests/test_utils.rs
```diff
@@ -0,0 +1,7 @@
+pub async fn determine_port_to_bind() -> async_std::net::SocketAddr {
```
yoshuawuyts, line 1: Perhaps a shorter name would be `find_port`.

tests/chunked-encode-large.rs
jbr: would socket_addr work? I initially had `port` (as still remains in the function name, which will also be fixed), but it's both `localhost` as well as the port

tests/test_utils.rs
jbr: as with above comment, would `find_socket_addr` be reasonably concise and clear? it's really something like "find a port and return a socket addr with that port" but that's awkward af

PR #456, opened 2020-04-22
Add initial SSE support

### LINKED ISSUE #234: Support for server-sent events
## Feature Request

Support for server-sent events in Tide apps.

## Detailed Description

Support for working creating server-sent event streams, easily, with Tide.

Some more general server-sent events info:

* Standard: https://html.spec.whatwg.org/multipage/#server-sent-events
  * Grammar: https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream
* Using Server-Sent Events: https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events

## Context

I would like to use server-sent events. I suppose other folks might want to as well.

## Possible Implementation

Ideally, I would be able to have some method on `Response` or some sort of type that implements `IntoResponse` that sets the `Content-Type: text/event-stream` header for me, and takes a `Stream<Item = ServerSentEvent>` where `ServerSentEvent` is a struct with an optional event type string and a data string.

Maybe it would automatically send heartbeats too, to keep the connection alive?

yoshuawuyts:
This adds initial support for Server-Sent Events. It provides both a pre-made endpoint handler, and an `upgrade` function for use inside manual endpoints. This is the first step towards more elaborate SSE handling. Closes #234. 

## Future Directions

As described in my [tide channels post](https://blog.yoshuawuyts.com/tide-channels/) there are many more things we could do around channels. But those rely on having a notion of _identity_ for connecting clients, which in turn relies on session support. Once we have that we can add persistent connection IDs and more!

Additionally we're not sending automated hearbeat messages yet, I believe we're still missing support for these in [http-rs/async-sse](https://github.com/http-rs/async-sse).

## Screenshots

Screenshot_2020-04-22 tide sse - Rust

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ docs = ["unstable"]
 unstable = []
 
 [dependencies]
+async-sse = "2.1.0"
 http-types = "1.0.1"
 http-service = "0.5.0"
 http-service-h1 = { version = "0.1.0", optional = true }
--- a/examples/sse.rs
+++ b/examples/sse.rs
@@ -4,8 +4,8 @@ use tide::sse;
 async fn main() -> Result<(), std::io::Error> {
     let mut app = tide::new();
     app.at("/sse").get(sse::endpoint(|_req, sender| async move {
-        sender.send("fruit", "banana").await;
-        sender.send("fruit", "apple").await;
+        sender.send("fruit", "banana", None).await;
+        sender.send("fruit", "apple", None).await;
         Ok(())
     }));
     app.listen("localhost:8080").await?;
--- a/examples/sse.rs
+++ b/examples/sse.rs
@@ -1,0 +1,13 @@
+use tide::sse;
+
+#[async_std::main]
+async fn main() -> Result<(), std::io::Error> {
+    let mut app = tide::new();
+    app.at("/sse").get(sse::endpoint(|_req, sender| async move {
+        sender.send("fruit", "banana").await;
+        sender.send("fruit", "apple").await;
+        Ok(())
+    }));
+    app.listen("localhost:8080").await?;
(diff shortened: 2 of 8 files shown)


jbr:
I'm curious about what this would look like in a use case where the events are triggered elsewhere in an application (which I assume is the vast majority of the reason to use SSE). If each connection gets a new async-sse sender/encoder pair, is the idea that within the sse:endpoint fn, users would do something like this?

```rust
app.at("/sse").get(sse::endpoint(|req: Request<SomeSharedEventStream>, sender: Sender| async move {
  while let Some(event) = req.state().recv().await {
    sender.send(event.name, event.data_as_u8_slice(), event.id);
  }
  Ok(())
}))
```

yoshuawuyts:
@jbr as I wrote in my post on Tide Streams we can provide high-level APIs for this, but we first need a notion of identity so we can reliably match incoming requests to existing connections. But what you shared so far could indeed be a way to attempt at building something like that before we fully support it.

goto-bus-stop:
I think the `id` argument to `encoder.send()` should prob not be hardcoded to `None`. You need that if you want to implement reconnections correctly without dropping messages, and a future session-based API would likely need to do just that.

In that vein, it might be handy to store the "Last-Event-ID" header value on the `sender` object, so you can access it easily:
```rust
sse::upgrade(req, |req, sse| async move {
  if let Some(event_id /*: &str */) = sse.last_event_id() {
    // handle that
  }
})
```
That could be a later addition since you can also find it on the `req` object already.

yoshuawuyts:
@goto-bus-stop good call! -- I've changed `Sender` so it now takes an optional third argument of id, matching the API in async-sse.

Issue #449, opened 2020-04-21
rfc/proposal: Add support for a content type map in ServeDir


jbr:
# Use case:

To ergonomically serve arbitrary files within a directory with specific mime types that may not be recognized by whatever best-effort mime guessing strategy tide uses

# Prior art:
- [https://hexdocs.pm/plug/Plug.Static.html#module-options] (:content_types key in elixir's Plug.Static)
accepts a simple map of file names to mime types. Advantages: Simple, straightforward, efficient. Disadvantages: No way to express "anything that ends with this extension should have this mime type." Possible improvements: Accept globs in the keys

- [http://expressjs.com/en/resources/middleware/serve-static.html](setHeaders option in node's express serve-static library) This is a function with the signature `(res, path, stat) → undefined`. Advantages: Useful for more than just setting content types, is a general purpose map on the response. Disadvantage: Potentially a lot of boilerplate for just "I want this custom file extension to have this mime type."

- [http://nginx.org/en/docs/http/ngx_http_core_module.html#types](nginx) just takes a map from file extension to mime type, but it's easy enough to carve out a specific path glob in nginx config to override this for a file that doesn't have an extension

---

# Proposal for tide:
Add something along the lines of:

```rust
app.at("/static")
  .serve_dir("./files")? // ← returns the ServeStatic or a builder instead of ()
  .with_content_type_mapping("*.custom", "text/custom")
  .with_content_type_mapping("special-file", "application/just-this-file")
```

Informing this proposal is an assumption based on other static file middlewares that there will likely need to be many options passed to ServeDir eventually.  [Plug.Static](https://hexdocs.pm/plug/Plug.Static.html#module-options) is a great example of the affordances users might reasonably expect from a directory endpoint eventually, including in-memory caching, different compression schemes, and cache control settings, including the configuration mentioned in #447

yoshuawuyts:
Really happy with the design we came up with yesterday; feels like it provides a solid foundation we can implement, and then use to iterate on.

A few further design thoughts:
- `tide::fs::File` may conflict with `async_std::fs::File` -- perhaps we should name it `tide::fs::FileEndpoint`?
    - Though a counterpoint is that it's not really an endpoint, and somewhat long to type.
- The starting point for these is probably to implement `serve_file` as well, and start placing these in a new (private) `fs` submodule.
- I think we should guess mimes by default.
- `filter` is indeed a key operation we should support
- we talked about `map` as an operation too, but having a `middleware` method is strictly more powerful
[...]

PR #461, opened 2020-04-23
🐞 add a minimal version of mime guessing to serve_dir

jbr:
this is probably just a stopgap, but resolves the issue wherein browsers download text files served with a generic binary stream content type

### DIFF
--- a/src/server/serve_dir.rs
+++ b/src/server/serve_dir.rs
@@ -69,6 +69,11 @@ impl<State> Endpoint<State> for ServeDir {
             // TODO: fix related bug where async-h1 crashes on large files
             let mut res = Response::new(StatusCode::Ok);
             res.set_body(body);
+
+            if let Some(content_type) = mime_guess::from_path(&file_path).first() {
+                res = res.set_mime(content_type);
+            }
+
             Ok(res)
         })
     }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,13 +39,14 @@ serde_json = "1.0.41"
 serde_qs = "0.5.0"
 async-std = { version = "1.4.0", features = ["unstable"] }
 mime = "0.3.14"
-cookie = { version="0.12.0", features = ["percent-encode"]}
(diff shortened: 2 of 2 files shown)


src/server/serve_dir.rs
```diff
@@ -73,6 +73,11 @@ impl<State> Endpoint<State> for ServeDir {
             // TODO: fix related bug where async-h1 crashes on large files
             let mut res = Response::new(StatusCode::Ok);
             res.set_body(body);
+
+            if let Some(content_type) = mime_guess::from_path(&path).first() {
+                res = res.set_mime(content_type);
```
jbr, line 78: this awkward line is because set_mime takes self. could it it take &mut self?

yoshuawuyts: We'll want to replace it with https://docs.rs/http-types/1.2.0/http_types/struct.Response.html#method.set_content_type in a future PR. This requires http-types 2.0 to be released first tho, which is something I'm working on right now. I think for the purpose of this patch this is okay (:

src/server/serve_dir.rs
```diff
@@ -31,7 +31,7 @@ impl<State> Endpoint<State> for ServeDir {
         log::info!("Requested file: {:?}", dir);
 
         Box::pin(async move {
-            let file = match async_std::fs::canonicalize(&dir).await {
+            let (file, path) = match async_std::fs::canonicalize(&dir).await {
```
yoshuawuyts, line 34: Sending back a tuple here feels a bit like a hack. I think inlining the mime guessing, while messy, at least won't change the overall code flow. We could even extract it to a function like:

```rust
fn guess_mime(headers: impl AsMut<Headers>, path: &Path) {
    // guess mime here
}
```
That we then call inline.

jbr: What if instead of `(file, path)` this returned a named `FileData` struct as mentioned in https://github.com/http-rs/tide/issues/449?

PR #466, opened 2020-04-24
Add Response::from_res

yoshuawuyts:
This adds `Response::from_res` which enables us to bridge with `Surf` for  authoring proxies conveniently. We couldn't add this as a generic `From` bound because it would conflict with existing implementations.

This was in response to this [this example](https://gist.github.com/jbr/d5c4fec7368eae5a372b6ae403b9c67f) which proxies from surf -> tide, which would allow us to rewrite the core logic from ~10 lines to 2 lines.

```rust
let res = surf::get(...).await?;
Ok(Response::from_res(res))
```

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -247,10 +247,23 @@ impl Response {
         self.res.local_mut().insert(val);
         self
     }
+
+    /// Create a `tide::Response` from a type that can be converted into an
+    /// `http_types::Response`.
+    pub fn from_res<T>(value: T) -> Self
+    where
+        T: Into<http_types::Response>,
+    {
+        let res: http_types::Response = value.into();
+        Self {
+            res,
+            cookie_events: vec![],
+        }
+    }
 }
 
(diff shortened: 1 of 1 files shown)


chrisdickinson:
Would it be possible to have a form of this where we could pass a HashMap to override headers on the way through?

```
Response::from_res(res, hashmap!{
    "ETag" => "a modified etag",
})
```

(Or possibly a lambda to modify them?)

```
Response::from_res(res, |headers: &mut Headers| {
  // edit the headers on the way through
})
```

yoshuawuyts:
The following would be possible with this patch:

```rust
let res = surf::get(...).await?;
let mut res = Response::from_res(res);
res.set_header("etag".parse()?, "<new etag>")?;
Ok(res)
```

In future patches well likely rename `set_header` to `insert_header`, and remove the need for calling `parse()?`. But the overall approach should remain the same.

PR #467, opened 2020-04-24
📖 add a mention of the http/1.1 protocol limitation

jbr:
and provide instructions for using nginx as a reverse proxy

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ $ cargo run --example hello
 
 ## TLS termination and reverse proxies
 
-Tide does not currently support TLS, and only supports HTTP/1.1.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -73,6 +73,19 @@ $ cargo run --example # shows a list of available examples
 $ cargo run --example hello
 ```
 
+## TLS termination and reverse proxies
+
+Tide does not currently support TLS, and only supports HTTP/1.1.
```
yoshuawuyts, line 78:
```suggestion
Tide's default backend currently does not support TLS, and only supports HTTP/1.1.
```

PR #468, opened 2020-04-24
Remove default middlewares from server creation

### LINKED ISSUE #396: Cookies are set multiply times when using *server.nest*

vladan:
Fixes #396 #476 #551 in that it removes the default logging and cookies middlewares from `Server::with_state()` and allows users to use their own middlewares, or use the default ones from the framework by calling `with_cookies()` and/or `with_logging()` on the server instance.

### DIFF
--- a/src/security/cors.rs
+++ b/src/security/cors.rs
@@ -367,7 +367,7 @@ mod test {
 
     #[async_std::test]
     async fn retain_cookies() {
-        let mut app = crate::Server::new();
+        let mut app = crate::Server::new().with_cookies();
         app.middleware(CorsMiddleware::new().allow_origin(ALLOW_ORIGIN));
         app.at(ENDPOINT).get(|_| async {
             let mut res = crate::Response::new(http_types::StatusCode::Ok);
--- a/src/server.rs
+++ b/src/server.rs
@@ -199,14 +199,31 @@ impl<State: Send + Sync + 'static> Server<State> {
     /// # Ok(()) }) }
     /// ```
     pub fn with_state(state: State) -> Self {
-        let mut server = Self {
+        Self {
             router: Arc::new(Router::new()),
             middleware: Arc::new(vec![]),
             state: Arc::new(state),
-        };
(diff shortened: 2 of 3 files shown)


vladan:
The approach is copied from #400 and updated to the latest version.

src/server/mod.rs
```diff
@@ -284,12 +282,25 @@ impl<State: Send + Sync + 'static> Server<State> {
...
+    // Enable loggin and cookie middleware
+    fn enable_default_middleware(&mut self) {
+        self.middleware(cookies::CookiesMiddleware::new());
+        self.middleware(log::LogMiddleware::new());
+    }
+
+    pub fn into_mock(mut self) -> io::Result<http_service_mock::TestBackend<Self>> {
+        self.enable_default_middleware();
+        http_service_mock::make_server(self)
+    }
```
yoshuawuyts, line 294: Oh, this code should probably go. We can probably replace this with the new `http_types::Server` trait instead.

vladan:
resolved by #662

PR #471, opened 2020-04-25
Remove Ok(()) with .await as end statement

pickfire:
Make `app.listen` more ergonomic.

Before

```rust
#[async_std::main]
    let mut app = tide::new();
    Ok(())
}
```

After

```rust
#[async_std::main]
    let mut app = tide::new();
    app.listen("127.0.0.1:8080").await
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,8 +62,7 @@ async-std = { version = "1.5.0", features = ["attributes"] }
 async fn main() -> Result<(), std::io::Error> {
     let mut app = tide::new();
     app.at("/").get(|_| async { Ok("Hello, world!") });
-    app.listen("127.0.0.1:8080").await?;
-    Ok(())
+    app.listen("127.0.0.1:8080").await
 }
 ```
 
--- a/examples/chunked.rs
+++ b/examples/chunked.rs
@@ -11,7 +11,6 @@ fn main() -> Result<(), std::io::Error> {
             let res = Response::new(StatusCode::Ok).body(BufReader::new(file));
             Ok(res)
         });
-        app.listen("127.0.0.1:8080").await?;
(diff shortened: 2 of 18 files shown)


yoshuawuyts:
I'm not sure if we should do this -- `?` marks where errors come from, and that's worthwhile having even if the result is smaller.

The lang team is exploring options to remove the need for `Ok(())` at the end of functions, which I think is what we really want to solve here. Until then I think keeping this as-is may be good enough (and can also serve as an example of where Ok-wrapping could really help).

pickfire:
The output is `Result<(), std::io::Error>` which is exactly the same type as `app.listen("127.0.0.1:8080").await`. I feel like it is easier to read this way since it will exit when there is `Err` returning from `listen`.

I don't think it is necessary to use `?` here, `?` is useful but here it is similar like using `return statement;` as the last statement, so it may just be better to return the result directly.

Licenser:
+1 to the suggestion. 

```rust
    Ok(())
```

makes it unclear what app.listen returns, without looking it up in the docs. do I need to consume something there? Is it something important?

```rust
    app.listen("127.0.0.1:8080").await
```

in contrast makes it clear with the context given that it returns an io::Result<()> so there is no data to look for coming form `app.listen`.

pickfire:
Yes, `?` does a conversion for the error with `From` type that may not be implicit, but `app.listen("127.0.0.1:8080").await` makes it explicit that it returns `io::Error`.

Fishrock123:
We should probably look to having a return value of `Result<!, std::io::Error>` for either tide 1.0 or 2.0, depending on when `never` stabilizes, as it is the best answer to this issue.

PR #473, opened 2020-04-25
fix: check directory traversal attack without file system visit

DCjanus:
We don't want to leak information of which files exist to adversaries, but the origin version of `ServeDir` did.

We are using `canonicalize` to make sure `file_path` is simple, but, unfortunately, it would visit file system, which might be slow, and attackers could count the time that server response, with hundreds of requests, attacker would know which file is exists even it is not sub path of `root`.

read more: https://en.wikipedia.org/wiki/Timing_attack

### DIFF
--- a/src/server/serve_dir.rs
+++ b/src/server/serve_dir.rs
@@ -4,6 +4,7 @@ use crate::{Body, Endpoint, Request, Response, Result, StatusCode};
 use async_std::fs::File;
 use async_std::io::BufReader;
 
+use std::ffi::OsStr;
 use std::path::{Path, PathBuf};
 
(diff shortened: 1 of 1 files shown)


src/server/serve_dir.rs
```diff
@@ -22,43 +23,38 @@ impl ServeDir {
...
+                continue;
+            } else if p == OsStr::new("..") {
+                file_path.pop();
+            } else {
+                file_path.push(&p);
+            }
         }
-        log::info!("Requested file: {:?}", dir);
+
+        log::info!("Requested file: {:?}", file_path);
```
yoshuawuyts, line 39: Ideally we could `warn` here with the full requested path if it contained a `..` instead.

PR #480, opened 2020-04-30
use unwrap_or_default instead of unwrap_or

fundon:


jbr:
What is the advantage of this change?

fundon:
https://doc.rust-lang.org/std/result/enum.Result.html#method.unwrap_or
> Arguments passed to unwrap_or are eagerly evaluated; if you are passing the result of a function call, it is recommended to use unwrap_or_else, which is lazily evaluated.

jbr:
But passing a static str isn’t a function call. Is there actually a performance difference? I would have guessed the compiler would treat them equivalently

PR #481, opened 2020-05-01
Add back multipart support.

JakeChampion:
Multipart support was disabled in the commit -- https://github.com/http-rs/tide/commit/c0d5783a7d8984bc91ac9b2908680d94c7cc57cf 

Multipart support was removed because it was including the `iron` and `nickel` frameworks.

This implementation does not include `iron` or `nickel` frameworks, hopefully that means we can bring the support back 🤞


yoshuawuyts:
Hey @JakeChampion, thanks for opening this PR! Ha; it's looks like the optional deps was an oversight on my part; thanks for pointing that out!

However an issue was opened on http-types about multipart bodies https://github.com/http-rs/http-types/issues/126 and overall it seems like that might be a better place to add this functionality. I drafted an initial design for what that could look like in https://github.com/http-rs/http-types/issues/126#issuecomment-629650395.

If you're still interested in helping get a base version of multipart back into Tide, we could definitely use a hand implementing it in http-types.

Fishrock123:
Looking at https://github.com/abonander/multipart-async, it seems like we probably would want to implement that kind of thing customized directly into `http-types`.

PR #484, opened 2020-05-02
🐞 correctly parse multiple values out of the Cookie header

### LINKED ISSUE #469: Tide mis-parses multiple cookies as a single cookie
I upgraded a personal project to 0.8, and I noticed that my session code began failing. After a while spent debugging, I noticed that there were multiple cookies being sent by my browser and Tide was only returning the first cookie in the `Cookie` header. I wrote up a short demonstration of the issue here: https://github.com/cassiemeharry/tide-0-8-cookie-bug

The output I get:

```sh-session
$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/tide-0-8-cookie-bug`
[src/main.rs:22] req.cookie("a") = Some(
    Cookie {
        cookie_string: Some(
            "a=a-val; b=b-val; c=c-val",
        ),
        name: Indexed(
            0,
            1,
        ),
        value: Indexed(
            2,
            7,
        ),
        expires: None,
        max_age: None,
        domain: None,
        path: None,
        secure: None,
        http_only: None,
        same_site: None,
    },
)
[src/main.rs:23] req.cookie("b") = None
[src/main.rs:24] req.cookie("c") = None
```

I suspect the `cookie_string` value isn't being copied somewhere inside the `Request::cookie` call, and instead the first `Cookie` is taking ownership of it, leaving no header for the remaining calls.

jbr:
closes #469

### DIFF
--- a/src/cookies/middleware.rs
+++ b/src/cookies/middleware.rs
@@ -83,7 +83,7 @@ impl CookieData {
             for cookie_header in cookie_headers {
                 // spec says there should be only one, so this is permissive
                 for pair in cookie_header.as_str().split(";") {
-                    if let Ok(cookie) = Cookie::parse(String::from(pair)) {
+                    if let Ok(cookie) = Cookie::parse_encoded(String::from(pair)) {
                         jar.add_original(cookie);
                     }
                 }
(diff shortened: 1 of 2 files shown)


jbr:
after i wrote this, @fundon noticed that this is a regression from https://github.com/http-rs/tide/blob/v0.6.0/src/middleware/cookies.rs#L82-L103, so the alternative fix would be reverting to that code but with the test in this pr. the code is identical though, and this version might be slightly easier to read

PR #485, opened 2020-05-04
Add Server::bind to bind a server to a TcpListener

connec:
> This makes it easier to start a server on a pre-existing TcpListener, such as one obtained from `listenfd`.

I was going to make an issue to propose something, but it was as easy to simply make a small PR.

Basically I wanted to get auto-reloading working with Tide. The main tool I've found for this is [listenfd](https://crates.io/crates/listenfd), which allows a `std::net::TcpListener` to be extracted from environment variables (which can be supplied by a compatible binary such as [systemfd](https://crates.io/crates/systemfd)).

Actually binding a `tide::Server` to a pre-existing `TcpListener` requires a direct dependency on `http-service-h1` and re-implementing `tide::Server::listen` in your own app, e.g.

```rust
use std::net::TcpListener;

#[async_std::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let mut listenfd = listenfd::ListenFd::from_env();
    let listener = listenfd
        .take_tcp_listener(0)?
        .unwrap_or_else(|| TcpListener::bind("127.0.0.1:8080").expect("failed to bind"));
    let listener = async_std::net::TcpListener::from(listener);

    let mut app = tide::new();

    // set up app

    let addr = format!(
        "http://{}",
        listener.local_addr().expect("failed to get addr")
    );
    let mut server = http_service_h1::Server::new(addr, listener.incoming(), app);

    server.run().await?;

    Ok(())
}
```

This PR splits the h1-server construction from `tide::Server::listen` to `tide::Server::bind`, and `listen` now calls `bind` with the `TcpListener` constructed from the address. With this PR, the above simplifies to:

```rust
use std::net::TcpListener;

#[async_std::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let mut listenfd = listenfd::ListenFd::from_env();
    let listener = listenfd
        .take_tcp_listener(0)?
        .unwrap_or_else(|| TcpListener::bind("127.0.0.1:8080").expect("failed to bind"));

    let mut app = tide::new();

    // set up app

    app.bind(listener).await?;

    Ok(())
}
```

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -2,7 +2,7 @@
 
 use async_std::future::Future;
 use async_std::io;
-use async_std::net::ToSocketAddrs;
+use async_std::net::{TcpListener, ToSocketAddrs};
 use async_std::sync::Arc;
 use async_std::task::{Context, Poll};
 use http_service::HttpService;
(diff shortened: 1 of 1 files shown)


connec:
The Windows tests are failing in `tests/server.rs` (`hello_world` and `echo_server`) due to:

> Only one usage of each socket address (protocol/network address/port) is normally permitted. (os error 10048)

I can't work out how my changes would affect port allocation, so any insights there would be helpful!

jbr:
I think the test failure is likely unrelated, and is possibly somehow related to #446. Maybe force-push to trigger another build?

connec:
@jbr – I rebased and force-pushed and that seems to have done the trick!

connec:
I'm going to close this in favour of #505 to keep the focus on enabling auto-reloading.

PR #487, opened 2020-05-05
Small docstring fix for Response::body_json()

birkenfeld:
I assume the "as a form" is copy-paste from body_form() above.


Fishrock123:
Thanks but this code has moved to `http-types` and this fix is no longer relevant.

https://github.com/http-rs/http-types/blob/f77c653d6703192430b4ba8fb016fe17ba8d457f/src/body.rs#L233-L256

PR #489, opened 2020-05-05
Correct version number in README.md

skade:


jbr:
This is the same as #475

PR #493, opened 2020-05-11
fixes #492

### LINKED ISSUE #492: Index of a nested router not found
When a nested router has an index, the path is ignored, leeding to a 404.

```rust
let mut router = tide::new();
router.at("/api").nest(|router| {
  router.at("/").get(async move |_| "api-root");
  router.at("/resource").get(async move |_| "resource");
});
```
Just adapt the _nested_ example to review.
So, if you package a component together in a nested router, you cannot mount its root path.
PR with fix and tests is on the way though :)

ennersk:
see #492


yoshuawuyts:
@ennersk hi, thanks so much for this PR! Could you perhaps clarify what you meant by https://github.com/http-rs/tide/issues/205#issuecomment-626737560:
> I think it is perfect as it is currently and vote for closing this one.

Are you in favor of this PR, or do you think we should keep things as-is? thanks heaps!

ennersk:
ouch, things are moving fast here :) after running into several conflicts after rebasing I think its better to close this one and submit my little change anew

ennersk:
> @ennersk hi, thanks so much for this PR! Could you perhaps clarify what you meant by [#205 (comment)](https://github.com/http-rs/tide/issues/205#issuecomment-626737560):
> 
> > I think it is perfect as it is currently and vote for closing this one.
> 
> Are you in favor of this PR, or do you think we should keep things as-is? thanks heaps!

Well, in the other ticket it was also about funny trailing slashes, so I linked them.
What I meant in the other issue is that in http the paths "/foo/bar" and with trailing/ "/foo/bar/" identify 2 different resources. Currently tide complies to this - what I like - 

However I understand many developers (like the one opening the ticket) favoring to treat them the same, as it is confusing, but I think tide should keep being compliant, so I voted for just closing #205. 

Maybe printing out all registered routes on start would be helpful, one would immediately spot two routes, one with and one without a trailing slash - and could fix it if it was unintentional. But this would be another issue...

PR #495, opened 2020-05-11
init default logger

yoshuawuyts:
Relies on https://github.com/lrlna/femme/pull/9. Adds a default logger that supports `log`'s key-value API. A base "hello world" example now looks like this:

```rust
#[async_std::main]
    tide::log::start();
    let mut app = tide::new();
    app.at("/").get(|_| async move { Ok("Hello, world!") });
    Ok(())
}
```

---

There's a bit of a design question whether we should expose this as a separate function, or integrate it directly into the framework. I think logging is kind of on the fringe here since it not only captures things going on *inside* the framework; it captures things going on outside of it as well.

I feel that making it a single line to enable logging strikes a good balance between the two -- we make it easy to get a good setup going with minimal work, but we don't enable things that have global implications for the program without consulting the user.

# Screenshots

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ async-h1 = { version = "1.1.2", optional = true }
 async-sse = "2.1.0"
 async-std = { version = "1.4.0", features = ["unstable"] }
 cookie = { version = "0.12.0", features = ["percent-encode"]}
+femme = "2.0.1"
 http-types = "1.0.1"
 kv-log-macro = "1.0.4"
 mime = "0.3.14"
@@ -42,7 +43,6 @@ route-recognizer = "0.1.13"
 serde = "1.0.102"
 serde_json = "1.0.41"
 serde_qs = "0.5.0"
-femme = { path = "../../femme" }
 
 [dev-dependencies]
 async-std = { version = "1.4.0", features = ["unstable", "attributes"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ route-recognizer = "0.1.13"
 serde = "1.0.102"
 serde_json = "1.0.41"
 serde_qs = "0.5.0"
-femme = "1.3.0"
+femme = { path = "../../femme" }
 
 [dev-dependencies]
 async-std = { version = "1.4.0", features = ["unstable", "attributes"] }
--- a/examples/hello.rs
+++ b/examples/hello.rs
@@ -1,10 +1,8 @@
-use async_std::task;
-
-fn main() -> Result<(), std::io::Error> {
-    task::block_on(async {
-        let mut app = tide::new();
-        app.at("/").get(|_| async move { Ok("Hello, world!") });
-        app.listen("127.0.0.1:8080").await?;
-        Ok(())
-    })
+#[async_std::main]
+async fn main() -> Result<(), std::io::Error> {
+    tide::log::start();
+    let mut app = tide::new();
+    app.at("/").get(|_| async move { Ok("Hello, world!") });
+    app.listen("127.0.0.1:8080").await?;
+    Ok(())
 }
--- a/examples/static_file.rs
+++ b/examples/static_file.rs
@@ -1,13 +1,9 @@
-use async_std::task;
-use tide::log;
-
-fn main() -> Result<(), std::io::Error> {
-    femme::start(log::Level::Info.to_level_filter()).unwrap();
-    task::block_on(async {
-        let mut app = tide::new();
-        app.at("/").get(|_| async move { Ok("visit /src/*") });
-        app.at("/src").serve_dir("src/")?;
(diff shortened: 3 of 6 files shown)


dignifiedquire:
Can this be filtered by an env variable? I think that is really a requirement to enabling this by default, especially having a way to make it only log things inside my app or just tide, not all dependencies.

yoshuawuyts:
@dignifiedquire still working on adding filtering on the `femme` through `femme::with_filter`, but it should then be a one-liner to hook up env vars to apply filtering.

jbr:
In reply to the question of whether logging should be integrated: _Eventually_ for a production logger, I think there'll be a lot more API surface area*, possibly justifying splitting this out into an independent crate (even if it's included by a default `logging` feature in tide).

*Other features people might reasonably eventually expect from a feature-complete local logger: Log rotation, [configurable formats](https://github.com/expressjs/morgan), splitting into different files depending on a predicate like status code (similar to nginx's approach to logging), [parameter filtering](https://hexdocs.pm/phoenix/Phoenix.Logger.html#module-parameter-filtering).

yoshuawuyts:
> It would be really nice if the default logger were able to support stuff like that, so that the path from development to production wouldn't be "replace the tide logger" but instead "configure the tide logger."

@jbr I love this phrasing and agree that is indeed the way we should be thinking of driving this feature. As such I probably should add some more context for this logger specifically, explaining the ideas behind it, the direction it's going, and shortcomings it has.

The idea behind `femme` is to create a logger in the tradition of [`winston`](https://github.com/winstonjs/winston), [`bole`](https://github.com/rvagg/bole), and [`pino`](https://github.com/pinojs/pino): using [ndjson](http://ndjson.org/) for logs so that each line can be unambiguously parsed, overcoming much of the problems around common-log formatting. In particular `femme` acts as a hybrid of `pino` and the [`pino-colada`](https://github.com/lrlna/pino-colada) pretty printer. When in development it will pretty-print messages. But in production the output is ndjson-formatted [1]. This allows for easy piping into e.g. logstash.

Because `femme`'s format is based on `pino`, we can use many of the same utilities. For example [log rotation](https://getpino.io/#/docs/help?id=rotate) exists as a tool already. And so does [splitting logs on log level](https://getpino.io/#/docs/help?id=multiple). The pino docs feature many tools and techniques that would likely get us most of the way, right away.

However some things we indeed do not support yet; in particular [redaction](https://getpino.io/#/docs/redaction) seems important. Maybe the right approach here is to have `femme::with_filter` act as some kind of `filter_map` operation that can both modify and edit values. Though a more elaborate override mechanism might be useful as well. We'll need to think this one over probably.

Hopefully this sheds some light on what `femme` is, what some of the ideas behind it are, and what it would enable for Tide today. None of this is to say that implementing `femme` in its current state would be the final word on logging for Tide. But I think it would get us really far out of the box. As we get feedback from users we could look to expend our logging facilities. But hopefully this can help us get off to a good start!

[1]: We probably should do tty detection here as well.

jbr:
I should have looked more into femme, that's really cool! I saw the pretty-printed output in the screenshot and assumed that it was just logging unstructured text, but the ability to repurpose pino's utilities (and transporters!) makes this approach way more feature-complete and production ready than I expected.

yoshuawuyts:
@jbr no worries; I understand the confusion and felt it was on me to share more details here heh 😅

Cargo.toml
```diff
@@ -45,8 +45,9 @@ futures = { version = "0.3.1", optional = true }
 http = { version = "0.2.0", optional = true }
 tokio = { version = "0.2.13", optional = true }
 url = "2.1.1"
-kv-log-macro = "1.0.4"
+kv-log-macro = "1.0.5"
 mime_guess = "2.0.3"
+femme = { path = "../../femme" }
```
isgj, line 50: I'd like to try it out. Can you post which version of `femme` is used?

yoshuawuyts: @isgj https://github.com/lrlna/femme/pull/9

isgj: When I build the app with the release flag the logs look like
```json
{"level":30,"time":1589717147052,"msg":"--> Response sent","method":"\"GET\"","status":"200","duration":"\"157ms\"","path":"\"/api/v1/users\""}
```
the string (of key: value, not the message) have also the double quotes escaped. Is that intended?

yoshuawuyts: @isgj that's definitely not expected and we should fix that. Filed https://github.com/lrlna/femme/issues/12 on femme. Thanks for reporting!

yoshuawuyts: Okay, this will require patching the `log` crate. I don't think we should block adding this feature on that, even if it might be worth fixing in either femme or tide for the time being.

yoshuawuyts:
Filed an issue on `log` as well. Merging this so folks get access to logging, and once `log` is patched we'll receive a new version as well thanks to semver.

PR #498, opened 2020-05-13
impl `Response::redirect_see_other`

ethanboxx:
I just felt that this seemed to be missing.

`307` `Temporary Redirect` and `308`  `Permanent Redirect` result in undesired behaviour when you  want to simply display another page after processing a `post` request as `post` requests remain `post`s on the new URL.

`303` `See Other` Does not have this issue as it converts everything to `get`.

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -95,6 +95,30 @@ impl Response {
             .set_header("location".parse().unwrap(), location)
     }
 
+    /// Creates a response that represents a see other redirect to `location`.
+    ///
+    /// GET methods are unchanged.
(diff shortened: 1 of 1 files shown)


Fishrock123:
Is there an issue with just calling it `redirect()` if it is the best way of doing so for most use-cases?

yoshuawuyts:
@Fishrock123 no issue if that's the case; could you say more why you think this is the right method for most cases tho? I'm not super familiar with how these methods are used.

yoshuawuyts:
Closing in favor of #513.

PR #501, opened 2020-05-14
Add Server::respond

yoshuawuyts:
This adds `Server::respond`, replacing most of the functionality provided by the `http-service` trait. This allows passing arbitrary requests into the server, and always get responses back out. The goal is that we remove the `http-service` functionality in a future PR entirely, and replace it with this method instead.

What this enables us to do is to construct requests using `Surf`, and then pass them into `Tide` without a problem:

```rust
let mut app = tide::new();
app.at("/").get(|_| async move { Ok("hello world") });

let req = surf::get("https://business.soup");
let res: surf::Response = app.respond(req).await?;

assert_eq!(res.status(), 200);
```

## Screenshots

Screenshot_2020-05-15 tide Server - Rust

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -302,30 +302,30 @@ impl<State: Send + Sync + 'static> Server<State> {
     /// # #[async_std::main]
     /// # async fn main() -> http_types::Result<()> {
     /// #
-    /// use tide::http::{self, Url, Method};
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
Tests are failing locally. Closing in favor of #503

PR #502, opened 2020-05-15
Change file structure to match external exports

yoshuawuyts:
Submodules now have subdirectories. Top-level exports now live in the project root. In particular `serve_dir` and `route` no longer live under `server/`; I found this quite unintuitive and this would make that a bit easier to reason about. Thanks!


yoshuawuyts:
These are purely internal changes; going to merge!

PR #503, opened 2020-05-15
Remove http service

yoshuawuyts:
Depends on #501. Removes all of our use of `http-service`, simplifying our tests significantly. This sets us up to use Surf for our request instantiation in the future as well. Thanks!

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -335,8 +335,8 @@ impl<State: Send + Sync + 'static> Server<State> {
     {
         let req = Request::new(self.state.clone(), req.into(), Vec::new());
         match self.call(req).await {
-            Ok(value) => {
-                let res: http_types::Response = value.into();
+            Ok(res) => {
+                let res: http_types::Response = res.into();
                 // We assume that if an error was manually cast to a
                 // Response that we actually want to send the body to the
                 // client. At this point we don't scrub the message.
--- a/src/server.rs
+++ b/src/server.rs
@@ -302,30 +302,30 @@ impl<State: Send + Sync + 'static> Server<State> {
     /// # #[async_std::main]
     /// # async fn main() -> http_types::Result<()> {
     /// #
-    /// use tide::http::{self, Url, Method};
+    /// use tide::http::{Url, Method, Request, Response};
     ///
     /// // Initialize the application with state.
     /// let mut app = tide::new();
     /// app.at("/").get(|_| async move { Ok("hello world") });
     ///
-    /// let req = http::Request::new(Method::Get, Url::parse("https://example.com")?);
-    /// let res = app.respond(req).await?;
+    /// let req = Request::new(Method::Get, Url::parse("https://example.com")?);
+    /// let res: Response = app.respond(req).await?;
     /// assert_eq!(res.status(), 200);
     /// #
     /// # Ok(()) }
     /// ```
-    pub async fn respond(
-        &self,
-        req: impl Into<http_types::Request>,
-    ) -> http_types::Result<http_types::Response> {
+    pub async fn respond<R>(&self, req: impl Into<http_types::Request>) -> http_types::Result<R>
+    where
+        R: From<http_types::Response>,
+    {
         let req = Request::new(self.state.clone(), req.into(), Vec::new());
         match self.call(req).await {
             Ok(value) => {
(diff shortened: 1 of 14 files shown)


src/server.rs
```diff
@@ -281,14 +279,81 @@ impl<State: Send + Sync + 'static> Server<State> {
...
+            task::spawn(async move {
+                let res = async_h1::accept(&addr, stream, |req| async {
+                    let res = this.respond(req).await;
+                    let res = res.map_err(|_| io::Error::from(io::ErrorKind::Other))?;
+                    Ok(res)
+                })
+                .await;
+
+                if let Err(err) = res {
+                    log::error!("async-h1 error", { error: err.to_string() });
```
jbr, line 302: would this always be `io::Error::from(io::ErrorKind::Other).to_string()` because of the map_err on 296?

tests/test_utils.rs
```diff
@@ -1,7 +1,5 @@
-pub async fn find_port() -> async_std::net::SocketAddr {
-    async_std::net::TcpListener::bind("localhost:0")
-        .await
-        .unwrap()
-        .local_addr()
-        .unwrap()
+use portpicker::pick_unused_port;
```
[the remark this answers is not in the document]
yoshuawuyts: Yeah this was the fix for the windows failures. It seems we were getting duplicate ports using our other approach.

src/server.rs
jbr: On a second read it seems like only _unhandled_ errors from the server/endpoint/middleware stack are transformed into io::ErrorKind::Other? That's certainly less concerning but we probably still want to log the details. Does this need to return io::Result or can it be a broader http_types::Result, since it potentially is representing any variety of error?

src/server.rs
```diff
@@ -281,14 +279,81 @@ impl<State: Send + Sync + 'static> Server<State> {
...
     /// Asynchronously serve the app at the given address.
     #[cfg(feature = "h1-server")]
-    pub async fn listen(self, addr: impl ToSocketAddrs) -> std::io::Result<()> {
+    pub async fn listen(self, addr: impl ToSocketAddrs) -> io::Result<()> {
         let listener = async_std::net::TcpListener::bind(addr).await?;
 
         let addr = format!("http://{}", listener.local_addr()?);
         log::info!("Server is listening on: {}", addr);
-        let mut server = http_service_h1::Server::new(addr, listener.incoming(), self);
+        let mut incoming = listener.incoming();
```
connec, line 287: Sorry to butt into a PR, but this could be a convenient opportunity to consider a separate method that takes a `TcpListener` (or `impl Into<TcpListener>`), even if it's private for now.

I have an agenda as the author of #485 – a side-effect of this change would be making it more difficult to set up Tide with auto-reloading. Specifically, this change would make it infeasible to 'reimplement' `tide::Server::listen` with a pre-existing `TcpListener`.

Apologies if this is redundant with other planned work.

jbr: There's work in the direction of supporting any transports that's AsyncRead+AsyncWrite+Clone+Send+Sync+'static. Once these changes hit, regardless of which convenience functions exit on Server, it will likely be able to directly implement your own listen loop for any transport, along the lines of https://gist.github.com/yoshuawuyts/e02cc9626aea265a97c181e6cc3baebe

connec: Awesome, I did think #486 would supersede my specific concerns.

jbr: And in the near future, it'll look something like this https://gist.github.com/jbr/183aaeb835aa30c3a5f94ee91cdc473b — very similar but slightly simpler, only two args to async_h1::accept

src/server.rs
```diff
@@ -281,14 +279,81 @@ impl<State: Send + Sync + 'static> Server<State> {
...
+    /// #
+    /// # Ok(()) }
+    /// ```
+    pub async fn respond<R>(&self, req: impl Into<http_types::Request>) -> http_types::Result<R>
+    where
+        R: From<http_types::Response>,
+    {
+        let req = Request::new(self.state.clone(), req.into(), Vec::new());
+        match self.call(req).await {
+            Ok(value) => {
```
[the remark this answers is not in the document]
yoshuawuyts: fair point; done!

PR #508, opened 2020-05-16
Response: expose `remove_header`

Fishrock123:
From `http_types::Response`

Useful for e.g. compression middleware where content-length is normally stripped.

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -111,6 +111,11 @@ impl Response {
         self.res.len()
     }
 
+    /// Remove a header.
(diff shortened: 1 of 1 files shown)


Fishrock123:
It occurs to me that this is probably using the old `HeaderName` signature, but that doesn't appear to be updated for `header` and `remove_header` in `http_types::Response` compared to `inster_header`: https://docs.rs/http-types/1.2.0/http_types/struct.Response.html

yoshuawuyts:
@Fishrock123 it'll change again with http-types 2.0, which the next release of Tide will be blocked on. I'm happy to merge this as-is to enable this functionality (:

PR #513, opened 2020-05-17
Better redirects

ethanboxx:
This PR addresses issues with my previous PR #498 by making more extreme breaking changes.

# Motivation.
- The way that redirects work now does not feel not very extensible. Having redirect endpoints and redirect responses be different things is confusing. This abstracts them into one type.
- Provides a single way of redirecting with sensible defaults rather than two confusing ones.
*Note I decided to use the response `302` `Found` as the default. It is only subtly different to `303` `See Other` and is used as default in [NestJS](https://docs.nestjs.com/controllers#redirection), [ExpressJS](https://expressjs.com/en/4x/api.html#res.redirect), [Flask](https://flask.palletsprojects.com/en/1.1.x/api/#useful-functions-and-classes) and [Django](https://docs.djangoproject.com/en/3.0/ref/request-response/#django.http.HttpResponseRedirect)

# Questions
- Should I have left the old functions or marked them as depreciated?


src/redirect.rs
```diff
@@ -0,0 +1,107 @@
...
+        Self {
+            status: StatusCode::SeeOther,
+            location,
+        }
+    }
+
+    /// Returns response with equivalent redirect.
+    pub fn response(&self) -> Response {
+        Response::new(self.status).set_header("location".parse().unwrap(), &self.location)
+    }
```
yoshuawuyts, line 90: This probably doesn't need to be public just yet. I like the `Into` impl for this purpose.

ethanboxx: I was able to remove it using an extra `Into` `impl`. See my latest commit.

src/redirect.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+impl<T: AsRef<str>> Redirect<T> {
+    /// Creates an endpoint that represents a redirect to `location`.
+    ///
+    /// Uses status code 302 Found.
+    pub fn new(location: T) -> Self {
+        Self {
+            status: StatusCode::SeeOther,
+            location,
```
Fishrock123, line 53: This can probably be improved by having `location` `.parse()` into a `HeaderName` already?

ethanboxx: @Fishrock123 
Not sure what you're getting at. The location argument on this method does not become a `HeaderName`.

Can you clarify?

Fishrock123: Whoops, my bad. I think I was probably thinking `HeaderValue` but `set_header` does that under the hood anyway.

yoshuawuyts:
@ethanboxx if you can rebase this on master we can merge this!

ethanboxx:
Done

PR #523, opened 2020-05-21
adds initial support for json literals, adds "json" feature

jbr:


src/response.rs
```diff
@@ -263,6 +263,14 @@ impl Into<http::Response> for Response {
     }
 }
 
+impl From<serde_json::Value> for Response {
+    fn from(json_value: serde_json::Value) -> Self {
+        Response::new(StatusCode::Ok)
+            .body_json(&json_value)
+            .unwrap_or_else(|_| Response::new(StatusCode::InternalServerError))
```
jbr, line 270: I'm not actually sure if this is reachable, since a serde_json::Value should basically be the safest thing to turn into json, since it's the intermediate representation before stringification

src/response.rs
```diff
@@ -263,6 +263,14 @@ impl Into<http::Response> for Response {
     }
 }
 
+impl From<serde_json::Value> for Response {
+    fn from(json_value: serde_json::Value) -> Self {
+        Response::new(StatusCode::Ok)
```
Fishrock123, line 268: One thing about this -- it'd be nice to have non-200 status codes supported for easy json responses.

Is it also possible to do something like `impl From<(StatusCode, serde_json::Value)>`?

jbr: I mean it's _possible_ to provide
```
impl<S: TryInto<StatusCode>, R: Into<Response>> From<(S, R)> for Response {
    fn from(tuple: (S, R)) -> Self {
        tuple.1.into().set_status(
            tuple
                .0
                .try_into()
                .unwrap_or_else(|_| panic!("Invalid status code")),
        )
    }
}
```

But I don't have a great sense of whether it's good design.  That would let you do like `Ok(200, json!({"status": "ok"}))` or `Ok((StatusCode::NotFound, "not found"))`, etc

jbr: Another option to make constructing responses even easier would be to provide a `respond!` macro that can take variadic args and is a shortcut for building an `Ok(Response::new(…))`

yoshuawuyts: The way I was thinking that should be done instead is by replying on the `Response:: new` constructor:

```async fn respond(_: Request) -> Result<Response> {
    let mut res = Response::new(203);
    res.set_body(json!({"color": "cyan"});
    Ok(res)
}
```
I don't feel converting from a tuple or introducing a new macro would improve much on this.

src/response.rs
```diff
@@ -263,6 +263,14 @@ impl Into<http::Response> for Response {
     }
 }
 
+impl From<serde_json::Value> for Response {
```
yoshuawuyts, line 266: We should consider adding this impl to `Body` as well!

jbr: What if instead of two different From<serde_json::Value>'s:
`impl From<serde_json::Value> for Body`
and
a `impl<T: Into<Body>> From<T> for Response` with `Response::new(StatusCode::Ok).set_body(self)`

Fishrock123: Does that still allow `Ok(json!({}))`? It certainly sounds more flexible.

jbr: @Fishrock123 It would, _if_ I could figure out how to make it work with coherence rules. http_types::Response [has something like this](https://github.com/http-rs/http-types/blob/master/src/response.rs#L609-L619)

yoshuawuyts: @jbr I think it's possible to do that, but believe there was a reason why we shouldn't do that in Tide. I forget what that was though. I think if we can add the Body impl manually we'll still be good, and we can perhaps explore generalizing things in a future PR.

src/response.rs
yoshuawuyts: we should probably use the `Status` trait here instead so we don't lose the original error message.

src/prelude.rs
```diff
@@ -1,2 +1,3 @@
 //! The Tide prelude.
 pub use http_types::Status;
+pub use serde_json::json;
```
dignifiedquire, line 3: would be nice to make `serde_json` a feature dependency, even if enabled by default

jbr: I expected to need to do that when starting this pr, but discovered it was already a normal dependency. We'd probably want to gate json_body on this as well

dignifiedquire: Yep, that would be a good improvement, given that serde_json is not the lightest dependency

jbr: Added a "json" feature, enabled by default, and gated everything json-related in tide on it. Probably could stand to do the same for http-types as well

yoshuawuyts:
@jbr it seems tests are still failing; I'd love to get these to pass so we can merge this PR.

src/response.rs
jbr: As far as returning status: From<serde_json::Value for Response needs to return a Response, not a tide::Error. I can imagine the value of adding an Option<Error> on Response (or using ext in a systematic way) and adding a TryInto or downcast, but that's way beyond the scope of this PR. If people want to retain the original error message from transforming a json literal into a Response, they'll need to explicitly do the `body_json` call for now

jbr:
@yoshuawuyts Fixed!

jbr:
Wait actually: I was just digging in on adding a similar json feature for http-types, but ran into this issue: [https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-to](This header) takes a json value, which means http-types csp stuff needs serde_json, which means tide always is going to have a serde_json dependency, in which case is there any reason to add the feature flag?

yoshuawuyts:
Heh, yeah probably always supporting JSON is fine. Serde even works on embedded systems, so don't think there's really any practical reason to make this opt-out.

jbr:
I'll back out the feature-gate parts of this commit

PR #526, opened 2020-05-21
Various Clippy Fixes

ethanboxx:
This satisfies all off `clippy::all` with some fixes from `clippy::pedantic` and `clippy::nursery`.

Note: It's probably easier to review this per commit rather than as the whole diff as it has lots of little changes.

### DIFF
--- a/src/fs/serve_dir.rs
+++ b/src/fs/serve_dir.rs
@@ -57,12 +57,11 @@ impl<State> Endpoint<State> for ServeDir {
                 Ok(file) => file,
             };
 
-            let len = match file.metadata().await {
-                Ok(metadata) => metadata.len() as usize,
-                Err(_) => {
-                    log::warn!("Could not retrieve metadata");
-                    return Ok(Response::new(StatusCode::InternalServerError));
-                }
+            let len = if let Ok(metadata) = file.metadata().await {
+                metadata.len() as usize
+            } else {
+                log::warn!("Could not retrieve metadata");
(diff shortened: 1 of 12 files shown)


src/fs/serve_dir.rs
```diff
@@ -57,12 +57,11 @@ impl<State> Endpoint<State> for ServeDir {
...
-                Ok(metadata) => metadata.len() as usize,
-                Err(_) => {
-                    log::warn!("Could not retrieve metadata");
-                    return Ok(Response::new(StatusCode::InternalServerError));
-                }
+            let len = if let Ok(metadata) = file.metadata().await {
+                metadata.len() as usize
+            } else {
+                log::warn!("Could not retrieve metadata");
+                return Ok(Response::new(StatusCode::InternalServerError));
```
yoshuawuyts, line 64: This is something I disagree with Clippy on.

src/security/cors.rs
```diff
@@ -151,12 +153,11 @@ impl<State: Send + Sync + 'static> Middleware<State> for CorsMiddleware {
...
-                Some(origin) => origin,
-                None => {
-                    // This is not a CORS request if there is no Origin header
-                    return next.run(req).await;
-                }
+            let origin = if let Some(origin) = origins.first() {
+                origin
+            } else {
+                // This is not a CORS request if there is no Origin header
+                return next.run(req).await;
```
yoshuawuyts, line 160: Clippy has an odd aversion to match statements that I don't share.

PR #529, opened 2020-05-22
Impl Index for Request,Response

yoshuawuyts:
Thanks!


yoshuawuyts:
@Fishrock123 this is for Tide's Req/Res types only. Unfortunately Surf's types are different, though similar. We can't update the lines you pointed out here.

Fishrock123:
shoot it's the second time I made that mistake >_<

PR #531, opened 2020-05-23
Unix socket listener, take 2

jbr:


yoshuawuyts:
Now that it's becoming clear that this will also work on Windows [ref](https://internals.rust-lang.org/t/unixstream-on-windows/12305/4?u=yoshuawuyts) I'm in favor of adding this as a method. That way "HTTP over RPC" can be "unix streams" on *every* platform. The PR looks good overall; let's make this part of the next release.

PR #535, opened 2020-05-23
Add body_html_string method to Response

gmosx:
This PR adds the `body_html_string` helper method proposed in the discussion of this issue:

https://github.com/http-rs/tide/issues/534

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -140,6 +140,17 @@ impl Response {
         self.set_mime(mime::TEXT_PLAIN_UTF_8)
     }
 
+    /// Pass an HTML string as the request body.
+    ///
+    /// # Mime
+    ///
+    /// The encoding is set to `text/html; charset=utf-8`.
+    #[must_use]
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
@gmosx thanks for opening this PR! We've recently removed all `body_*` methods in favor of `set_body`. We understand there's a desire to make it easier to return HTML from endpoints, but this is probably not it.

I'm wondering if we could perhaps expand [`serve_dir`](https://docs.rs/tide/0.9.0/tide/struct.Route.html#method.serve_dir) to serve single files, which would make the case of serving a single file easier.

gmosx:
Please note that my use-case is to serve  *dynamically-generated* html content, so `server_dir` is not appropriate.

mendelt:
I think this does the same as https://github.com/http-rs/tide/pull/292. I think it can be closed. Maybe a new pr can be opened for a `from_html` method on the `Body` type in the `http-types`. That is where `from_json` and `from_str` have moved.

PR #536, opened 2020-05-24
use http_types::mime instead of the mime crate

jbr:
Fix the use of the mime crate in the graphql example (brought up by #534), which was a mistake. This pr removes dependence on the mime crate entirely, in preference to http_types::Mime

This pr also allows set_mime to accept an Into<Mime>, and calls through to the http_types::Response::set_content_type

### DIFF
--- a/src/fs/serve_dir.rs
+++ b/src/fs/serve_dir.rs
@@ -1,6 +1,6 @@
+use crate::http::Mime;
 use crate::log;
 use crate::{Body, Endpoint, Request, Response, Result, StatusCode};
-use crate::http::Mime;
 
 use async_std::fs::File;
 use async_std::io::BufReader;
--- a/src/fs/serve_dir.rs
+++ b/src/fs/serve_dir.rs
@@ -1,5 +1,6 @@
 use crate::log;
 use crate::{Body, Endpoint, Request, Response, Result, StatusCode};
+use crate::http::Mime;
 
 use async_std::fs::File;
 use async_std::io::BufReader;
@@ -70,7 +71,7 @@ impl<State> Endpoint<State> for ServeDir {
             res.set_body(body);
 
             if let Some(content_type) = mime_guess::from_path(&file_path).first() {
-                res = res.set_mime(content_type.to_string().parse().unwrap());
+                res = res.set_mime(content_type.to_string().parse::<Mime>().unwrap());
             }
 
             Ok(res)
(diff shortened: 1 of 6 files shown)


src/fs/serve_dir.rs
```diff
@@ -70,7 +71,7 @@ impl<State> Endpoint<State> for ServeDir {
             res.set_body(body);
 
             if let Some(content_type) = mime_guess::from_path(&file_path).first() {
-                res = res.set_mime(content_type);
+                res = res.set_mime(content_type.to_string().parse::<Mime>().unwrap());
```
Fishrock123, line 74: This is not very nice, should `http_types::HeaderValue` have an into for this?

jbr: I think this code is temporary and will be replaced by the mime sniffing stuff in http_types

jbr: This is converting from a mime::Mime to a http_types::Mime, and I think the goal is for neither tide nor http_types to have a dep on the mime crate

yoshuawuyts: http-types Body now has a `from_file` constructor that applies the inference. Could as maybe use that here?

The trickiest part I see is checking the error type so that we can set the right status codes, but that should be possible through downcasting.

That way we can get rid of the `mime` dep effective immediately!

jbr: Addressed in 5000e1c

src/response.rs
```diff
@@ -125,8 +125,9 @@ impl Response {
     ///
     /// [Read more on MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types)
     #[must_use]
-    pub fn set_mime(self, mime: Mime) -> Self {
-        self.set_header(CONTENT_TYPE, mime.to_string())
+    pub fn set_mime(mut self, mime: impl Into<Mime>) -> Self {
+        self.res.set_content_type(mime.into());
+        self
```
Fishrock123, line 130: This line should probably be reverted now?

jbr: Why?

Fishrock123: I mean due to #537 and [the discord chaining conversation yesterday](https://discordapp.com/channels/598880689856970762/649056551835009097/714252404111179856).

jbr: This PR doesn’t change the signature of set_mime, and doing so would be out of its scope. Although Request has changed, Response has not yet as of current master

PR #539, opened 2020-05-24
Update docs from Request.local to Request.ext

bluk:
The method was renamed in Tide 0.9.

Just 2 cents, it was a bit confusing to me to use "local state" and "global state" as the names because IMHO those terms are normally associated with entire programs. Other frameworks call "local state" as "per request state" or "request scoped state". The "global state" is sometimes referred to as "application scoped state" as well. I'm not familiar with the design and terminology of this framework, but my suggestion would be to clarify/rename those concepts.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -104,10 +104,14 @@
 //!
 //! ## State
 //!
-//! Middleware often needs to share values with the endpoint. This is done through "local state".
-//! Local state is built using a typemap that's available through [`Request::local`].
+//! Middleware often needs to share values with the endpoint. This is done through "request scoped
+//! state".  Request scoped state is built using a typemap that's available through
+//! [`Request::ext`].
 //!
-//! Global state is used when a complete application needs access to a particular
+//! If the endpoint needs to share values with middleware, response scoped state can be set via
+//! [`Response::set_ext`] and is available through [`Response::ext`].
+//!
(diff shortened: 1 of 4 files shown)


bluk:
I added a commit to change the "local"/"global" state doc references since the first commit is trivial. Feel free to modify/discuss if needed.

src/lib.rs
```diff
@@ -104,10 +104,11 @@
 //!
 //! ## State
 //!
-//! Middleware often needs to share values with the endpoint. This is done through "local state".
-//! Local state is built using a typemap that's available through [`Request::local`].
+//! Middleware often needs to share values with the endpoint. This is done through "request scoped
+//! state".  Request scoped state is built using a typemap that's available through
+//! [`Request::ext`].
```
Fishrock123, line 109: Should also mention `Response::ext`

bluk: I added a simple paragraph for Response extensions. Thanks.

yoshuawuyts:
@bluk I think CI is failing because of #531; I'm trying a fix in https://github.com/http-rs/tide/pull/552. Could you rebase on master once we merge? Thanks heaps!

bluk:
Rebased and squashed to one commit. If there's anything else, please feel free to leave a comment.

PR #540, opened 2020-05-24
Async jobs

ivan770:
Currently, you can access shared state only in request scope.

This causes a lot of troubles, when you, for example, need to save state to file based on timer, or do any actions with state without HTTP request.

This PR introduces `Job` and `JobContext`, that represent async job and job context with app state respectively.

API is pretty straightforward:
```rust
#[async_std::main]
    let mut app = tide::new();

    app.at("/").get(|_| async move {
        Ok("I am executed on HTTP request")
    });

    app.spawn(|_| async move {
        println!("I am executed when you call 'listen'!");
    });

    Ok(())
}
```

[Example with app state](https://github.com/ivan770/tide/blob/cc568e153ecdb5d02c46e82b1d7388aa961bb154/tests/jobs.rs)


Fishrock123:
I think this was solved by https://github.com/http-rs/tide/pull/675 - you can now access state again off of the server, and so you could make an external task which does this.

Please let me know if I missed something though!

PR #541, opened 2020-05-24
Add support for signed and private/encrypted cookies

kyrias:
This PR enables using the `cookie` crate's signed and encrypted cookies in Tide applications.

The `cookie` crate uses the AEAD construct from `ring` to provide authenticated encrypted (what the `cookie` crate calls "private") cookies.

Requires https://github.com/http-rs/http-types/pull/152.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ async-h1 = { version = "2.0.1", optional = true }
 async-sse = "3.0.0"
 async-std = { version = "1.6.0", features = ["unstable"] }
 femme = "2.0.1"
-http-types = "2.2.1"
+http-types = { version = "2.2.1", default-features = false }
 kv-log-macro = "1.0.4"
 serde = "1.0.102"
 serde_json = "1.0.41"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,9 +24,10 @@ features = ["docs"]
 rustdoc-args = ["--cfg", "feature=\"docs\""]
 
 [features]
-default = ["h1-server"]
+default = ["h1-server", "secure-cookies"]
 h1-server = ["async-h1"]
 docs = ["unstable"]
+secure-cookies = ["http-types/cookie-secure"]
 unstable = []
 # DO NOT USE. Only exists to expose internals so they can be benchmarked.
 __internal__bench = []
--- a/src/request.rs
+++ b/src/request.rs
@@ -8,6 +8,8 @@ use std::{str::FromStr, sync::Arc};
 
 use crate::cookies::CookieData;
 use crate::http::cookies::Cookie;
+#[cfg(feature = "secure-cookies")]
+use crate::http::cookies::Key;
 use crate::http::headers::{self, HeaderName, HeaderValues, ToHeaderValues};
 use crate::http::{self, Body, Method, Mime, StatusCode, Url, Version};
 use crate::Response;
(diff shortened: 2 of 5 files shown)


kyrias:
(Turns out that cookie 0.14.0 will use the RustCrypto crates instead of ring, though the CookieJar interface appears to be unchanged so this PR will thus not be impacted by that.)

.github/workflows/ci.yaml
```diff
@@ -65,6 +65,12 @@ jobs:
         command: test
         args: --all --features unstable
 
+    - name: tests secure-cookies
+      uses: actions-rs/cargo@v1
+      with:
+        command: test
+        args: --all --features secure-cookies
```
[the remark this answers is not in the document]
kyrias: No real point, other than wanting to make it optional.  I'll add it to the default feature.

src/request.rs
```diff
@@ -369,13 +371,32 @@ impl<State> Request<State> {
...
+    /// Returns a decrypted `Cookie` by name of the cookie if it authenticated and decrypted
+    /// successfully.
+    #[cfg(any(feature = "secure-cookies", feature = "docs"))]
+    #[cfg_attr(feature = "docs", doc(cfg(feature = "secure-cookies")))]
+    #[must_use]
+    pub fn private_cookie(&self, key: &Key, name: &str) -> Option<Cookie<'static>> {
+        self.ext::<CookieData>()
+            .and_then(|cookie_data| cookie_data.content.write().unwrap().get_private(key, name))
+    }
+
```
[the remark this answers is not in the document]
kyrias: Yeah, the cookie crate has a really nice-to-use API for these.  :)

kyrias:
Added `secure-cookies` to the default feature now.  Once there's been a new `http-types` release I'll drop the first commit and it'll be ready.

jbr:
@kyrias Tide is now on http-types 2.2.1!

kyrias:
@jbr @yoshuawuyts Ready for review now.  :)

Cargo.toml
```diff
@@ -36,7 +37,7 @@ async-h1 = { version = "2.0.1", optional = true }
 async-sse = "3.0.0"
 async-std = { version = "1.6.0", features = ["unstable"] }
 femme = "2.0.1"
-http-types = "2.2.1"
+http-types = { version = "2.2.1", default-features = false }
```
Fishrock123, line 40: What is the purpose of this?

Cargo.toml
```diff
@@ -24,9 +24,10 @@ features = ["docs"]
 rustdoc-args = ["--cfg", "feature=\"docs\""]
 
 [features]
-default = ["h1-server"]
+default = ["h1-server", "secure-cookies"]
 h1-server = ["async-h1"]
 docs = ["unstable"]
+secure-cookies = ["http-types/cookie-secure"]
```
Fishrock123, line 30: This is enabled by default.

src/request.rs
```diff
@@ -386,13 +388,32 @@ impl<State> Request<State> {
...
     pub fn cookie(&self, name: &str) -> Option<Cookie<'static>> {
         self.ext::<CookieData>()
             .and_then(|cookie_data| cookie_data.content.read().unwrap().get(name).cloned())
     }
 
+    /// Returns a signed `Cookie` by name of the cookie if it verified successfully.
+    #[cfg(feature = "secure-cookies")]
+    #[cfg_attr(feature = "docs", doc(cfg(feature = "secure-cookies")))]
+    #[must_use]
+    pub fn signed_cookie(&self, key: &Key, name: &str) -> Option<Cookie<'static>> {
```
Fishrock123, line 402: This seems awkward, passing the key everywhere? I don't think that's the right way to do this, more details in review comment.

kyrias:
I would consider having `req.cookie()` behave differently based on how the
`Cookies` middleware was created to be quite surprising, and something I would
not expect or want outside of big magic frameworks like Django or Rails.  It
would also mean that each time you call `req.cookie()` in a project, you have
to ensure that the `Cookies` middleware was created with the appropriate level
of security required for the new cookie, and anyone changing how it was created
would need to inspect each `req.cookie()` callsite to ensure that it's still
safe.

I would consider this to be a much more brittle interface, and believe it would
inherently affect the security of all tide-based applications.

This interface would additionally make it much harder to migrate from signed to
encrypted cookies, since all cookies would end up being ignored after that,
whereas when it's a per-cookie decision it's easy to make a decision to convert
or reset the cookie as necessary for that cookie.

An interface more or less identical to the one that I've already implemented
would be required for a lot of usecases either way though, since the per-cookie
decision would be required in quite a few cases.  You would e.g. have to
explicitly allow getting a plaintext cookie if it's supposed to be set/get by
either JavaScript or by the API client, and you would have to be able to get a
cookie with a specific different encryption key for SSO usecases where you
would not want the same cookie encryption key to be used for both.

Cargo.toml
kyrias: So that applications that do not need signed or encrypted cookies don't have to needlessly pull in a bunch of crypto crates.

Cargo.toml
kyrias: See above.

Fishrock123:
> Perhaps there can still be a get/set raw/signed/encrypted if people need it

I'm not suggesting there can't be individual methods, but I do think having `res/req .cookie()` deal with this automatically (with good docs) is probably best for most people, and worth not having to pass keys around everywhere.

> and believe it would
inherently affect the security of all tide-based applications.

I'm not really seeing how - if anything this would make it easier. I'd make it so that `Cookies::new()` points to `Cookies::signed(secret)` instead of `Cookies::unsigned()`.

kyrias:
> I'm not really seeing how - if anything this would make security easier. I'd make it so that Cookies::new() points to Cookies::signed(secret) instead of Cookies::unsigned().

Even if `Cookies::new` would end up creating a signed version, any other method whose security changes based on how other parts of the application was configured will eventually cause someone to miss the fact that one application happened to be configured differently, leading to either a privacy leak or a security vulnerability.

As I do not feel that this is a good way to design this API, I would not feel comfortable implementing it, and will keep my version internal to my application instead.

PR #545, opened 2020-05-25
allow for function middlewares by dropping Debug bound and add example

jbr:


src/middleware.rs
```diff
@@ -17,9 +17,13 @@ pub trait Middleware<State>: 'static + Send + Sync {
     /// Asynchronously handle the request, and return a response.
     fn handle<'a>(
         &'a self,
-        cx: Request<State>,
+        request: Request<State>,
         next: Next<'a, State>,
     ) -> BoxFuture<'a, crate::Result>;
+
+    fn name(&self) -> &str {
```
yoshuawuyts, line 24: I like this a lot!

We should probably document this function though.

Issue #525, opened 2020-05-21
CorsMiddleware doesn't apply headers to error responses


davidpdrsn:
It seems that because of [this `?`](https://github.com/http-rs/tide/blob/master/src/security/cors.rs#L171) `CorsMiddleware` doesn't apply headers to errors. So when the error gets turned into a response the client will still get errors about a missing CORS header.

I have looked a bit at the implementation but don't see a good solution currently, because you're not able to add headers to errors. Since that would require a patch to [http_types](https://github.com/http-rs/http-types) I thought I would open an issue here first. Maybe there is an easier solution.

rkusa:
Here is a workaround until this gets fixed. Just add this as your last middleware:

```rust
#[derive(Debug)]
struct CorsFix;
impl<State: Send + Sync + 'static> Middleware<State> for CorsFix {
    fn handle<'a>(
        &'a self,
        req: Request<State>,
        next: Next<'a, State>,
    ) -> BoxFuture<'a, Result> {
        Box::pin(async move {
            // Convert any error to a response so CORS headers can be added
            Ok(match next.run(req).await {
                Ok(res) => res,
                Err(err) => {
                    let mut res = http_types::Response::new(err.status());
                    if !res.status().is_server_error() {
                        res.set_body(err.to_string());
                    }
                    res.into()
                }
            })
        })
    }
}
app.middleware(CorsFix);
```

rkusa:
This should btw be fixed with recent changes to tide - see also the green unit test here https://github.com/http-rs/tide/pull/546.

PR #546, opened 2020-05-26
Add CORS headers to error responses

rkusa:
Browsers are currently unable to read responses the originate from `tide::Error`s when the `CORS` middleware is used. This PR makes sure to add CORS headers to error responses as well.

I guess we want to implement parts of the PR differently (I'll highlight the line I am talking about via a code comment), but I still wanted to open a PR to start the discussion and maybe at least provide a unit test another implementation could reuse.

Refs #525

### DIFF
--- a/src/security/cors.rs
+++ b/src/security/cors.rs
@@ -379,4 +379,20 @@ mod test {
 
         assert_eq!(res[http_types::headers::SET_COOKIE][0], "foo=bar");
     }
+
+    #[async_std::test]
+    async fn set_cors_headers_to_error_responses() {
+        let mut app = crate::Server::new();
+        app.at(ENDPOINT).get(|_| async {
+            Err::<&str, _>(crate::Error::from_str(
+                StatusCode::BadRequest,
+                "bad request",
+            ))
+        });
+        app.middleware(CorsMiddleware::new().allow_origin(Origin::from(ALLOW_ORIGIN)));
+
+        let res: crate::http::Response = app.respond(request()).await.unwrap();
+        assert_eq!(res.status(), 400);
+        assert_eq!(res[headers::ACCESS_CONTROL_ALLOW_ORIGIN], ALLOW_ORIGIN);
+    }
 }


src/utils.rs
```diff
@@ -1,6 +1,23 @@
...
+    let mut res = http_types::Response::new(err.status());
+    res.set_content_type(http_types::mime::PLAIN);
+    // Only send the message if it is a non-500 range error. All
+    // errors default to 500 by default, so sending the error
+    // body is opt-in at the call site.
+    if !res.status().is_server_error() {
+        res.set_body(err.to_string());
+    }
+    res
+}
```
rkusa, line 23: This is the implementation part I am not sure about. I have the feeling that having this as a utility method is not the best solution, so I am definitely open to other ideas of how to streamline the error responses here?

davidpdrsn:
Not sure if it’s ideal that this middleware would “swallow” all errors. What if you have another middleware that logs errors or otherwise does stuff to errors?

That’s why I suggested instead being able to add headers directly to the error itself.

rkusa:
> Not sure if it’s ideal that this middleware would “swallow” all errors. What if you have another middleware that logs errors or otherwise does stuff to errors?
> That’s why I suggested instead being able to add headers directly to the error itself.

Thanks for bringing up that valid concern 👍! IMHO, you'd just have to add the middlewares in the order they should do stuff to errors. I think, headers shouldn't be the concern of a `tide::Error` - I rather see it within the responsibility of a middleware to transform an error into a response for any use-case where the error response should be customized (adding headers, adding another body, adding a JSON body, customize which status code includes the message as a body ...).

davidpdrsn:
Hm alright. Maybe it would be nice to also update the docs to make it clear that it’s important to add this middleware last, otherwise your other middlewares won’t see errors.

src/utils.rs
```diff
@@ -1,6 +1,23 @@
...
+/// retrieved from the error and the response body is set to the error message
+/// for non-500 errors.
+pub(crate) fn create_error_response(err: Error) -> http_types::Response {
+    let mut res = http_types::Response::new(err.status());
+    res.set_content_type(http_types::mime::PLAIN);
+    // Only send the message if it is a non-500 range error. All
+    // errors default to 500 by default, so sending the error
+    // body is opt-in at the call site.
+    if !res.status().is_server_error() {
+        res.set_body(err.to_string());
```
jbr, line 20: I don't think all apps would want the same error response here. I'm not sure what the right solution is, but perhaps we need a way of adding an Option<tide::Error> to a Response or something like that.

rkusa: I agree, however, I've only extracted the current error handling into that utility method, to make it re-usable. So this PR is not changing the current error handling. As far as I remember from Discord, the current approach for custom error handling would be a custom middleware that intercepts errors and converts them into a response (basically similar to this one https://github.com/http-rs/tide/issues/525#issuecomment-634144680).

Fishrock123:
Btw I opened https://github.com/http-rs/http-types/pull/169 which should hopefully make this kind of code nicer and more robust.

rkusa:
It looks like this got fixed with recent changes like https://github.com/http-rs/http-types/pull/174
I've reduced the PR to the unit-test, in case we want to add the test to prevent future regressions?

Issue #452, opened 2020-04-22
Ergonomics issue with enabling `?` in endpoint


Licenser:
The way `?` is now supported in Endpoint introduces some issues for more complex applications that do not use anonymous functions or need some control over how errors are transmitted.

To prefix this, I don't think there is a clear right or wrong in this. Allowing `?` in endpoints is a definite benefit for some use cases as the examples shared on twitter and other channels. However, I have the feeling that the impact on more complex API's was considered.

I want to start with a before -> after going from 0.6 to 0.7.

This is w/ tide 0.6 (and a custom fix to implement IntoResponse for Result). It is rather elegant to write:
```rust
    app.at("/version").get(api::version::get);
    app.at("/binding")
        .get(api::binding::list_artefact)
        .post(api::binding::publish_artefact);
    app.at("/binding/{aid}")
        .get(api::binding::get_artefact)
        .delete(api::binding::unpublish_artefact);
    app.at("/binding/{aid}/{sid}")
        .get(api::binding::get_servant)
        .post(api::binding::link_servant)
        .delete(api::binding::unlink_servant);
    app.at("/pipeline")
        .get(api::pipeline::list_artefact)
        .post(api::pipeline::publish_artefact);
    app.at("/pipeline/{aid}")
        .get(api::pipeline::get_artefact)
        .delete(api::pipeline::unpublish_artefact);
    app.at("/onramp")
        .get(api::onramp::list_artefact)
        .post(api::onramp::publish_artefact);
    app.at("/onramp/{aid}")
        .get(api::onramp::get_artefact)
        .delete(api::onramp::unpublish_artefact);
    app.at("/offramp")
        .get(api::offramp::list_artefact)
        .post(api::offramp::publish_artefact);
    app.at("/offramp/{aid}")
        .get(api::offramp::get_artefact)
        .delete(api::offramp::unpublish_artefact);
```

After migrating to 0.7 (with the same custom extension to IntoResponse) it looks like this. Which is rather painful to both read and write it introduces a whole lot of boilerplate that wasn't required before and adds quite a bit of redundancy.

```rust
    app.at("/version")
        .get(|r| async { Ok(api::version::get(r).await.into_response()) });
    app.at("/binding")
        .get(|r| async { Ok(api::binding::list_artefact(r).await.into_response()) })
        .post(|r| async { Ok(api::binding::publish_artefact(r).await.into_response()) });
    app.at("/binding/{aid}")
[...]

PR #547, opened 2020-05-26
Response, Result: add From-Into for each

Fishrock123:
Related to https://github.com/http-rs/tide/issues/452

As detailed in https://github.com/http-rs/tide/issues/452#issuecomment-633744876, I'm not sure these are actually presently that useful, but if someone has a good case for them, they are here...

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -276,3 +276,9 @@ where
 
 /// A specialized Result type for Tide.
 pub type Result<T = Response> = std::result::Result<T, Error>;
+
+impl From<Response> for Result<Response> {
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
I'm wary of this since it removes Ok-wrapping in some cases, but very strongly deviates from stdlib conventions. Tide used to have similar bounds, and one of the reasons we moved away from it was how confusing diagnostics could get.

I feel like the feature we really want here is language-level support for Ok-wrapping.

Fishrock123:
Yeah that's why I made it a draft. It seems most of the ways to make the return nicer depend on the `Try` trait.

PR #549, opened 2020-05-26
experimental error handler middleware for discussion

jbr:
Related to some of the recent discussions around error handling, I tried to sketch out an approach for error handling middleware that handles downcasting to a specific type of error. If the result is an Ok variant or an Err variant that does not downcast to the specified type, the middleware stack continues as normal.

This is a fully-functional proof of concept for discussion and experimentation, but not a finished product.

### outstanding concerns
* **request cloning**: if a middleware later on in the stack modifies the request, the error handler middleware only has a snapshot clone of the inbound request. This was the least-invasive change to the rest of tide to get this api functioning, but probably needs further thought
* **not async yet** this PR makes no attempt at getting this to work with async error handlers, but that seems like an important further revision.


examples/error_handler_middleware.rs
```diff
@@ -0,0 +1,83 @@
+use tide::{Request, Response, Result, StatusCode};
+#[async_std::main]
+async fn main() -> tide::Result<()> {
+    tide::log::with_level(tide::log::LevelFilter::Trace);
+    let mut app = tide::new();
+
+    app.middleware(ErrorHandler::new(
```
Fishrock123, line 7: Is there a good way to make this take any arbitrary number of these, i.e. a slice of closures or something?

jbr: I think it's probably clearer to just have a bunch of app.middleware()'s, especially since that allows them to be interleaved with other middlewares

jbr:
@yoshuawuyts I've been holding off on docs for these sorts of experimental proposals / code sketches because I've been expecting more large-change iterations than have been required! Expect more consistent documentation in the future!

src/error/error_handler.rs
```diff
@@ -0,0 +1,62 @@
...
+impl<E> ErrorHandler<E>
+where
+    E: Display + Debug + Send + Sync + 'static,
+{
+    pub fn new(f: impl ErrorMapper<E>) -> Self {
+        Self(Box::new(f))
+    }
+}
+
+impl<State: Clone + Send + Sync + 'static, E> Middleware<State> for ErrorHandler<E>
```
langbamit, line 41: I 'm not sure we need `Clone` state.

src/middleware.rs
```diff
@@ -22,8 +22,8 @@ pub trait Middleware<State>: 'static + Send + Sync {
     ) -> BoxFuture<'a, crate::Result>;
 
     /// Set the middleware's name. By default it uses the type signature.
-    fn name(&self) -> &str {
-        std::any::type_name::<Self>()
+    fn name(&self) -> String {
```
jbr, line 25: Calling attention to this signature change, which allows middlewares to return dynamic strings in their names, as used in https://github.com/http-rs/tide/pull/549/files#diff-c606b36cae12145a41d5fb07309603bfR108

src/error/error_handler.rs
jbr: Addressed in rebase, thanks!

PR #550, opened 2020-05-26
Request, Response: more method parity

Fishrock123:
Brings the methods on both `Request` and `Response` more to parity with each other and to `http_types`.

(Also includes some minor docs clarifications.)

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -9,7 +9,7 @@ use std::{str::FromStr, sync::Arc};
 use crate::cookies::CookieData;
 use crate::http::cookies::Cookie;
 use crate::http::headers::{self, HeaderName, HeaderValues, ToHeaderValues};
-use crate::http::{self, Method, StatusCode, Url, Version};
+use crate::http::{self, Body, Method, Mime, StatusCode, Url, Version};
 use crate::Response;
 
 /// An HTTP request.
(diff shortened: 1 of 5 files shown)


src/request.rs
```diff
@@ -149,6 +149,16 @@ impl<State> Request<State> {
         self.req.host()
     }
 
+    /// Get the request MIME.
+    ///
+    /// This gets the request `Content-Type` header.
+    ///
+    /// [Read more on MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types)
+    #[must_use]
+    pub fn get_mime(&self) -> Option<Mime> {
```
yoshuawuyts, line 158: I don't believe this has a counterpart in http-types?

Fishrock123: @yoshuawuyts Right - but `tide::Response::set_mime()` exists, and `http_types` has the `content_type` methods on each.

Personally, I don't like "mime", and think "content type" is clearer, but a lot of people know it as mime, some specs and rfcs, etc...

yoshuawuyts: We had a long-ish conversation about this on `http-types` and chose to go with `content_type`. I don't really love it either, but think it's the right choice since "mime" is ambiguous (do we set the encoding, or do we set the accept?)

src/request.rs
```diff
@@ -283,6 +293,16 @@ impl<State> Request<State> {
         })
     }
 
+    /// Take the request body as a Body.
```
ethanboxx, line 296: I don't know if it is conventional over all of tides docs, but I feel like `Body` this should be in backticks (\`) as it references a rust type.

Fishrock123:
Depends on https://github.com/http-rs/tide/pull/536

yoshuawuyts:
@Fishrock123 merged #536!

Fishrock123:
Updated, now changes `*_mime()` to be `content_type`.

PR #552, opened 2020-05-28
Fix unix tests

yoshuawuyts:
Noticed https://github.com/http-rs/tide/pull/539 was failing, which seems caused by #531 not having been updated for master. This patch fixes that. Thanks!


yoshuawuyts:
Fixes the CI failure on master -- merging!

PR #553, opened 2020-05-28
Add AsRef/AsMut<Headers> for Request/Response

yoshuawuyts:
Adds `{AsRef,AsMut}<Headers>` impls for `Request` and `Response`. This is useful for generic functions that want to operate on `http_types::Headers`. With this Tide's req/res types can be passed directly to functions that operate on `http-types`'s types:
```rust
fn set_custom(headers: impl AsMut<http_types::Headers>) {
    headers.insert("x-custom", "meow");
}

set_custom(&mut req);
set_custom(&mut res);
```

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -264,15 +264,27 @@ impl Response {
     }
 }
 
+impl AsRef<http::Response> for Response {
+    fn as_ref(&self) -> &http::Response {
+        &self.res
+    }
+}
+
 impl AsMut<http::Response> for Response {
     fn as_mut(&mut self) -> &mut http::Response {
         &mut self.res
     }
 }
 
-impl AsRef<http::Response> for Response {
-    fn as_ref(&self) -> &http::Response {
-        &self.res
+impl AsRef<http::Headers> for Response {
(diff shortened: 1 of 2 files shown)


src/response.rs
```diff
@@ -264,15 +264,27 @@ impl Response {
...
         &mut self.res
     }
 }
 
-impl AsRef<http::Response> for Response {
-    fn as_ref(&self) -> &http::Response {
-        &self.res
+impl AsRef<http::Headers> for Response {
+    fn as_ref(&self) -> &http::Headers {
+        self.as_ref()
```
jbr, line 281: I assume this should be `self.res.as_ref()`?

yoshuawuyts: ooops, yes

yoshuawuyts:
@Fishrock123 I was thinking probably not for `Body` -- we already forward `AsyncRead` -- but perhaps for `Url`, `StatusCode` and `Method` as well. But the right way to go about those is to add them to `http-types` first, and then forward them from `tide` second.

PR #555, opened 2020-05-28
Use query from http types

plippe:
https://github.com/http-rs/http-types/issues/154


plippe:
I don't believe I am doing the "right" thing here. I was expecting some tests to fail locally. 
I will have a deeper look tomorrow.

PR #556, opened 2020-05-28
introduce the ideas of Before and After middlewares

jbr:
For discussion, a _functional_ sketch

### context
Due to the lack of AsyncFn types in rust currently, we cannot describe the bounds for an async function that has a lifetime in the arguments. As a result, there's no way to write
```rust
// this example is not possible
app.middleware(|req: Request<()>, next: Next<()>| async move {
  next.run(req).await
});
```

and instead function middlewares look like:
```rust
// current signature
fn minimal_middleware<'a>(
    request: Request<()>,
    next: Next<'a, ()>,
) -> Pin<Box<dyn Future<Output = Result> + Send + 'a>> {
    Box::pin(async { next.run(request).await })
}

//--snip--

app.middleware(minimal_middleware);
```

### proposed solution

Because the issue is that Next is not owned, it is possible to write simple async signatures for middlewares that exist on either side of the Next call.  This PR called them `Before` and `After` middlewares currently, but those names probably can be improved. This PR is predicated on an assumption that a meaningful percentage of app-written middlewares exist as transformations of the request OR the response. This enables the following code.

```rust
// possible with this pr
app.middleware(Before(|request: Request<()>| async move { request })));
app.middleware(After(|result: Result| async move { result })));
```

### DIFF
--- a/src/middleware.rs
+++ b/src/middleware.rs
@@ -27,6 +27,23 @@ pub trait Middleware<State>: 'static + Send + Sync {
     }
 }
 
+/// Define a middleware that operates on incoming requests.
+///
+/// This middleware is useful because it is not possible in Rust yet to use
+/// closures to define inline middleware.
+///
+/// # Examples
+///
+/// ```rust
+/// use tide::{Before, Request};
+/// use std::time::Instant;
+///
+/// let mut app = tide::new();
+/// app.middleware(Before(|mut request: Request<()>| async move {
+///     request.set_ext(Instant::now());
+///     request
+/// }));
+/// ```
 #[derive(Debug)]
 pub struct Before<F>(pub F);
 impl<State, F, Fut> Middleware<State> for Before<F>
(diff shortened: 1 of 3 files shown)


src/middleware.rs
```diff
@@ -27,6 +27,46 @@ pub trait Middleware<State>: 'static + Send + Sync {
...
+        next: Next<'a, State>,
+    ) -> BoxFuture<'a, crate::Result> {
+        Box::pin(async move {
+            let request = (self.0)(request).await;
+            next.run(request).await
+        })
+    }
+}
+
+#[derive(Debug)]
```
yoshuawuyts, line 50: This should have documentation. Bonus if we can add usage examples.

src/middleware.rs
```diff
@@ -27,6 +27,46 @@ pub trait Middleware<State>: 'static + Send + Sync {
     }
 }
 
+#[derive(Debug)]
```
yoshuawuyts, line 30: This should have documentation. Bonus if we can add usage examples.

PR #562, opened 2020-05-29
Bring response in line with http-types

yoshuawuyts:
This is a follow-up to #537. Brings Response in line with Request. Thanks!

## Changes
- Removed `Response::redirect` in favor of `tide::Redirect`
- Renamed `Reponse::set_cookie` to `Response::insert_cookie`.
- Various `Response` methods no longer return `Self`.
- `Response::new` now accepts `u16` as well as `StatusCode` as arguments.
- Added `Response::header_mut`
- Renamed `Response::set_header` to `Response::insert_header`.
- Removed `Response::{body_string, body_form, body_json, body}` in favor of `Response::set_body`.
- Added `Response::swap_body`.
- Renamed `Reponse::set_ext` to `Response::insert_ext`.

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -157,7 +157,7 @@ impl Response {
     /// # fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
     /// # async_std::task::block_on(async {
     /// #
-    /// use tide::{Body, Response};
+    /// use tide::Response;
     ///
     /// let mut req = Response::new(200);
     /// req.set_body("Hello, Nori!");
--- a/src/response.rs
+++ b/src/response.rs
@@ -29,7 +29,7 @@ impl Response {
         S: TryInto<StatusCode>,
         S::Error: Debug,
     {
-        let res = http_types::Response::new(status);
+        let res = http::Response::new(status);
         Self {
             res,
             cookie_events: vec![],
--- a/src/response.rs
+++ b/src/response.rs
@@ -176,7 +176,7 @@ impl Response {
     }
 
     /// Insert cookie in the cookie jar.
-    pub fn insert_cookie(&mut self, cookie: Cookie<'static>) {
+    pub fn append_cookie(&mut self, cookie: Cookie<'static>) {
         self.cookie_events.push(CookieEvent::Added(cookie));
     }
 
--- a/src/response.rs
+++ b/src/response.rs
@@ -176,7 +176,7 @@ impl Response {
     }
 
     /// Insert cookie in the cookie jar.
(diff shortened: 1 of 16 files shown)


src/response.rs
```diff
@@ -27,52 +24,18 @@ pub struct Response {
...
     where
-        R: BufRead + Unpin + Send + Sync + 'static,
+        S: TryInto<StatusCode>,
+        S::Error: Debug,
     {
-        let status = crate::StatusCode::try_from(status).expect("invalid status code");
-        let mut res = http_types::Response::new(status);
-        res.set_body(Body::from_reader(reader, None));
-
+        let res = http_types::Response::new(status);
```
Fishrock123, line 32:
```suggestion
        let res = http::Response::new(status);
```

src/response.rs
```diff
@@ -161,85 +128,15 @@ impl Response {
...
-        self
-    }
-
-    /// Pass a string as the response body.
-    ///
-    /// # Mime
-    ///
-    /// The content type is set to `text/plain; charset=utf-8`.
-    #[must_use]
-    pub fn body_string(mut self, string: String) -> Self {
```
Fishrock123, line 176: Not so sure about removing these - are we depending on always mime sniffing now? There's definitely some overhead if that's the case.

Otherwise the docs should probably be clear that the user needs to set it themselves.

yoshuawuyts: > Not so sure about removing these - are we depending on always mime sniffing now? There's definitely some overhead if that's the case.

These methods have been moved to rely on [`Body`](https://docs.rs/http-types/2.1.0/http_types/struct.Body.html) instead and go through the `set_body` method instead:

```rust
// old API
res.body(io::Cursor::new(b"hello world".to_owned()));
res.body_string("hello world".to_string());

// new API
res.set_body(b"hello world");
res.set_body("hello world");
```

The only set of APIs that become a bit more awkward by this are the Serde-related APIs. But at least https://github.com/http-rs/tide/pull/523 is showing us improvements for some applications, and with a `impl From<Body> for Response` bound we could move this even further ahead (in a future patch):

```rust
// old API
res.body_json(Cat { name: "chashu" });

// new API
res.set_body(Body::from_json(Cat { name: "chashu" });

// http-rs/tide#523 (return type from function)
Ok(json!({ "name": "chashu" }));

// future option #1 
res.set_body(json!({ "name": "chashu" });

// future option #2 (return type from function)
Ok(Body::from_json(Cat { name: "chashu" }));
```

Either way, I hope this shows what the intended workflow around Response bodies is with this patch, and shows some options we could explore in the future to improve this further.

PR #565, opened 2020-06-01
Delete CHANGELOG.md

yoshuawuyts:
CHANGELOG.md hasn't been updated in a while in favor of GH releases. This removes it which should make it less confusing where track changes. Thanks!


Fishrock123:
I, personally, like in-source changelogs.

yoshuawuyts:
@Fishrock123 I literally do not have the time to keep multiple changelogs up to date. If we had better tooling -- maybe. But most projects in `http-rs` use GH releases to track changes, and I would like to have more consistency between them.

PR #566, opened 2020-06-01
Add missing Debug impls

yoshuawuyts:
Adds missing Debug impls instead of the `#[allow(missing_debug_impls)]` we had before. Thanks!

### DIFF
--- a/src/middleware.rs
+++ b/src/middleware.rs
@@ -1,5 +1,6 @@
 //! Middleware types.
 
+use std::fmt::Debug;
 use std::sync::Arc;
 
 use crate::endpoint::DynEndpoint;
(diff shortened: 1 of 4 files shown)


Fishrock123:
If all this does is print the struct properties, isn't there like `#[Derive(Debug)]` or something?

kyrias:
It doesn't print any of the properties, just the name of the structs.

src/middleware.rs
```diff
@@ -139,3 +139,9 @@ impl<'a, State: 'static> Next<'a, State> {
         }
     }
 }
+
+impl<State> Debug for Next<'_, State> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Next").finish()
```
jbr, line 145: Since Middleware has a `name()` now, would it make sense for this to include `self.next_middleware.iter().map(|m| m.name())` somehow? So inspecting Next would make clear what came after the middleware?

yoshuawuyts:
Stale; going to go ahead and close.

PR #571, opened 2020-06-06
Fix panic in sse endpoint

lord:
The `sse.rs` example was panicking for me; this seems to fix it.


lord:
Looks like there's a docs error but doesn't look like a result from this change?

yoshuawuyts:
Ah dang, I missed this. We closed this through #588 #590 after I got a DM on twitter about it. Pushing a release today. Thanks heaps for this!

lord:
all good! thanks for fixing!

PR #580, opened 2020-06-10
Response builder and related changes

jbr:
This PR does several things that are related, but could be distinct PRs. Please let me know if you'd prefer I break them out for independent review.

* changes insert_ext from taking `mut self` to taking `&mut self` to conform to the rest of the response interface
* allows Response::set_status to take any `TryInto<StatusCode>` like `Response::new()` does
```rust
res.set_status(203);
```
* ~~provides a `From<Body> for Response` and uses that conversion in other Body -> Response conversions that already existed.~~ Extracted to #584 

* adds a response builder:
```rust
app.at("/").get(|_| async {
    Ok(Response::builder(203)
        .body("body")
        .header("custom-header", "value"))
})
```

### DIFF
--- a/examples/chunked.rs
+++ b/examples/chunked.rs
@@ -1,14 +1,10 @@
 use async_std::task;
-use tide::{Body, Response, StatusCode};
 
 fn main() -> Result<(), std::io::Error> {
     task::block_on(async {
         let mut app = tide::new();
-        app.at("/").get(|_| async {
-            let mut res = Response::new(StatusCode::Ok);
-            res.set_body(Body::from_file(file!()).await.unwrap());
-            Ok(res)
-        });
+        app.at("/")
+            .get(|_| async { Ok(tide::Body::from_file(file!()).await?) });
         app.listen("127.0.0.1:8080").await?;
         Ok(())
     })
--- a/src/response_builder.rs
+++ b/src/response_builder.rs
@@ -1,0 +1,53 @@
+use crate::http::headers::{HeaderName, ToHeaderValues};
+use crate::http::{Body, Mime, StatusCode};
+use crate::Response;
+use std::convert::TryInto;
+
+#[derive(Debug)]
+pub struct ResponseBuilder(Response);
+
+impl ResponseBuilder {
+    pub fn new() -> Self {
+        Self(Response::new(StatusCode::Ok))
+    }
+
+    pub fn unwrap(self) -> Response {
+        self.0
+    }
+
+    pub fn status<S>(mut self, status: S) -> Self
+    where
+        S: TryInto<StatusCode>,
+        S::Error: std::fmt::Debug,
+    {
+        self.0.set_status(status);
+        self
+    }
+
+    pub fn header(mut self, key: impl Into<HeaderName>, value: impl ToHeaderValues) -> Self {
+        self.0.insert_header(key, value);
+        self
+    }
+
+    pub fn content_type(mut self, content_type: impl Into<Mime>) -> Self {
+        self.0.set_content_type(content_type);
+        self
+    }
+
+    pub fn body(mut self, body: impl Into<Body>) -> Self {
+        self.0.set_body(body);
+        self
+    }
+}
+
+impl Into<Response> for ResponseBuilder {
+    fn into(self) -> Response {
+        self.unwrap()
+    }
+}
+
+impl Into<crate::Result> for ResponseBuilder {
+    fn into(self) -> crate::Result {
+        Ok(self.unwrap())
+    }
+}
--- a/src/response_builder.rs
+++ b/src/response_builder.rs
@@ -4,17 +4,55 @@ use crate::Response;
 use std::convert::TryInto;
 
 #[derive(Debug)]
+
+/// Response Builder
+///
+/// Provides an ergonomic way to chain the creation of a response. This is generally accessed through [`Response::build`](crate::Response::build)
+///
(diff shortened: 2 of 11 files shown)


Fishrock123:
Good point, I think for some reason I was thinking this was in `http-types`.

yoshuawuyts:
Thanks so much for kicking this off @jbr! I'm not quite convinced on builders yet, but think this is def worth exploring. With a bit of luck this indeed turns out to be a much more convenient way to create Responses and that would be neat. I do have a few design notes though:

1. We should rename the main method to `Response::builder`; `build` is the conventional name for the finalizer method, `builder` for the constructor.
2. `Response::builder` should take `Into<StatusCode>` as the argument. It's 3 extra characters, that in many cases will save an extra line. It also creates parity with `Response::new` which doesn't assume a default status code either. For surf we'd require any form of `Request::builder` to take a `url` as an argument, so this creates symmetry there as well.
3. Many of the examples use `.into()` as the finalizer. We should add `build` as an explicit method for this as well.
4. We can remove the need to invoke the finalizer method in most cases if we change the return type of the enclosing function from `-> tide::Result` into `-> tide::Result<Into<Response>>`. This allows `ResponseBuilder::Into` to be called behind the scenes and saves an extra line in many cases.
4b. Note that we can't yet express `tide::Result` as `http_types::Result<impl Into<tide::Response>>`. This requires lang support.
5. In the examples for `ResponseBuilder` we should show the path we want people to take: namely by importing `tide::Response` and calling `Response::builder`. That's the canonical way to gain access to the struct, and I think we should use that in our examples as well.

jbr:
Addressed 1, 2, 3, and 5. I think 4 is a big enough task it's worth splitting out as a distinct PR

src/response_builder.rs
```diff
@@ -0,0 +1,90 @@
...
+    fn into(self) -> Response {
+        self.build()
+    }
+}
+
+impl Into<crate::Result> for ResponseBuilder {
+    fn into(self) -> crate::Result {
+        Ok(self.build())
+    }
+}
```
yoshuawuyts, line 90: Not too sure about this bound for similar reasons reasons outlined in https://github.com/http-rs/tide/pull/596#issuecomment-647162442.

tests/chunked-encode-small.rs
```diff
@@ -21,13 +19,8 @@ async fn chunked_large() -> Result<(), http_types::Error> {
...
         let mut app = tide::new();
-        app.at("/").get(|_| async {
-            let mut res = Response::new(StatusCode::Ok);
-            let body = Cursor::new(TEXT.to_owned());
-            res.set_body(Body::from_reader(body, None));
-            res.set_content_type(mime::PLAIN);
-            Ok(res)
-        });
+        app.at("/")
+            .get(|_| async { Ok(Body::from_reader(Cursor::new(TEXT), None)) });
```
yoshuawuyts, line 23: Isn't it possible to rewrite this as:

```suggestion
        app.at("/").get(|_| async { Ok(TEXT) });
```

tests/chunked-encode-large.rs
```diff
@@ -72,13 +70,8 @@ async fn chunked_large() -> Result<(), http_types::Error> {
...
         let mut app = tide::new();
-        app.at("/").get(|mut _req: tide::Request<()>| async {
-            let mut res = Response::new(StatusCode::Ok);
-            let body = Cursor::new(TEXT.to_owned());
-            res.set_body(Body::from_reader(body, None));
-            res.set_content_type(mime::PLAIN);
-            Ok(res)
-        });
+        app.at("/")
+            .get(|_| async { Ok(Body::from_reader(Cursor::new(TEXT), None)) });
```
yoshuawuyts, line 74: Isn't it possible to rewrite this as:

```suggestion
        app.at("/").get(|_| async { Ok(TEXT) });
```

examples/chunked.rs
```diff
@@ -1,14 +1,10 @@
...
 fn main() -> Result<(), std::io::Error> {
     task::block_on(async {
         let mut app = tide::new();
-        app.at("/").get(|_| async {
-            let mut res = Response::new(StatusCode::Ok);
-            res.set_body(Body::from_file(file!()).await.unwrap());
-            Ok(res)
-        });
+        app.at("/")
+            .get(|_| async { Ok(tide::Body::from_file(file!()).await?) });
```
yoshuawuyts, line 7: As an aside: would be fun if we could add a bound where we have `From<Path> for Response` that turns this into a shorthand. -- Wouldn't quite work in this case since it's a `&'static str`, but probably in some other cases it would.

tests/chunked-encode-small.rs
jbr: It is! And I initially made that change. But the purpose of this test is to explicitly make the body chunked, and Ok(TEXT) doesn't stream the result

tests/chunked-encode-large.rs
jbr: Same as above, this test is explicitly a chunked test, so making that change would change the purpose of the test

src/response_builder.rs
jbr: I'll back this out and the usages of it

examples/chunked.rs
jbr: I looked into this as a distinct PR, but it requires an awkward task::block_on because From::from isn't async so I'm not convinced it's worthwhile

PR #587, opened 2020-06-11
Don't crash the server when there's a listen error

### LINKED ISSUE #577: "Too many open files" error when benchmarking
I wanted to benchmark the example in the readme but got this `Error: Os { code: 24, kind: Other, message: "Too many open files" }` error.

- macOS 10.15.5
- 2.3 GHz 8-Core Intel Core i9 (16 logical cores)
- rustc 1.44.0 (49cae5576 2020-06-01)

```bash
$ ulimit --all
Maximum size of core files created                           (kB, -c) 0
Maximum size of a process’s data segment                     (kB, -d) unlimited
Maximum size of files created by the shell                   (kB, -f) unlimited
Maximum size that may be locked into memory                  (kB, -l) unlimited
Maximum resident set size                                    (kB, -m) unlimited
Maximum number of open file descriptors                          (-n) 256
Maximum stack size                                           (kB, -s) 8192
Maximum amount of cpu time in seconds                   (seconds, -t) unlimited
Maximum number of processes available to a single user           (-u) 5568
Maximum amount of virtual memory available to the shell      (kB, -v) unlimited
```

Code:
```rust
use std::io;

#[async_std::main]
async fn main() -> io::Result<()> {
    let mut app = tide::new();
    app.at("/hello").get(|_| async { Ok("Hello, world!") });
    app.listen("127.0.0.1:8000").await?;
    Ok(())
}
```

Dependencies:
```toml
tide = "0.10.0"
async-std = { version = "1.6.0", features = ["attributes"] }
```

Tide command and output (same for both `"/"` and `"/hello"`):
```bash
$ cargo run --release
    Finished release [optimized] target(s) in 0.13s
     Running `target/release/test_tide`
Error: Os { code: 24, kind: Other, message: "Too many open files" }
```

`wrk` command and output for `"/"`:
```bash
$ wrk -t10 -c1000 -d10s --latency http://127.0.0.1:8000
Running 10s test @ http://127.0.0.1:8000
  10 threads and 1000 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     3.05ms    1.29ms   5.36ms   63.78%
    Req/Sec   380.00    229.91     0.85k    70.00%
  Latency Distribution
     50%    2.19ms
     75%    4.45ms
     90%    4.80ms
     99%    5.13ms
  392 requests in 10.07s, 31.39KB read
  Socket errors: connect 757, read 1702, write 8, timeout 0
  Non-2xx or 3xx responses: 392
Requests/sec:     38.91
Transfer/sec:      3.12KB
```

`wrk` command and output for `"/hello"`:
```bash
$ wrk -t10 -c1000 -d10s --latency http://127.0.0.1:8000/hello
Running 10s test @ http://127.0.0.1:8000/hello
  10 threads and 1000 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     3.19ms    1.53ms   7.90ms   86.61%
    Req/Sec   419.80    216.12   817.00     60.00%
  Latency Distribution
     50%    2.53ms
     75%    4.49ms
     90%    4.80ms
     99%    7.84ms
  433 requests in 10.06s, 54.55KB read
  Socket errors: connect 757, read 1521, write 22, timeout 0
Requests/sec:     43.05
Transfer/sec:      5.42KB
```

jbr:
This change is an attempt to address #577, and also seems like correct behavior for a server. Regardless of what goes wrong with any individual request, we attempt to continue to respond to subsequent requests. Previously, the use of ? was returning from the `listen` function scope, which shut down the server.

Extracting the inner functions was mostly for clarity and will be further refactored/extracted in future PRs

closes #577

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -275,6 +275,23 @@ impl<State: Send + Sync + 'static> Server<State> {
         self
     }
 
+    #[cfg(feature = "h1-server")]
+    async fn handle_tcp(&self, stream: io::Result<async_std::net::TcpStream>) -> crate::Result<()> {
+        let stream = stream?;
+        let this = self.clone();
+        let local_addr = stream.local_addr().ok();
+        let peer_addr = stream.peer_addr().ok();
+        task::spawn(async move {
+            async_h1::accept(stream, |mut req| async {
+                req.set_local_addr(local_addr);
+                req.set_peer_addr(peer_addr);
+                this.respond(req).await
+            })
+            .await
+        })
+        .await
+    }
+
     /// Asynchronously serve the app at the given address.
     #[cfg(feature = "h1-server")]
     pub async fn listen(self, addr: impl ToSocketAddrs) -> io::Result<()> {
@@ -291,25 +308,11 @@ impl<State: Send + Sync + 'static> Server<State> {
 
         let mut incoming = listener.incoming();
         while let Some(stream) = incoming.next().await {
-            let stream = stream?;
-            let this = self.clone();
-            let local_addr = stream.local_addr().ok();
-            let peer_addr = stream.peer_addr().ok();
-            task::spawn(async move {
-                let res = async_h1::accept(stream, |mut req| async {
-                    req.set_local_addr(local_addr);
-                    req.set_peer_addr(peer_addr);
-                    let res = this.respond(req).await;
-                    let res = res.map_err(|_| io::Error::from(io::ErrorKind::Other))?;
-                    Ok(res)
-                })
-                .await;
-
-                if let Err(err) = res {
-                    log::error!("async-h1 error", { error: err.to_string() });
-                }
-            });
+            if let Err(error) = self.handle_tcp(stream).await {
+                log::error!("async-h1 error", { error: error.to_string() });
+            }
         }
+
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
I think we should merge this; though the more structural fix seems to be https://github.com/http-rs/tide/pull/386 by @tailhook which allows setting more fine-grained limits. Happy to merge this in the interim though!

tailhook:
As far as I can see this will hang single CPU with 100% usage in case `ENFILE` or `EMFILE` is encountered (depending on one's use case it may be worse than just crashing, because in case of crashing supervisor would restart the process). I've tried to explain that in the book: https://book.async.rs/patterns/accept-loop.html#handling-errors Let me know if book is not clear enough on this issue.

Sorry, for not having enough time for finishing #386

src/server.rs
```diff
@@ -333,28 +352,40 @@ impl<State: Send + Sync + 'static> Server<State> {
...
-                    Ok(res)
-                })
-                .await;
-
-                if let Err(err) = res {
-                    log::error!("async-h1 error", { error: err.to_string() });
+            let stream = match stream {
+                Err(ref e) if is_transient_error(e) => continue,
+                Err(error) => {
+                    let delay = std::time::Duration::from_millis(500);
```
Fishrock123, line 358: This seems kind of large? I guess it should be pretty unusual so maybe it's fine. Ideally we'd have a dynamically increasing backoff here.

jbr: I didn't have a good sense of an appropriate number here so I used the same number in the book section cited above

tailhook: Well, it's 500ms in the book because it's reasonably fine to write a log message every 500ms. Skipping log message always will make the issue a nightmare to debug. Sleeping less but keeping message rarely will substantially complicate the code (which is fine for prod, but bad for a book).

But at the end of the day, you should not expect code to hit this timeout for a long time. I always consider this as an infrastructure issue: either bump file descriptor limit if you have enough memory, or lower listen queue, or put a load-balancer/queue in front of the process to get rid of error. So large timeout here is fine. Also, TCP delays can be larger than 500ms, so any client should be fine to wait that much.

jbr:
Merging this based on @yoshuawuyts' comment above, especially since this is a short-term fix for multiple reasons:

* When/if async-listen is Clone (tailhook/async-listen#3), tide will be able to use that instead of this code
* When tide supports multiple simultaneous listeners (a near-term goal), we'll need a totally different approach that pools open fds across multiple listeners of different types, and that will require rewriting this yet again

PR #588, opened 2020-06-12
fix SSE unwrap

yoshuawuyts:
Reported in https://twitter.com/fasterthanlime/status/1271398187484491776. Thanks!


yoshuawuyts:
Fixing CI in https://github.com/http-rs/tide/pull/589. Once that passes we should re-run CI here.

yoshuawuyts:
> Okay, quite convinced this is an issue with crates.io, GitHub Actions, or the interaction between them. I can't reproduce this locally with either Windows or Linux; so just going to pray this passes.

Merging this!

fasterthanlime:
@yoshuawuyts note that there's *another* insert_header in `upgrade`

PR #589, opened 2020-06-12
Fix CI

yoshuawuyts:
tempdir is now broken, use tempfile crate instead. Spotted CI failure in https://github.com/http-rs/tide/pull/588/checks?check_run_id=765236618. Thanks!


yoshuawuyts:
Okay, it seems Cargo is sharing broken results? https://crates.io/crates/remove_dir_all 0.5.2 is published and hasn't been changed in over a year. Unsure what's going on.

yoshuawuyts:
Okay, quite convinced this is an issue with crates.io, GitHub Actions, or the interaction between them. I can't reproduce this locally with either Windows or Linux; so just going to pray this passes.

Regardless still a good idea to deprecate this crate in favor of one that's still maintained.

PR #598, opened 2020-06-15
return a result from sse sender

### LINKED ISSUE #591: SSE handler never dropped
First off: thanks for v0.11.0!

Here's a repro for the issue I [raised on twitter](https://mobile.twitter.com/fasterthanlime/status/1271482486158241793):

  * https://github.com/fasterthanlime/sse-never-dies

just `cargo run`, it gives instructions on how to run `curl` in a separate terminal to see the effect.

## Expected behavior

When a client disconnect, the internal SSE task gets dropped, and so does the handler Future, resulting in `PrintOnDrop` being dropped.

## Observed behavior

The internal SSE task never stops running, the handler is never dropped, it just gets stuck on `send` after a while - presumably after it's filled the internal buffer of sse::Encoder?

jbr:
in order to allow endpoints to break out of loops

dependent on: http-rs/async-sse#3

closes #591

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ __internal__bench = []
 
 [dependencies]
 async-h1 = { version = "2.0.1", optional = true }
-async-sse = "3.0.0"
+async-sse = "3.0.1"
 async-std = { version = "1.6.0", features = ["unstable"] }
 femme = "2.0.1"
 http-types = "2.2.1"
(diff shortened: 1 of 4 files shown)


jbr:
This is now blocked on an async-sse release (which I think is a major release because this changes the api)

jbr:
Merging this because async-sse 3.0.1 was published, breaking tide master (and tide 0.11). We'll probably want to yank async-sse 3.0.1 and republish as 4.0.0, which will require bumping tide's dep on master to async-sse 4.0.0, but merging this will fix tide master CI

yoshuawuyts:
@jbr thanks for catching that and running the release!

PR #601, opened 2020-06-16
introduce an extension trait for testing servers

jbr:
This is a first step in the direction of a [superagent](https://visionmedia.github.io/superagent/)/[supertest](https://github.com/visionmedia/supertest) type testing utility for tide. Currently it just supports simple requests and is not exported for usage outside of tide, but the intent is that as it accumulates more features and stabilizes, it might eventually be public under a tide::testing namespace for tide apps to use.

```rust
// this:
    let req = Request::new(
        Method::Get,
        Url::parse("http://example.com/foo/echo").unwrap(),
    );
    let mut res: Response = app.respond(req).await.unwrap();

// becomes:
    let mut res = app.get("/foo/echo").await;
```

and

```rust
 // this:
    let req = http::Request::new(
        Method::Get,
        Url::parse("http://localhost/add_one/3").unwrap(),
    );
    let mut res: http::Response = app.respond(req).await.unwrap();
    assert_eq!(&res.body_string().await.unwrap(), "4");

// becomes:
    assert_eq!(app.get_body("/add_one/3").await, "4");
```

### DIFF
--- a/tests/nested.rs
+++ b/tests/nested.rs
@@ -1,7 +1,4 @@
-use http_types::{Method, Request, Response, Url};
-use test_utils::BoxFuture;
-use tide::{Middleware, Next};
-
+use test_utils::ServerTestingExt;
 mod test_utils;
 
 #[async_std::test]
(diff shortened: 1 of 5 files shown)


src/endpoint.rs
```diff
@@ -66,6 +66,12 @@ where
     }
 }
 
+impl<State> Endpoint<State> for &'static str {
```
Fishrock123, line 69: Why's this in the same change?

jbr: See note above ☝️

Fishrock123: It'd be nice to have it in a separate commit

jbr: Entirely backed it out, wasn't used anywhere

Fishrock123:
still lgtm

Issue #593, opened 2020-06-13
* route priority


spacekookie:
Currently `serde_dir()` seems to be broken. It works if it is the only route in an app, but fails to is never routed to when it's not.

```rust
pub(crate) fn setup(app: &mut Server<StateRef>, root: PathBuf) {
    app.at("/static").serve_dir(root.join("static")).unwrap();
}
```

This works, but the following does not: 

```rust
pub(crate) fn setup(app: &mut Server<StateRef>, root: PathBuf) {
    app.at("/:owner/:repo").get(|_| async move { Ok("Hello world!") });
    app.at("/static").serve_dir(root.join("static")).unwrap();
}
```

jbr:
I believe this is the same issue as #595 — if you add a distinct prefix to the dynamic route, like `app.at("/api/:owner/:repo").get(|_| async { Ok("hello world") });`, it will route correctly

spacekookie:
That's not really doable for me though because I'm not building an api, I need the paths to be what users see because I'm rendering templates. The /static route needs to exist for the templates o pull in css files

Fishrock123:
Does this work? That is, switching registration order?

```rust
pub(crate) fn setup(app: &mut Server<StateRef>, root: PathBuf) {
    app.at("/static").serve_dir(root.join("static")).unwrap();
    app.at("/:owner/:repo").get(|_| async move { Ok("Hello world!") });
}
```

`at("/:param")` should reasonably capture anything that is `/v`, `/v/`, `/v/*`, but `at("/:p/:p")` should not match just `/v` or `/v/` - if it does that's a bug, imo.

jbr:
I've reproduced this problem and I'm not sure what the right solution is, but here's an update. Route prioritization exists in [code outside of http-rs](https://github.com/http-rs/route-recognizer/blob/master/src/lib.rs#L30-L48).

This issue can be reduced to the following failing test (using the test helpers in #601 for convenience):
```rust
#[async_std::test]
async fn static_routing() {
    let mut app = tide::Server::new();
    app.at("/:one/:two").get("two segments");
    app.at("/a/*").get("static + star");
    assert_eq!(app.get_body("/a/a").await, "static + star");
}

```

which results in

```
---- static_routing stdout ----
thread 'static_routing' panicked at 'assertion failed: `(left == right)`
  left: `"two segments"`,
 right: `"static + star"`', tests/wildcard.rs:127:5
```

[...]

PR #603, opened 2020-06-16
make ServeDir public

jbr:
In trying to find a workaround for #593, it turned out to be really useful to be able to create a ServeDir outside of tide itself.  With this PR, the following workaround for conflicting routing is possible:

```rust
use tide::{fs::ServeDir, Endpoint, Request};

#[async_std::main]
    tide::log::start();
    let mut app = tide::new();
    app.at("/static").serve_dir("./src")?;
    app.at("/:owner/:repo")
        .get(|request: Request<_>| async move {
            if request.param::<String>("owner").unwrap() == "static" {
                ServeDir::new("/static", "./src").call(request).await
            } else {
                Ok("do whatever we'd be doing otherwise".into())
            }
        });

    Ok(())
}
```

---

Although this use is not especially normal, there might reasonably be other situations in which people would want to conditionally use ServeDir logic on a route that is not mounted by Route::serve_dir

### DIFF
--- a/src/fs/mod.rs
+++ b/src/fs/mod.rs
@@ -1,3 +1,3 @@
 mod serve_dir;
 
-pub(crate) use serve_dir::ServeDir;
+pub use serve_dir::ServeDir;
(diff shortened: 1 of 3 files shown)


jbr:
update: I have no idea if this is worthwhile

PR #606, opened 2020-06-18
do not await the spawned task in listen

jbr:
this fixes a critical bug identified in discord, in which multiple simultaneous requests hang indefinitely, introduced in ff00be9414bab3899bb3364bab14390d6e4a641d. Prior to that commit, tide was not awaiting the spawned task, which was the source of this problem.


Fishrock123:
retroactive lgtm

PR #607, opened 2020-06-18
use route-recognizer 0.2

### LINKED ISSUE #593: * route priority

jbr:
closes #593 
refs http-rs/route-recognizer#40

This doesn't seem like an ideal solution, but this resolves the conflicting routes issue until there's a route-recognizer release

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,9 +38,9 @@ async-std = { version = "1.6.0", features = ["unstable"] }
 femme = "2.0.1"
 http-types = "2.0.1"
 kv-log-macro = "1.0.4"
-route-recognizer = "0.1.13"
 serde = "1.0.102"
 serde_json = "1.0.41"
+route-recognizer = "0.2.0"
 
 [dev-dependencies]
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
@jbr this will prevent us from publishing new releases. Happy to release a new version tho!

yoshuawuyts:
Published route-recognizer 0.2.0; think we can update to that now!

jbr:
Yay! That’s the result I was hoping for, but it sounded like that was unlikely to happen. Didn’t know whose call that was. Does crates.io still depend on route recognizer?

PR #609, opened 2020-06-20
Change log macro to be directly referenced

austinabell:
Someone had an issue compiling Tide, and although I can't replicate it seems maybe some compiler versions/env don't like this circular reference (more of a guess). I don't see any reason why this shouldn't be this way anyway, but let me know if I'm missing something :)

https://github.com/ChainSafe/forest/issues/510

### DIFF
--- a/src/log/mod.rs
+++ b/src/log/mod.rs
@@ -27,11 +27,11 @@ pub use middleware::LogMiddleware;
 /// Start logging.
 pub fn start() {
     femme::start();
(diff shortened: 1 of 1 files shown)


jbr:
This was an issue with tide 0.9 and femme, I believe. Does cargo update resolve this? If not, I believe updating to tide 0.11 will, in which case the patch is unnecessary

rtyler:
@austinabell I was also bit by this, and as @jbr mentioned, running `cargo update` will fix it.

Issue #586, opened 2020-06-11
Allow for configuration of maximum server concurrency limits


jbr:
When servers support multiple listeners, this should be expanded to include both a global maximum as well as a per-listener maximum. This could either be a hard ceiling in which the server does not accept new connections at all until open connections are closed, or a soft ceiling at which an efficient static 503 response is sent instead of running the router.

References #577

Fishrock123:
Could also do both a soft and hard cap.

tailhook:
The hard limit is needed because otherwise, while you're responding with 503, your other request can be non-functional (can't open file or an upstream connection because file descriptor limit is exhausted). Having both makes sense, though.

Per-listener limit is clear how to do. With global limit it's a bit more complex. Because at the moment there is a backpressure, you have to choose which listener to get connections from first. And that might mean that other listeners are starving (in case there are priority listeners), or vice versa loosing performance on polling listeners having no connections if trying to do "fair" accept (did not benchmark if that performance difference is significant though).

andrewbanchich:
I think it would also be useful if `tide` could respond with an [HTTP 429](https://httpstatuses.com/429) when it approaches some global concurrency limit.

Functions like [`Body::from_file`](https://docs.rs/tide/0.13.0/tide/struct.Body.html#method.from_file) could also check system limits on file descriptors and respond with a 429 as it gets close to the system limit.

jbr:
Thanks for reviving this issue — I forgot about it

PR #610, opened 2020-06-21
extensible entrypoint for app.listen

jbr:
image
image
image

The idea is that it should be possible for a third-party crate to implement `tide::listener::ToListener` and `tide::listener::Listener` for any newtype (eg multiple varieties of TlsListener) and support passing it to Server::listen and adding it to a MultiListener alongside other Listeners.

Working experimental example of tls on tide: https://github.com/jbr/tide-tls

---

Note for future changes:

In order to provide hard/soft concurrency limits on open requests as per #586, the Listener trait will likely need to change to include a hook of some sort for backpressure, but that was beyond the scope of this already-large PR. Similarly, the Listener trait would be a good place to introduce a graceful shutdown mechanism


src/listener/multi_listener.rs
```diff
@@ -0,0 +1,50 @@
...
+
+impl<State: 'static + Send + Sync> Listener<State> for MultiListener<State> {
+    fn listen<'a>(&'a mut self, app: Server<State>) -> BoxFuture<'a, io::Result<()>> {
+        let mut fut: Option<BoxFuture<'a, io::Result<()>>> = None;
+
+        for listener in self.0.iter_mut() {
+            let app = app.clone();
+            let listened = listener.listen(app);
+            if let Some(f) = fut {
+                fut = Some(Box::pin(f.race(listened)));
```
jbr, line 42: This is awkward and improvements are very welcome. I was looking for a way to race an iter of an arbitrary number of futures and this was what I came up with

src/listener/tcp_listener.rs
```diff
@@ -0,0 +1,80 @@
...
+use crate::utils::BoxFuture;
+use crate::{log, Server};
+
+use async_std::net::{self, SocketAddr, TcpStream};
+use async_std::prelude::*;
+use async_std::{io, task};
+
+#[derive(Debug)]
+pub enum TcpListener {
+    FromListener(net::TcpListener),
```
jbr, line 13: This is an enum in order to avoid binding to a string spec until .listen is called.
```rust
let mut multi = MultiListener::new();
multi.add("http://localhost:3000");
//at this point, 3000 should not be bound
app.listen(multi) // but now it is
```

src/listener/to_listener.rs
```diff
@@ -0,0 +1,124 @@
...
+impl<State: 'static + Send + Sync> ToListener<State> for std::net::TcpListener {
+    fn to_listener(self) -> Box<dyn Listener<State>> {
+        Box::new(TcpListener::from_listener(self))
+    }
+}
+
+impl<State: 'static + Send + Sync> ToListener<State> for (&str, u16) {
+    fn to_listener(self) -> Box<dyn Listener<State>> {
+        Box::new(TcpListener::from_addrs(
+            async_std::task::block_on(self.to_socket_addrs())
```
jbr, line 36: There doesn't seem to be a way to make an async_std::net::SocketAddr from a std::net::SocketAddr, which would be nice to be able to avoid block_on here

src/listener/tcp_listener.rs
```diff
@@ -0,0 +1,115 @@
...
+    FromAddrs(Vec<SocketAddr>, Option<net::TcpListener>),
+}
+
+impl TcpListener {
+    pub fn from_addrs(addrs: Vec<SocketAddr>) -> Self {
```
Fishrock123, line 20: Should this take `ToSocketAddrs`?

src/listener/to_listener.rs
```diff
@@ -0,0 +1,401 @@
...
+}
+
+impl<State: Send + Sync + 'static> ToListener<State> for std::net::SocketAddr {
+    type Listener = TcpListener;
+    fn to_listener(self) -> io::Result<Self::Listener> {
+        Ok(TcpListener::from_addrs(vec![self]))
+    }
+}
+
+impl<TL: ToListener<State>, State: Send + Sync + 'static> ToListener<State> for Vec<TL> {
```
Fishrock123, line 230: Could probably be implemented for `IntoIterator<TL>`?

src/listener/to_listener.rs
```diff
@@ -0,0 +1,401 @@
...
+    type Listener = Self;
+    fn to_listener(self) -> io::Result<Self::Listener> {
+        Ok(self)
+    }
+}
+
+impl<State: Send + Sync + 'static> ToListener<State> for std::net::SocketAddr {
+    type Listener = TcpListener;
+    fn to_listener(self) -> io::Result<Self::Listener> {
+        Ok(TcpListener::from_addrs(vec![self]))
```
Fishrock123, line 226: Should `TcpListener` have `from_addr` for singles?

Fishrock123:
Overall this seems like a good direction to me.

src/listener/tcp_listener.rs
jbr: I guess it could, but this isn't public outside of tide and there are already ToListeners for socket addr conversion

src/listener/to_listener.rs
jbr: I think it can't be, because Vec is a concrete type and IntoIterator is a trait. I tried that

src/listener/to_listener.rs
jbr: It could, but what problem does that solve? I think this is the only place it's constructed with just one SocketAddr

src/listener/to_listener.rs
jbr: But also what iterators are people going to pass into Server::listen other than a Vec?

src/listener/tcp_listener.rs
jbr: Oh right, I remember now. If we have a Vec<SocketAddr> and we want to pass it to something that takes a ToSocketAddrs, we have to pass a &[SocketAddr] slice which then gets cloned into a new Vec. Taking a Vec avoids that

tests/unix.rs
```diff
@@ -17,30 +17,25 @@ mod unix_tests {
...
-                app.listen_unix(sock_path).await?;
+                app.listen(sock_path).await?;
                 http_types::Result::Ok(())
             });
 
             let client = task::spawn(async move {
                 task::sleep(Duration::from_millis(100)).await;
                 let listener = UnixStream::connect(&sock_path_for_client).await?;
-                let req = Request::new(Method::Get, Url::parse("unix://local.socket/").unwrap());
+                let req = Request::new(Method::Get, Url::parse("http://local.socket/").unwrap());
```
yoshuawuyts, line 29:
```suggestion
                let req = Request::new(Method::Get, Url::parse("http+unix://local.socket/").unwrap());
```

src/listener/to_listener.rs
```diff
@@ -0,0 +1,405 @@
...
+    /// unsuccessful bind attempt.
+    fn to_listener(self) -> io::Result<Self::Listener>;
+}
+
+impl<State: Send + Sync + 'static> ToListener<State> for Url {
+    type Listener = ParsedListener;
+
+    fn to_listener(self) -> io::Result<Self::Listener> {
+        match self.scheme() {
+            "unix" | "file" | "http+unix" => {
```
yoshuawuyts, line 72:
```suggestion
            "http+unix" => {
```

src/listener/tcp_listener.rs
```diff
@@ -0,0 +1,116 @@
...
+
+use async_std::net::{self, SocketAddr, TcpStream};
+use async_std::prelude::*;
+use async_std::{io, task};
+
+#[derive(Debug)]
+pub enum TcpListener {
+    FromListener(net::TcpListener),
+    FromAddrs(Vec<SocketAddr>, Option<net::TcpListener>),
+}
```
yoshuawuyts, line 17: Hmm, I guess I'm somewhat confused by this structure. In the stdlib `ToSocketAddr` is resolved the moment it is created. Is there a practical reason we're not doing the same here?

That way we can store a single `async_std::net::TcpListener` that we then later listen on. The same goes for `unix_listener.rs`. `ToListener::to_listener` is fallible, so I don't believe it should be a problem.

src/listener/to_listener.rs
```diff
@@ -0,0 +1,405 @@
...
+                Url::parse(self)
+                    .map_err(|_| {
+                        io::Error::new(
+                            io::ErrorKind::InvalidInput,
+                            format!("unable to parse listener from `{}`", self),
+                        )
+                    })
+                    .and_then(ToListener::<State>::to_listener)
+            })
+    }
```
yoshuawuyts, line 137: This is quite hard to follow. Using `match` seems preferable.

src/listener/to_listener.rs
```diff
@@ -0,0 +1,405 @@
...
+        Ok(self)
+    }
+}
+
+impl<State: Send + Sync + 'static> ToListener<State> for ParsedListener {
+    type Listener = Self;
+    fn to_listener(self) -> io::Result<Self::Listener> {
+        Ok(self)
+    }
+}
```
yoshuawuyts, line 221: What does parsedlistener do?

src/listener/to_listener.rs
```diff
@@ -0,0 +1,405 @@
...
+pub trait ToListener<State: Send + Sync + 'static> {
+    type Listener: Listener<State>;
+    /// Transform self into a
+    /// [`Listener`](crate::listener::Listener). Unless self is
+    /// already bound/connected to the underlying io, converting to a
+    /// listener does not initiate a connection. An Err return
+    /// indicates an unsuccessful conversion to a listener, not an
+    /// unsuccessful bind attempt.
+    fn to_listener(self) -> io::Result<Self::Listener>;
+}
```
yoshuawuyts, line 65: One question that I have is: why don't we perform the listening logic inside of the `to_listener` function's body? Intuitively it feels like we may be having one layer of traits more than is required. We can still reuse logic by having an e.g. `listen_unix` function that knows how to listen on a unix socket, but that can be called from this function.

I guess what I'm trying to say is that I'm unclear why "convert into a listener" and "listen on said listener" are two separate steps. My guess is that it's probably coming from the desire to have MultiListener?

During our chat yesterday I showed how to manually `MultiListener`'s functionality using `task::spawn` and `future::join`. We agreed that doing this manually isn't ideal, so having an intermediate abstraction makes sense. But what I think that shows is that servers that have been started *can* in fact be combined after the fact.

I'm not entirely convinced that I've quite covered all the details here -- but I feel like there might be something to it. Would be keen to hear your thoughts on this!

src/listener/tcp_listener.rs
jbr: The difference is that ToSocketAddr doesn't immediately _bind_ to that address. In this version of the code, we do as much resolution as we can, but we avoid binding until listen is called. That way this doesn't exhibit confusing behavior:
```rust
let mut multi = MultiListener::new();
multi.add("localhost:8000")?;
multi.add("unix+http://file")?;
// are 8000 and ./file bound here? I wouldn't expect them to be yet

multi.listen().await?; //but now I would expect them to be.

File: src/listener/to_listener.rs (hunk 10 above)
Comment (jbr) [reply src/listener/to_listener.rs:137]: Totally agreed. I dislike this code a lot and went through a bunch of iterations of it. I'll see what I can do to improve it

File: src/listener/to_listener.rs (hunk 11 above)
Comment (jbr) [reply src/listener/to_listener.rs:221]: This was a simple approach to wrapping two specific listener types without having to make them dyn. We can only parse a limited number of types currently, so this is the associated Listener type that's returned from string parsing

File: src/listener/tcp_listener.rs (hunk 9 above)
Comment (Fishrock123) [reply src/listener/tcp_listener.rs:17]: I also would not expect resources by string to be opened until `listen` is called, unless someone passes in something that is already explicitly bound, i.e. a struct for a file descriptor or something.

File: src/listener/to_listener.rs (hunk 10 above)
Comment (Fishrock123) [reply src/listener/to_listener.rs:137]: I went to suggest that but I wasn't sure since the match doesn't end up being much shorter... is probably still easier to read though. Maybe this could be clarified even more:

```rust
if let Some(socket_addrs) = self.to_socket_addrs()
    Ok(ParsedListener::Tcp(TcpListener::from_addrs(
        socket_addrs.collect(),
    )))
} else {
    Url::parse(self)
        .map_err(|_| {
            io::Error::new(
                io::ErrorKind::InvalidInput,
                format!("unable to parse listener from `{}`", self),
             )
         })
         .and_then(ToListener::<State>::to_listener)
}
```

examples/multi_listener.rs
```diff
@@ -0,0 +1,16 @@
...
+    let mut app = tide::new();
+
+    app.at("/").get(|request: tide::Request<_>| async move {
+        Ok(format!(
+            "Hi! You reached this app through: {}",
+            request.local_addr().unwrap_or("an unknown port")
+        ))
+    });
+
+    app.listen(vec!["localhost:8000", "localhost:8081"]).await?;
```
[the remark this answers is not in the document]
Fishrock123: Personally, that seems like a surprising behavior, `try_listen` seems like a better name for that but I suppose that could already be idiomatic in std rust.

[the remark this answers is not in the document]
jbr: I started writing an implementation of the fallback port multi-listener, but I realized I don't understand the use case. When ToSocketAddr resolves to multiple ports, those ports represent different notions of the same thing, so like binding to [::1]:8000 vs 127.0.0.1:8000 are both representing the notion of "localhost:8000" so a system could reliably provide "localhost:8000" and then access it from port 8000 by reverse proxying to that port, for example.  However, providing a vec of ["localhost:8000", "localhost:8001", "localhost:8002"] does not represent one resource, and I don't understand how the behavior of fallback ports would actually be used.  If 8000 is currently bound, isn't that a concerning state that should halt app startup?

In comparison, binding to multiple ports concurrently has several slightly-more-common use cases, such as supporting both uds for fast clients (local to the machine) also to tcp for network clients, or binding to both a public network interface as well as a private network interface, or potentially binding to both tcp://0.0.0.0:80 and tls://0.0.0.0:443 with appropriate permissions (authbind, etc).

src/listener/tcp_listener.rs
```diff
@@ -0,0 +1,116 @@
...
+
+impl TcpListener {
+    pub fn from_addrs(addrs: Vec<SocketAddr>) -> Self {
+        Self::FromAddrs(addrs, None)
+    }
+
+    pub fn from_listener(tcp_listener: impl Into<net::TcpListener>) -> Self {
```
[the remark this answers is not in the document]
jbr: Everything on listener::TcpListener is effectively pub(crate) because TcpListener itself is only pub(crate). Is there a reason to annotate each method as such as well?

src/listener/tcp_listener.rs
```diff
@@ -0,0 +1,116 @@
...
+use crate::{log, Server};
+
+use std::fmt::{self, Display, Formatter};
+
+use async_std::net::{self, SocketAddr, TcpStream};
+use async_std::prelude::*;
+use async_std::{io, task};
+
+#[derive(Debug)]
+pub enum TcpListener {
```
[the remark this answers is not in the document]
jbr: This only pub(crate) — would a doc comment actually get rendered, or is it more just like a traditional comment (which would still be useful here)

src/listener/to_listener.rs
jbr: I made an experimental branch at https://github.com/jbr/tide/tree/simpler-listener-trait for comparison. Docs haven't been updated, but: There's no longer a ToListener trait, just a [Listener](https://github.com/jbr/tide/blob/simpler-listener-trait/src/listener/to_listener.rs) trait. Calling `Listener::<State>::listen("http://localhost:8000").await` immediately binds, and Listener is implemented for all of the types that ToListener had been implemented for.  MultiListener also binds immediately, though, which feels like a step backwards.  I couldn't figure out how to get it to correctly hold the `Box<dyn Listener>`s. Listen needs to take self because [some listeners](https://github.com/jbr/tide/blob/simpler-listener-trait/src/listener/to_listener.rs#L154-L158) need to move self into the listener.

I still prefer the two-step version, but it's true that there's slightly less code in the one-trait version: https://github.com/jbr/tide/commit/e86e9575f119f2854f58486017364c8b9ffeb619

src/listener/to_listener.rs
jbr: I made this change, but as a user of tide, I'm personally likely to end up typing unix:/// and file:/// and getting error messages and having to fix them because so many other tools use those schemes. I generally figure if software knows what you mean, it should just do that. What if unix:// and file:// worked but also logged a warning?

src/listener/multi_listener.rs
```diff
@@ -0,0 +1,116 @@
...
+/// # if false {
+///        app.listen(multi).await?;
+/// # }
+///        Ok(())
+///    })
+///}
+///```
+
+#[derive(Default)]
+pub struct MultiListener<State>(Vec<Box<dyn Listener<State>>>);
```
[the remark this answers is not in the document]
jbr: addressed by renaming MultiListener to ConcurrentListener and adding FailoverListener. I don't understand the use case well enough to write good/representative docs or examples, though: https://github.com/http-rs/tide/pull/610/files#diff-92873655898e8cfdba457e044d652af1 Importantly, this sort of thing would be easy for apps to implement on their own, or third party crates, etc. I think the only tide-internal question is what Vec turns into

src/listener/parsed_listener.rs
```diff
@@ -0,0 +1,33 @@
+#[cfg(unix)]
+use super::UnixListener;
+use super::{Listener, TcpListener};
+use crate::{utils::BoxFuture, Server};
+use async_std::io;
+use std::fmt::{self, Display, Formatter};
+
+#[derive(Debug)]
+pub enum ParsedListener {
```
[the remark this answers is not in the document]
jbr: This is an enum that holds the two types that can be parsed from strings. If/when we add "tls" and "https+unix" schemes to the string parser, those listener types would be included here.

Cargo.toml
```diff
@@ -41,6 +41,7 @@ kv-log-macro = "1.0.4"
 serde = "1.0.102"
 serde_json = "1.0.41"
 route-recognizer = "0.2.0"
+futures = "0.3.5"
```
Fishrock123, line 44: Does this need to pull in all of `futures`?

yoshuawuyts: We can probably only input `futures-util` for this as well.

Fishrock123: Fwiw `futures-util` has [`FuturesUnordered`](https://docs.rs/futures-util/0.3.5/futures_util/stream/struct.FuturesUnordered.html).

jbr: addressed in rebase

PR #615, opened 2020-06-24
Don't panic on missing path param

vladan:
Resolves: #324

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -4,7 +4,7 @@ use route_recognizer::Params;
 
 use std::ops::Index;
 use std::pin::Pin;
-use std::{str::FromStr, sync::Arc};
+use std::{fmt, str::FromStr, sync::Arc};
 
 use crate::cookies::CookieData;
 use crate::http::cookies::Cookie;
@@ -25,6 +25,23 @@ pub struct Request<State> {
     pub(crate) req: http::Request,
     pub(crate) route_params: Vec<Params>,
 }
+
+#[derive(Debug)]
+pub enum ParamError<E> {
+    NotFound(String),
+    ParsingError(E),
+}
+
+impl<E: fmt::Debug + fmt::Display> fmt::Display for ParamError<E> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            ParamError::NotFound(name) => write!(f, "Param \"{}\" not found!", name),
+            ParamError::ParsingError(err) => write!(f, "Param failed to parse: {}", err),
+        }
+    }
+}
+
+impl<T: fmt::Debug + fmt::Display> std::error::Error for ParamError<T> {}
 
 impl<State> Request<State> {
     /// Create a new `Request`.
@@ -265,19 +282,38 @@ impl<State> Request<State> {
     ///
     /// # Errors
     ///
-    /// Yields an `Err` if the parameter was found but failed to parse as an
+    /// Yields a `ParamError::ParsingError` if the parameter was found but failed to parse as an
     /// instance of type `T`.
     ///
-    /// # Panics
+    /// Yields a `ParamError::NotFound` if `key` is not a parameter for the route.
     ///
-    /// Panic if `key` is not a parameter for the route.
-    pub fn param<T: FromStr>(&self, key: &str) -> Result<T, T::Err> {
+    /// # Examples
+    ///
+    /// ```no_run
+    /// # use async_std::task::block_on;
+    /// # fn main() -> Result<(), std::io::Error> { block_on(async {
+    /// #
+    /// use tide::{Request, Result};
+    ///
+    /// async fn greet(req: Request<()>) -> Result<String> {
+    ///     let name = req.param("name").unwrap_or("world".to_owned());
+    ///     Ok(format!("Hello, {}!", name))
+    /// }
+    ///
+    /// let mut app = tide::new();
+    /// app.at("/hello").get(greet);
+    /// app.at("/hello/:name").get(greet);
+    /// app.listen("127.0.0.1:8080").await?;
+    /// #
+    /// # Ok(()) })}
+    /// ```
+    pub fn param<T: FromStr>(&self, key: &str) -> Result<T, ParamError<T::Err>> {
         self.route_params
             .iter()
             .rev()
             .find_map(|params| params.find(key))
-            .unwrap()
-            .parse()
+            .ok_or_else(|| ParamError::NotFound(key.to_string()))
+            .and_then(|param| param.parse().map_err(ParamError::ParsingError))
     }
 
     /// Parse the URL query component into a struct, using [serde_qs](serde_qs). To get the entire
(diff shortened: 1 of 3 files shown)


src/request.rs
```diff
@@ -271,13 +297,13 @@ impl<State> Request<State> {
     /// # Panics
     ///
     /// Panic if `key` is not a parameter for the route.
-    pub fn param<T: FromStr>(&self, key: &str) -> Result<T, T::Err> {
+    pub fn param<T: FromStr>(&self, key: &str) -> Result<T, ParamError<T>> {
```
jbr, line 300: Does it work to make ParamError generic over `T::Err`, like `Result<T, ParamError<T::Err>>` and then simplify the bounds on `ParamError` to refer directly to the contained error?

examples/multipath.rs
```diff
@@ -0,0 +1,18 @@
+use async_std::task;
+use http_types::Result;
```
jbr, line 2: this should be `tide::Result`. Examples (and tide apps) should not need to directly depend on http_types

src/request.rs
```diff
@@ -26,6 +26,32 @@ pub struct Request<State> {
     pub(crate) route_params: Vec<Params>,
 }
 
+#[derive(Debug)]
+pub enum ParamError<T: FromStr> {
```
jbr, line 30: what is the `use` path to get to this from a tide app if someone needs to match on it? that probably needs to be specified in lib.rs

src/request.rs
vladan: Not AFAICT, but will try to trim down the bounds, looks a bit messy this way.

src/request.rs
vladan: Yes, I added it in lib.rs [here](https://github.com/http-rs/tide/pull/615/files#diff-b4aea3e418ccdb71239b96952d9cddb6R219), thanks!

src/request.rs
jbr: something like this is what i had in mind:
```diff
diff --git a/src/request.rs b/src/request.rs
index fe8e16c..7950a28 100644
--- a/src/request.rs
+++ b/src/request.rs
@@ -27,15 +27,14 @@ pub struct Request<State> {
 }
 
 #[derive(Debug)]
-pub enum ParamError<T: FromStr> {
+pub enum ParamError<T> {
     NotFound(String),
-    ParsingError(T::Err),
+    ParsingError(T),
 }
 
 impl<T> std::fmt::Display for ParamError<T>
 where
-    T: FromStr + std::fmt::Debug,
-    <T as std::str::FromStr>::Err: std::fmt::Debug + std::fmt::Display,
+    T: std::fmt::Debug + std::fmt::Display,
 {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
@@ -45,12 +44,7 @@ where
     }
 }
 
-impl<T> std::error::Error for ParamError<T>
-where
-    T: FromStr + std::fmt::Debug,
-    <T as std::str::FromStr>::Err: std::fmt::Debug + std::fmt::Display,
-{
-}
+impl<T> std::error::Error for ParamError<T> where T: std::fmt::Debug + std::fmt::Display {}
 
 impl<State> Request<State> {
     /// Create a new `Request`.
@@ -297,7 +291,7 @@ impl<State> Request<State> {
     /// # Panics
     ///
     /// Panic if `key` is not a parameter for the route.
-    pub fn param<T: FromStr>(&self, key: &str) -> Result<T, ParamError<T>> {
+    pub fn param<T: FromStr>(&self, key: &str) -> Result<T, ParamError<T::Err>> {
         self.route_params
             .iter()
             .rev()
```

vladan: Great! Thanks for the quick reply!!

vladan:
> Thanks for this! I'd also love to see a test that distinguishes between the two error return types and asserts on their contents. Looks good +1

I get an empty body in the response:
https://github.com/http-rs/tide/pull/615/files#diff-921be2a94c39ec66dbc3b3e4f8dc2ae3R104
Is there a way to read the error(s) from the response somehow? 
... or should I do a simple unit test on `param` only? Should be easier and straightforward. Also, should it be placed in `tests/request.rs`, or somewhere else?

src/request.rs
vladan: ... and thanks for filling in for my dead brain cells :) it's 4am, need to go to bed.
Thanks again for the help and quick responses!

tests/params.rs
```diff
@@ -0,0 +1,69 @@
+use http_types::{self, Method, Url};
+use tide::{self, Request, Response, Result, StatusCode};
+
+#[async_std::test]
+async fn test_param_invalid_type() {
+    async fn get_by_id(req: Request<()>) -> Result<Response> {
+        assert_eq!(
+            format!("{}", req.param::<i32>("id").err().unwrap()),
```
jbr, line 8: This might read easier
```suggestion
            req.param::<i32>("id").unwrap_err().to_string(),
```

tests/params.rs
```diff
@@ -0,0 +1,69 @@
...
+    );
+    let res: http_types::Response = server.respond(req).await.unwrap();
+    assert_eq!(res.status(), StatusCode::InternalServerError);
+}
+
+#[async_std::test]
+async fn test_missing_param() {
+    async fn greet(req: Request<()>) -> Result<Response> {
+        assert_eq!(
+            format!("{}", req.param::<String>("name").err().unwrap()),
```
jbr, line 29:
```suggestion
            req.param::<String>("name").unwrap_err().to_string(),
```

examples/multipath.rs
```diff
@@ -0,0 +1,17 @@
...
+
+fn main() -> Result<()> {
+    task::block_on(async {
+        let mut app = tide::new();
+        app.at("/hello").get(greet);
+        app.at("/hello/:name").get(greet);
+        app.listen("127.0.0.1:8080").await?;
+        Ok(())
+    })
+}
```
yoshuawuyts, line 17: Instead of adding this as an example could we add this as a doctest on `Request::param`?

vladan: Done! Thanks for the review!

vladan: Updated blank lines in the doc strings ... looks better now :)

src/request.rs
```diff
@@ -265,19 +282,38 @@ impl<State> Request<State> {
...
+    /// }
+    ///
+    /// let mut app = tide::new();
+    /// app.at("/hello").get(greet);
+    /// app.at("/hello/:name").get(greet);
+    /// app.listen("127.0.0.1:8080").await?;
+    /// #
+    /// # Ok(()) })}
+    /// ```
+    pub fn param<T: FromStr>(&self, key: &str) -> Result<T, ParamError<T::Err>> {
```
Fishrock123, line 310: Isn't this the same as saying `Result<Option<T>, T::Err>`?

jbr: Yeah, very similar and equivalent. See #324 for the initial discussion with both of those options proposed. Advantage of this is a single ? in the happy/normal path

vladan: Yes, what @jbr said :)
I went with this approach because unwrapping twice can be a bit tedious and IMHO the downside of introducing the ParamError enum is lower that the benefit of dealing with only one "effect".

Issue #604, opened 2020-06-17
serve_dir does not serve an index path


jbr:
In order to serve an index file, the following is currently required:

```rust
let static_dir = PathBuf::from("/some/path");
app.at("/")
    .get(|_| async { Ok(Body::from_file(static_dir.join("index.html"))?.into()) })
    .serve_dir(static_dir)?;
```

Many uses of serve_dir would want to attach an index route, so either this pattern should be documented, or serve_dir/ServeDir should support the definition of an index route/file.

bbigras:
@jbr Any idea how to avoid the error?

```
error[E0373]: closure may outlive the current function, but it borrows `static_dir`, which is owned by the current function
   --> src/main.rs:431:10
    |
431 |     .get(|_| async { Ok(Body::from_file(static_dir.join("index.html")).await?) })
    |          ^^^                            ---------- `static_dir` is borrowed here
    |          |
    |          may outlive borrowed value `static_dir`
    |
note: function requires argument type to outlive `'static`
   --> src/main.rs:430:1
    |
430 | / app.at("/")
431 | |     .get(|_| async { Ok(Body::from_file(static_dir.join("index.html")).await?) })
    | |_________________________________________________________________________________^
help: to force the closure to take ownership of `static_dir` (and any other referenced variables), use the `move` keyword
    |
431 |     .get(move |_| async { Ok(Body::from_file(static_dir.join("index.html")).await?) })
    |          ^^^^^^^^

error[E0505]: cannot move out of `static_dir` because it is borrowed
   --> src/main.rs:432:16
    |
430 | / app.at("/")
431 | |     .get(|_| async { Ok(Body::from_file(static_dir.join("index.html")).await?) })
    | |__________---____________________________----------______________________________- argument requires that `static_dir` is borrowed for `'static`
    |            |                              |
    |            |                              borrow occurs due to use in closure
    |            borrow of `static_dir` occurs here
432 |       .serve_dir(static_dir)?;
    |                  ^^^^^^^^^^ move out of `static_dir` occurs here
```

zelda-at-tempus:
I accomplished this by setting a separate handler for just the index file:
```rust
fn main()  {
    // ...
    app.at("/").get(index);
    app.at("/").serve_dir("static/").unwrap();
    // ...
}

async fn index(_req: tide::Request<AppState>) -> tide::Result<Response> {
    let mut res = Response::new(StatusCode::Ok);
[...]

PR #621, opened 2020-06-26
Route: add serve_index()

Fishrock123:
I chose this way of doing the "index" thing to avoid having to do an expensive directory iteration find in order to know if and which file to serve for an "index".

This makes the developer have to specify it up-front, ensuring that it actually goes to what they want without any extra processing.

Refs: https://github.com/http-rs/tide/issues/604

Note: i think this works but haven't tested it - I'd like to create an automated test but I also don't know how to get automated tests to have the correct static paths to text fixtures in rust/cargo tests...

### DIFF
--- a/src/route.rs
+++ b/src/route.rs
@@ -116,6 +116,8 @@ impl<'a, State: 'static> Route<'a, State> {
     /// Each file will be streamed from disk, and a mime type will be determined
     /// based on magic bytes.
     ///
+    /// Note: this does not serve "index" files. See [serve_index].
+    ///
     /// # Examples
     ///
     /// Serve the contents of the local directory `./public/images/*` from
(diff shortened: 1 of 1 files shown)


src/route.rs
```diff
@@ -138,6 +140,33 @@ impl<'a, State: 'static> Route<'a, State> {
...
+    /// ```no_run
+    /// #[async_std::main]
+    /// async fn main() -> Result<(), std::io::Error> {
+    ///     let mut app = tide::new();
+    ///     app.at("/public/images").serve_index("images/index.json")?;
+    ///     app.listen("127.0.0.1:8080").await?;
+    ///     Ok(())
+    /// }
+    /// ```
+    pub fn serve_index(&mut self, dir: impl AsRef<Path>) -> io::Result<()> {
```
jbr, line 162: Is `dir` intended to be a single file? It's confusing that it's named `dir` and we use ServeDir to resolve a single file instead of just `Ok(Body::from_file(index_path).await?)`

src/route.rs
```diff
@@ -138,6 +140,33 @@ impl<'a, State: 'static> Route<'a, State> {
         Ok(())
     }
 
+    /// Serve a index file statically.
```
jbr, line 143:
```suggestion
    /// Serve an index file statically.
```

src/route.rs
```diff
@@ -138,6 +140,33 @@ impl<'a, State: 'static> Route<'a, State> {
...
+    ///
+    /// The file will be streamed from disk, and a mime type will be determined
+    /// based on magic bytes.
```
jbr, line 146:
```suggestion
    /// based on the resolution defined in [Body::from_file](http_types::Body::from_file)
```

although that documentation should be improved as well.  Alternatively, this could explain that it's guessed based on the binary data _or_ the path extension for ascii data, but probably that logic should be documented in just one place.

Fishrock123:
Ah right for some reason I didn't think about subdirectories. 🤦 

I guess there's not much use trying to avoid directory iteration.

PR #624, opened 2020-06-29
handle tcp client disconnection

jbr:
## Description
This PR provides a way for the endpoint future to be dropped if the client disconnects

## Motivation and Context
Run this code on tide master
```rust
#[async_std::main]
    tide::log::start();
    let mut app = tide::new();
    app.at("/").get(|_| async {
        let mut i: i32 = 0;
        loop {
            i += 1;
            println!("{:?}", i);
            async_std::task::sleep(std::time::Duration::from_secs(1)).await;
        }

        #[allow(unreachable_code)]
        Ok("unreachable")
    });
    Ok(())
}
```
and `curl 127.0.0.1:8080` and then C-c.  The server will continue printing until it's stopped.

With this change, the server stops printing when the client disconnects.

Although this is a contrived example, this change allows potentially long-running operations to be canceled by early-disconnect.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -286,14 +286,15 @@ impl<State: Send + Sync + 'static> Server<State> {
         let local_addr = stream.local_addr().ok();
         let peer_addr = stream.peer_addr().ok();
         task::spawn(async move {
-            let result = async_h1::accept(stream, |mut req| async {
+            let fut = async_h1::accept(stream.clone(), |mut req| async {
                 req.set_local_addr(local_addr);
                 req.set_peer_addr(peer_addr);
                 self.respond(req).await
-            })
-            .await;
+            });
 
-            if let Err(error) = result {
+            let fut = fut.race(peek_disconnect(stream));
+
+            if let Err(error) = fut.await {
                 log::error!("async-h1 error", { error: error.to_string() });
             }
         });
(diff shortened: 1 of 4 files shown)


src/server.rs
```diff
@@ -491,6 +492,20 @@ impl<State: Sync + Send + 'static, InnerState: Sync + Send + 'static> Endpoint<S
     }
 }
 
+async fn peek_disconnect(stream: async_std::net::TcpStream) -> http_types::Result<()> {
+    let mut buf: Vec<u8> = vec![0; 1];
+    loop {
+        match stream.peek(&mut buf).await {
```
Fishrock123, line 498: How often does this happen? Should this be on a timer?

jbr: I initially thought so, but debugging indicates it only loops once (as `Ok(0)`) on my machine. I'm not entirely sure why, and this might be platform-dependent

jbr:
Closing this because it doesn't behave correctly for requests with bodies. This approach won't work because we need to wait for the actual body to finish reading before attempting to read in order to wait for 0 bytes

yoshuawuyts:
@jbr hmm, yeah something like this would be great. I'm not sure how well we can execute on this though; for this to behave as expected I think we would need [async drop](https://boats.gitlab.io/blog/post/poll-drop/) to land first.

I think perhaps a better approach for now would be to integrate something like stop tokens into tide so that we can trigger a safe shutdown procedure which will allow all existing connections to run to completion before finishing. This may require some noodling in `async-std` to get it right ([design proposal](https://github.com/yoshuawuyts/notes/blob/11ce15f29012ff9d3de5342357eb996805ec2660/rust/cancellation.md)) but perhaps we can use something [from userland](https://github.com/async-rs/stop-token) for this first.

jbr:
I started down another approach that I’ll revisit in a few days which wraps the async-std RW/io thing and wakes a future when it reads Ok(0).

The remaining blocker on that is something like Take/take on AsyncRead, but instead of only reading up to the limit, it returns two Read-ers: one that reads up to a limit and then returns 0/eof, and the other that is pending until the limit is reached, and accesses the contained Read after that. We can read the full request body and hand that to Body::from_reader and then elsewhere attempt to read one more byte from the second part, which will poll::pending until the connection closes

yoshuawuyts:
oh wait, just now realizing what I suggested is about something different. Hmm. Curious what you come up with!

PR #625, opened 2020-06-29
Send an empty response body to HEAD request

### LINKED ISSUE #623: Body still being sent for HEAD requests
#### Current Behavior

Whenever a HEAD request is made, if it's routed to a GET handler, a tide server will reply with the correct headers *but* it'll also serve the full body.

#### Code/Gist

I don't have a small reproduction test case right now, but hopefully it should be pretty easy to verify that this happens. I'm curious why the test introduced in #179 doesn't catch it - maybe because it uses a test server, not the real thing?

#### Expected behavior/code

I would expect tide to serve the correct headers (including `content-length`), but not the body.

As things stand, I can't implement HEAD correctly, because either I let the GET handler do its thing, and tide sends the whole body, or I add a wrapper that runs the GET handler and overrides the body to be empty - but `insert_header("content-length")` is always overwritten later by tide itself, so that would be an invalid HEAD response.

#### Environment

- tide version: 0.11.0
- Rust toolchain version(s): 1.44.1
- OS: Linux, Windows 10, any tbh

#### Possible Solution

I haven't looked too closely at the how responses are actually handled by tide, so I'm not sure.

#### Additional context/Screenshots

This is how curl behaves:

```shell
$ curl -X HEAD http://localhost:1111/ 2>&1 | head -20
Warning: Setting custom HTTP method to HEAD with -X/--request may not work the 
Warning: way you want. Consider using -I/--head instead.
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0<!DOCTYPE html>
<html lang="en">

<head>
    <title>
        
        fasterthanli.me
        
    </title>
    <meta http-equiv="content-type" content="text/html; charset=utf-8">
<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1">
<meta name="robots" content="noodp"/>

<link rel="stylesheet" href="/style.01ebxdp48thp88vghj60msqtx4.css">
<link rel="shortcut icon" href="/img/logo-round.png">
```

For comparison, `curl -X HEAD -v http://example.org` shows the headers, then hangs - because curl expects the body to be there.

Note that curl's `--head` or `-I` options *hide the problem*. They never read the body, whether the server sends it or not.

jbr:
## Description
HTTP HEAD requests [MUST NOT](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.4) return a response body

## Motivation and Context
Tide currently returns a response body, in violation of the spec, as reported in linked issue
Closes #623

## How Has This Been Tested?

Re-added the test mentioned in #623 and updated it for current tide api.

## Upstream changes:
This depends on http-rs/http-types#190

## Implementation concerns:
I'm not sure if this should be implemented as a default middleware and initially implemented it as one, but I cannot think of a situation in which tide users would want to break spec.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,3 +60,5 @@ required-features = ["unstable"]
 name = "router"
 harness = false
 
+[patch.crates-io]
+http-types = { git = "https://github.com/jbr/http-types", branch = "head-body" }
--- a/src/response.rs
+++ b/src/response.rs
@@ -306,6 +306,12 @@ impl AsMut<http::Headers> for Response {
     }
 }
 
+impl AsMut<Body> for Response {
+    fn as_mut(&mut self) -> &mut Body {
+        self.res.as_mut()
+    }
+}
+
 impl Into<http::Response> for Response {
     fn into(self) -> http_types::Response {
         self.res
--- a/src/server.rs
+++ b/src/server.rs
@@ -427,14 +427,8 @@ impl<State: Send + Sync + 'static> Server<State> {
             next_middleware: &middleware,
         };
 
-        match next.run(req).await {
-            Ok(value) => {
-                let res: http_types::Response = value.into();
-                // We assume that if an error was manually cast to a
-                // Response that we actually want to send the body to the
-                // client. At this point we don't scrub the message.
-                Ok(res.into())
-            }
+        let mut response: http_types::Response = match next.run(req).await {
+            Ok(value) => value.into(),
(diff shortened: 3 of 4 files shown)


jbr:
Closing because async-h1 hangs indefinitely, so this needs to be addressed there

PR #642, opened 2020-07-11
Server: add with_shared_state()

Fishrock123:
Allows true state sharing for multiple tide server applications without `Arc<Arc<State>>`.

See discord discussion: https://discordapp.com/channels/598880689856970762/649056551835009097/731205374832541746

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -191,6 +191,8 @@
 #![doc(html_favicon_url = "https://yoshuawuyts.com/assets/http-rs/favicon.ico")]
 #![doc(html_logo_url = "https://yoshuawuyts.com/assets/http-rs/logo-rounded.png")]
 
+use std::sync::Arc;
+
 mod cookies;
 mod endpoint;
 mod fs;
@@ -245,7 +247,7 @@ pub fn new() -> server::Server<()> {
     Server::new()
 }
 
-/// Create a new Tide server with shared application scoped state.
+/// Create a new Tide server with application scoped state.
 ///
 /// Application scoped state is useful for storing items
 ///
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
I can't open the discord link so I'm not sure what the motivation is — but guessing it seems like we want to have access to `Arc<State>` somehow?

Not opposed to always have require State to be `Arc`d when passed to the `with_state` constructor — or perhaps it's even better to require `State: Clone` which would provide the same guarantees in a more direct way.

Fishrock123:
Yeah, the idea is to be able to have more than one server share the same state without passing an `Arc`, which would become `Arc<Arc<>>` and then passed to middleware as an `Arc<State>`.

I think this was the intention to begin with but probably was mistaken?

Does `State: Clone` prevent us from having an `Arc` internally that is different from a user-passed `Arc`?

yoshuawuyts:
`State: Clone` indeed means we would no longer need to have an internal Arc since the state can be freely cloned already. `with_state(Arc::new(state))` would be valid.

One downside however is that depending on how State is constructed, `Arc` may show up in the Server's type signature — but I don't think this needs to be a problem.

Fishrock123:
I think that was part of the stated issue though, that it's annoying to have to deal with an `Arc<>` in the middleware when it's probably going to be anyways?

yoshuawuyts:
I mean in some cases it may very well be an `Arc<RwLock>` or perhaps only some fields need to be wrapped in an `Arc<Mutex>` for the whole struct to simply be `#[derive (Clone)]`. There are patterns we can introduce people to that extend to Rust as a whole.

I think `State: Clone` is probably the best way ahead here.

Fishrock123:
Uhhh, I'm not sure that is possible. `impl<State> Read for Request<State>` requires `DerefMut`...

Fishrock123:
So @jbr got the above to compile using `pin-project-lite`: https://gist.github.com/jbr/b0fd2964eb57a1d8d512e4d167ad6554

However I think that `Request` probably shouldn't hold the `State`. See https://github.com/http-rs/tide/issues/643

src/lib.rs
```diff
@@ -283,5 +285,47 @@ where
...
+/// #
+/// # Ok(()) }) }
+/// ```
+pub fn with_shared_state<State>(state: Arc<State>) -> server::Server<State>
+where
+    State: Send + Sync + 'static,
+{
+    Server::with_shared_state(state)
+}
+
```
yoshuawuyts, line 329: Now that `State: Clone` this method is no longer needed.

```suggestion
```

src/server.rs
```diff
@@ -167,7 +167,7 @@ impl Default for Server<()> {
 }
 
 impl<State: Send + Sync + 'static> Server<State> {
-    /// Create a new Tide server with shared application scoped state.
+    /// Create a new Tide server with application scoped state.
```
yoshuawuyts, line 170: Application state is and always was shared within the application. Now that `State: Clone` it's true both locally to applications, and between applications.

```suggestion
    /// Create a new Tide server shared with application scoped state.
```

Fishrock123:
Closing in favor of https://github.com/http-rs/tide/pull/644

PR #644, opened 2020-07-12
Server: require State to be Clone

Fishrock123:
Alternative to
https://github.com/http-rs/tide/pull/642

This approach is more flexible but requires the user ensure that their
state implements/derives `Clone`, or is wrapped in an `Arc`.

Co-authored-by: Jacob Rothstein <hi@jbr.me>

Notably while this has less new API surface, this requires more from user code.

### DIFF
--- a/examples/graphql.rs
+++ b/examples/graphql.rs
@@ -1,3 +1,5 @@
+use std::sync::Arc;
+
 use async_std::task;
 use juniper::{http::graphiql, http::GraphQLRequest, RootNode};
 use std::sync::RwLock;
@@ -37,8 +39,9 @@ impl NewUser {
     }
 }
 
+#[derive(Clone)]
 pub struct State {
-    users: RwLock<Vec<User>>,
+    users: Arc<RwLock<Vec<User>>>,
 }
 impl juniper::Context for State {}
 
@@ -96,7 +99,7 @@ async fn handle_graphiql(_: Request<State>) -> tide::Result<impl Into<Response>>
 fn main() -> std::io::Result<()> {
     task::block_on(async {
         let mut app = Server::with_state(State {
-            users: RwLock::new(Vec::new()),
+            users: Arc::new(RwLock::new(Vec::new())),
         });
         app.at("/").get(Redirect::permanent("/graphiql"));
         app.at("/graphql").post(handle_graphql);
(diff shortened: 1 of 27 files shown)


examples/graphql.rs
```diff
@@ -72,7 +74,7 @@ fn create_schema() -> Schema {
     Schema::new(QueryRoot {}, MutationRoot {})
 }
 
-async fn handle_graphql(mut request: Request<State>) -> tide::Result {
+async fn handle_graphql(mut request: Request<Arc<State>>) -> tide::Result {
```
jbr, line 77: If state (L40-42) is
```rust
#[derive(Clone)]
pub struct State {
    users: Arc<RwLock<Vec<User>>>
}
```, 
the endpoint signatures don't need to change

File: examples/middleware.rs [hunk 2]
```diff
@@ -25,8 +25,8 @@ impl UserDatabase {
 // application state. Because it depends on a specific request state,
 // it would likely be closely tied to a specific application
 fn user_loader<'a>(
```
jbr, line 27: similarly to the graphql example, UserDatabase could just have a derive(Clone) and nothing else would need to change

src/server.rs
```diff
@@ -190,8 +191,8 @@ impl<State: Send + Sync + 'static> Server<State> {
     /// };
     ///
     /// // Initialize the application with state.
-    /// let mut app = tide::with_state(state);
-    /// app.at("/").get(|req: Request<State>| async move {
+    /// let mut app = tide::with_state(Arc::new(state));
```
jbr, line 194: This State could also have a `derive(Clone)` and not need to change the request signature

Fishrock123:
I was on the edge about it, but i should have said in my OP: "Is non-Arc Clone" actually what a user would expect?"

I'm not sure it is personally? Like, your state will _not_ actually be shared between anything at all if it's not an `Arc` - it will only contain the data set originally and set within any specific request. That isn't "shared" state in any sense. If this is really for shared state I'm not honestly sure this is the correct direction...

yoshuawuyts:
> "Is non-Arc Clone" actually what a user would expect?"

I think so; this is not an uncommon thing to provide, and think it's entirely reasonable thing to allow people to implement. Certain performance sensitive applications may actually want this degree of control, and empowering them to implement this is useful.

Fishrock123:
Updated as per suggestions. 😄

Fishrock123:
Updated that example... i think there is a bit more work to do on the examples more generally outside of this PR.

Issue #643, opened 2020-07-12
Move `State` off of `Request`


Fishrock123:
It would be nice to not have `Request` have a generic parameter, for e.g. the `AsyncRead` trait which requires pinning. See https://github.com/http-rs/tide/pull/642#issuecomment-657141853 for where this was originally encountered.

However in interest of keeping middleware simple it would still be nice to allow `(Request, Next) -> Fut<Result<>>`, but for `State` it would probably ideal to have an extra parameter, kinda like Surf's `Client`, e.g. `(Request, State, Next) -> Fut<Result<>>`.

I was thinking maybe `server.middleware()` / endpoints could take `impl Into<Middleware<State>>` / endpoint, and so then that trait could be applied to both of the above with an extra function wrapper for the simple case, allowing both to work, and scoping `State` to the middleware/endpoint stack.

Silentdoer:
please consider actix-web design, it's state can be multi different type in a server.

Fishrock123:
@Silentdoer Could you elaborate more?

Fishrock123:
https://github.com/http-rs/tide/pull/645 has an impl conflict for some reason but is effectively what I'm proposing here.

fundon:
Actix uses a struct `Data<T>` to wrap a state `T`, so `T` can be different types.

```rust
async handler(a: Data<DB>, b: Data<Template>) ...
```

jbr:
I think there are likely performance implications of doing this, but ext and State could be the same thing potentially.

As it is, if you want multiple `State` types, you could set up a middleware that clones whatever state into each request's ext typemap.

Similarly, if State weren't a generic, it could be a starting value for each request's ext typemap, which then gets additional values inserted into it.

It seems like there might be a performance hit from making that change, but it might not be significant.

Silentdoer:
Ok, thanks for your patience.

yoshuawuyts:
@Fishrock123 This is not a direction I feel comfortable with; as per https://github.com/http-rs/tide/pull/642#issuecomment-657131458 in case we cannot set `State: Clone` we should make it so `App::with_state(Arc::new(state))`. But it seems like https://github.com/http-rs/tide/pull/644 enables `State: Clone`, so I don't understand what the problem is? It's normal to use `pin-project` in internals; that's entirely internal to the framework and abstracts away complexity from end-users.

Fishrock123:
[...]

PR #645, opened 2020-07-13
src: move State off of Request (WIP)

Fishrock123:
```
error[E0119]: conflicting implementations of trait `endpoint::Endpoint<_>`:
  --> src\endpoint.rs:70:1
   |
54 | / impl<State, F, Fut, Res> Endpoint<State> for F
55 | | where
56 | |     State: Send + Sync + 'static,
57 | |     F: Send + Sync + 'static + Fn(Request) -> Fut,
...  |
67 | |     }
68 | | }
   | |_- first implementation here
69 |
70 | / impl<State, F, Fut, Res> Endpoint<State> for F
71 | | where
72 | |     State: Send + Sync + 'static,
73 | |     F: Send + Sync + 'static + Fn(Request, State) -> Fut,
...  |
83 | |     }
84 | | }
   | |_^ conflicting implementation
```

Not sure I understand this error. `F` should be a distinct type for both??

### DIFF
--- a/src/cookies/middleware.rs
+++ b/src/cookies/middleware.rs
@@ -37,7 +37,7 @@ impl CookiesMiddleware {
 impl<State: Send + Sync + 'static> Middleware<State> for CookiesMiddleware {
     fn handle<'a>(
         &'a self,
-        mut ctx: Request<State>,
+        mut ctx: Request,
         next: Next<'a, State>,
     ) -> BoxFuture<'a, crate::Result> {
         Box::pin(async move {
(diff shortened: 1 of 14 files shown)


tirr-c:
I guess the compiler thinks `F` might implement both `Fn(Request) -> Fut` and `Fn(Request, State) -> Fut`. You can make such a type with `#![feature(fn_traits, unboxed_closures)]` by manually implementing `Fn` and its supertraits.

Fishrock123:
Err... this isn't solving the issue that #642 and #644 solve... this is just general ergonomics, imo? Likewise see https://github.com/http-rs/tide/issues/643#issuecomment-657621351

yoshuawuyts:
One interesting implication of this PR is that it would've made https://github.com/http-rs/tide/pull/670 easier to implement; no `State` means no `Default` either.

Issue #9, opened 2018-11-08
Session management


aturon:
Work out a good story for session management in Tide. No need to reinvent the wheel here: this can probably follow an approach taken by an existing framework, like actix-web.

skade:
Ideally, I'd prefer all the algorithmic parts of session handling (and strategies) implemented as libraries somewhere.

aturon:
Definitely! The whole idea with Tide is to keep as much in the library ecosystem as possible. In some cases we're starting with code here for convenience and then moving it out.

I'm not sure what's currently available in the ecosystem here that's not tied to an existing framework. If we need new general purpose crates, can add to the rust-net-web org.

bIgBV:
I wen through the following frameworks when understanding how they track state for a request during its lifecycle.

* [Rocket](https://rocket.rs)
* [Actix-web](https://actix-web.rx)
* [Tower-web](https://github.com/carllerche/tower-web)

I'm typing out what I learned about the different implementations with a recommendation of how to proceed with the same in Tide at the end.

## Rocket
Rocket only seems to support storing state in a request through the [`Cookie`] type. I think it is because they use a custom [`Request`] type. The cookies themselves are per request and are stored in a `RefCell` (though I'm not entirely sure why).

This approach seems a little inflexible as there is no state store across requests on the server side of things. You have to store information as a private cookie and retrieve it every time in order to persist state across requests.

## Tower-web
Tower web doesn't seem to have a user facing API for storing cookies or any kind of session for a given request. It does have a [`AnyMap`] for storing app wide config.

I think since tower-web is still under development, the API ins't decided yet.

## Actix-web
Out of the three, actix-web seems to have the most complete session storage implementation. It has a generic [`Session`] which provides the interface the session storage. The backend for this storage could be anything that implements the [`SessionBackend`] trait. The interface for accessing data from the backend is through the [`SessionImpl`] trait. By default, the `SessionBackend` trait this is implemented for the [`CookieSessionBackend`], which wraps over the [`CookieSession`] type which implements the `SessionImpl` trait. This is built on top of the [`cookie`] crate.

This is definitely the most flexible and complete solution out of the three frameworks I've looked into.

## Conclusion
[...]

PR #646, opened 2020-07-14
tide sessions

jbr:
this is an initial implementation of a session middleware in tide, using a somewhat modified async-sessions crate. this is implemented as a `sessions` feature, which is enabled by default.

this pr reuses some of the signed cookie behavior from the cookie crate without requiring the entire application use signed cookies.  this is to ensure that all users of sessions have signed cookies regardless of the rest of their application settings.

closes: #9 
blocked on: ~~http-rs/async-session#2~~ ~~Publishing async-session 2.0.0 (@yoshuawuyts)~~ @Fishrock123 clarifying change request
related: #541

as part of exploring this approach, I've fully implemented these session stores:
* [sqx+sqlite](https://github.com/jbr/async-sqlx-session)
* [redis](https://github.com/jbr/async-redis-session)
* [in-memory/ephemeral](https://github.com/jbr/async-session/blob/tide/src/mem.rs) (included in async-session)
* [cookie](https://github.com/jbr/async-session/blob/tide/src/cookie.rs) (included in async-session)

### DIFF
--- a/examples/sessions.rs
+++ b/examples/sessions.rs
@@ -1,0 +1,39 @@
+#[async_std::main]
+async fn main() -> Result<(), std::io::Error> {
+    tide::log::start();
+    let mut app = tide::new();
+
+    app.middleware(tide::sessions::SessionMiddleware::new(
+        tide::sessions::MemoryStore::new(),
+        std::env::var("TIDE_SECRET")
+            .expect(
+                "Please provide a TIDE_SECRET value of at \
+                      least 32 bytes in order to run this example",
+            )
+            .as_bytes(),
+    ));
+
+    app.middleware(tide::utils::Before(
+        |mut request: tide::Request<()>| async move {
+            let session = request.session_mut();
+            let visits: usize = session.get("visits").unwrap_or_default();
+            session.insert("visits", visits + 1).unwrap();
+            request
+        },
+    ));
+
+    app.at("/").get(|req: tide::Request<()>| async move {
+        let visits: usize = req.session().get("visits").unwrap();
+        Ok(format!("you have visited this website {} times", visits))
+    });
+
+    app.at("/reset")
+        .get(|mut req: tide::Request<()>| async move {
+            req.session_mut().destroy();
+            Ok(tide::Redirect::new("/"))
+        });
+
+    app.listen("127.0.0.1:8080").await?;
+
+    Ok(())
+}
--- a/src/request.rs
+++ b/src/request.rs
@@ -260,6 +260,12 @@ impl<State> Request<State> {
         self.req.ext().get()
     }
 
+    /// Get a mutable reference to value stored in request extensions.
+    #[must_use]
+    pub fn ext_mut<T: Send + Sync + 'static>(&mut self) -> Option<&mut T> {
+        self.req.ext_mut().get_mut()
+    }
+
     /// Set a request extension value.
     pub fn set_ext<T: Send + Sync + 'static>(&mut self, val: T) -> Option<T> {
         self.req.ext_mut().insert(val)
@@ -504,6 +510,32 @@ impl<State> Request<State> {
     pub fn cookie(&self, name: &str) -> Option<Cookie<'static>> {
         self.ext::<CookieData>()
             .and_then(|cookie_data| cookie_data.content.read().unwrap().get(name).cloned())
+    }
+
+    /// Retrieves a reference to the current session.
+    ///
+    /// # Panics
+    ///
+    /// This method will panic if a tide::sessions:SessionMiddleware has not
+    /// been run.
+    #[cfg(feature = "sessions")]
+    pub fn session(&self) -> &crate::sessions::Session {
+        self.ext::<crate::sessions::Session>().expect(
+            "request session not initialized, did you enable tide::sessions::SessionMiddleware?",
+        )
+    }
+
+    /// Retrieves a mutable reference to the current session.
+    ///
+    /// # Panics
+    ///
+    /// This method will panic if a tide::sessions:SessionMiddleware has not
+    /// been run.
+    #[cfg(feature = "sessions")]
+    pub fn session_mut(&mut self) -> &mut crate::sessions::Session {
+        self.ext_mut().expect(
+            "request session not initialized, did you enable tide::sessions::SessionMiddleware?",
+        )
     }
 
     /// Get the length of the body stream, if it has been set.
--- a/src/sessions/middleware.rs
+++ b/src/sessions/middleware.rs
@@ -1,0 +1,290 @@
+use super::{Session, SessionStore};
+use crate::http::{
+    cookies::{Cookie, Key, SameSite},
+    format_err,
+};
+use crate::{utils::async_trait, Middleware, Next, Request};
+use std::time::Duration;
+
+use async_session::{
+    base64,
+    hmac::{Hmac, Mac, NewMac},
+    sha2::Sha256,
+};
+
+const BASE64_DIGEST_LEN: usize = 44;
+
+/// # Middleware to enable sessions.
+/// See [sessions](crate::sessions) for an overview of tide's approach to sessions.
+///
+/// ## Example
+/// ```rust
+/// # async_std::task::block_on(async {
+/// let mut app = tide::new();
+///
+/// app.middleware(tide::sessions::SessionMiddleware::new(
+///     tide::sessions::MemoryStore::new(),
+///     b"we recommend you use std::env::var(\"TIDE_SECRET\").unwrap().as_bytes() instead of a fixed value"
+/// ));
+///
+/// app.middleware(tide::utils::Before(|mut request: tide::Request<()>| async move {
+///     let session = request.session_mut();
+///     let visits: usize = session.get("visits").unwrap_or_default();
+///     session.insert("visits", visits + 1).unwrap();
+///     request
+/// }));
+///
+/// app.at("/").get(|req: tide::Request<()>| async move {
+///     let visits: usize = req.session().get("visits").unwrap();
+///     Ok(format!("you have visited this website {} times", visits))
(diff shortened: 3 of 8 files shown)


examples/sessions.rs
```diff
@@ -0,0 +1,37 @@
+use tide::prelude::*;
+use tide::sessions::SessionMiddleware;
```
Fishrock123, line 2: This is bigger but I think we may want to reorganize middleware into `tide::middleware` similar to Surf.

Fishrock123:
We should probably also provide a `CookieSession` by default.

jbr:
I agree about CookieSessionStore, but think we maybe need secure cookies first? Not sure about that. I figured I'd implement an external datastore-backed cookie store first to see if all of the pieces are in place for that (they weren't — Session needs to be Deserialize as well as Serialize)

src/sessions/middleware.rs
```diff
@@ -0,0 +1,255 @@
...
+            .field("cookie_name", &self.cookie_name)
+            .field("session_ttl", &self.session_ttl)
+            .field("key", &"..")
+            .finish()
+    }
+}
+
+impl SessionMiddleware<MemoryStore> {
+    /// Builds a new SessionMiddleware with an ephemeral in-memory
+    /// store. See [`CookieStore`](crate::session::MemoryStore) for
```
aka-bash0r, line 103:
```suggestion
    /// store. See [`MemoryStore`](crate::session::MemoryStore) for
```

src/sessions/middleware.rs
```diff
@@ -0,0 +1,255 @@
...
+        if cookie_value.len() < BASE64_DIGEST_LEN {
+            return Err("length of value is <= BASE64_DIGEST_LEN");
+        }
+
+        // Split [MAC | original-value] into its two parts.
+        let (digest_str, value) = cookie_value.split_at(BASE64_DIGEST_LEN);
+        let digest = base64::decode(digest_str).map_err(|_| "bad base64 digest")?;
+
+        // Perform the verification.
+        let mut mac = Hmac::<Sha256>::new_varkey(&self.key.signing()).expect("good key");
```
Fishrock123, line 249: _`Hmac` can't actually fail as of the current time, but maybe it should at some future version? Dunno._
```suggestion
        let mut mac = Hmac::<Sha256>::new_varkey(&self.key.signing()).expect("Hmac key failure");
```

src/sessions/middleware.rs
```diff
@@ -0,0 +1,255 @@
...
+        if let Some(ttl) = self.session_ttl {
+            session.expire_in(ttl);
+        }
+
+        let secure_cookie = request.url().scheme() == "https";
+        request.set_ext(session.clone());
+
+        let mut response = next.run(request).await;
+
+        if let Some(cookie_value) = self.store.store_session(session).await {
```
Fishrock123, line 141: This is a bug (unless `Session` is in an `Arc`).

You are storing a pre-cloned session, anything set will be lost.

Above should be `request.set_ext(session);` and this should actually be a `Session` from `request.take_ext()` (does not yet exist).

src/sessions/middleware.rs
```diff
@@ -0,0 +1,255 @@
...
+
+        self.sign_cookie(&mut cookie);
+
+        cookie
+    }
+
+    /// Signs the cookie's value providing integrity and authenticity.
+    fn sign_cookie(&self, cookie: &mut Cookie<'_>) {
+        // Compute HMAC-SHA256 of the cookie's value.
+        let mut mac = Hmac::<Sha256>::new_varkey(&self.key.signing()).expect("good key");
```
Fishrock123, line 227: Ditto as the other `Hmac` `.expect()`.

src/sessions/middleware.rs
```diff
@@ -0,0 +1,255 @@
...
+        mac.update(cookie.value().as_bytes());
+
+        // Cookie's new value is [MAC | original-value].
+        let mut new_value = base64::encode(&mac.finalize().into_bytes());
```
Fishrock123, line 231:
```suggestion
        let mut combined = base64::encode(&mac.finalize().into_bytes());
```

src/request.rs
```diff
@@ -507,6 +507,23 @@ impl<State> Request<State> {
...
+        )
+    }
+
+    #[cfg(feature = "sessions")]
+    pub fn session_get(&self, key: impl AsRef<str>) -> Option<String> {
+        self.session().get(key.as_ref())
+    }
+
+    #[cfg(feature = "sessions")]
+    pub fn session_insert(&self, key: String, value: String) -> Option<String> {
```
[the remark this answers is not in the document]
Fishrock123: Also, shouldn't this be impossible without `ext_mut()`?

src/sessions/middleware.rs
jbr: this code block is verbatim from Cookie

src/sessions/middleware.rs
jbr: this is not true. look at the implementation of Session. it uses interior mutability around the data. the only data that might be lost is the expiry, and it is not currently  part of the intended api for applications to modify the expiry directly

src/sessions/middleware.rs
jbr: this is verbatim from Cookie

src/sessions/middleware.rs
jbr: this is verbatim from Cookie

src/sessions/middleware.rs
```diff
@@ -0,0 +1,255 @@
...
+    pub fn cookie_store(secret: &[u8]) -> Self {
+        Self::new(CookieStore::new(), secret)
+    }
+}
+
+#[async_trait]
+impl<Store, State> Middleware<State> for SessionMiddleware<Store>
+where
+    Store: SessionStore,
+    State: Send + Sync + 'static,
```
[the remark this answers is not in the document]
jbr: yeah, this branch needs to be rebased, but i'm going to wait until it feels done-ish and just squash this whole thing down into one commit

src/request.rs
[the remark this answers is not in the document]
jbr: interior mutability

examples/sessions.rs
```diff
@@ -0,0 +1,31 @@
+#[async_std::main]
+async fn main() -> Result<(), std::io::Error> {
+    tide::log::start();
+    let mut app = tide::new();
+
+    app.middleware(tide::sessions::SessionMiddleware::memory_store(
```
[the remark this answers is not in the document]
jbr: i don't think there's anything wrong with starting people out on the memory store, since its flaws are more obvious (losing data on server restart). we want to strongly encourage people not to use either of them, and i think the sled store will probably be the recommended choice for apps that don't already have an external db

src/sessions/middleware.rs
Fishrock123: Ugh hmm, that pattern is quite confusing, I'll have to give it a look again soon.

src/sessions/middleware.rs
jbr: marking similar pr review comments as resolved, since changes to this code should probably be introduced in the cookie crate. keeping this code as similar as possible to the upstream code will make audits and downstreaming patches easier

src/sessions/middleware.rs
jbr: Would really like to remove the interior mutability, but without being able to retrieve the Ext, there has to be an inner Arc<RwLock<_>> in which there's no need for a mutable outer reference. Marking as resolved, since this issue has been turned into [an http-types pr for discussion](https://github.com/http-rs/http-types/pull/201)

src/sessions/middleware.rs
```diff
@@ -0,0 +1,255 @@
...
+const BASE64_DIGEST_LEN: usize = 44;
+const KEY_LEN: usize = 32;
+
+/// # Tide Session Middleware
+///
+/// This middleware provides session support for tide, allowing small
+/// amounts of information to be securely attached to each browser
+/// session.
+///
+/// Tide's session implementation provides guest sessions, meaning
```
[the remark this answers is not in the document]
jbr: addressed, at least somewhat. there's now a "save_unchanged" config setting. if save_unchanged is disabled, we won't save a session until it's been changed from the Default::default()

src/request.rs
```diff
@@ -506,6 +511,30 @@ impl<State> Request<State> {
...
+
+    #[cfg(feature = "sessions")]
+    pub fn session_get<T: serde::de::DeserializeOwned>(&self, key: &str) -> Option<T> {
+        self.session().get(key)
+    }
+
+    #[cfg(feature = "sessions")]
+    pub fn session_insert(&mut self, key: &str, value: impl serde::Serialize) -> crate::Result<()> {
+        Ok(self.session_mut().insert(key, value)?)
+    }
```
yoshuawuyts, line 536: I'm wondering whether we should expose these methods. These methods are also directly available on the session object. Perhaps we could omit these for now and only add them back if we see they're frequently used? That way we don't add a lot of methods to `Request` all at once.

jbr: Totally agree. It feels a bit like a law of Demeter thing, but in theory ext has the same problem — we don't do .ext().get() and .ext_mut().insert(…), we do .ext() and .set_ext(…). Wasn't sure if I should stick with that convention or encourage people to chain off of session() and session_mut(). Relatedly, the mutability in insert and remove isn't strictly necessary because it's all interior mutation. I just added the mutability constraint so future changes that eliminate the interior mutability don't require a breaking change.

jbr:
@Fishrock123 Have your concerns been addressed? This is otherwise ready to merge, but you still have a change request on this

src/request.rs
jbr: Removed it for now, can always add it back in on a minor release

PR #647, opened 2020-07-15
Relax Sync bound on Fut required on sse::upgrade

rubendura:
## Description

Relax Sync bound on Fut required on sse::upgrade

## Motivation and Context
`async_std::spawn` doesn't seem to require a `Sync` bound on any `Futures`
used on it. `tide::sse::upgrade` will just `spawn` the `Future` returned by
the `Fn` parameter. With this in mind, the `Sync` bound on that returned
`Future` seems unnecessary.

This, in my experience, seems to conflict with any usages of `async-trait` within that future, since `async-trait` related futures don't appear to be Sync. The main README in the `async-trait` repo seems to suggest the same
> Async fns get transformed into methods that return Pin<Box<dyn Future + Send + 'async_trait>> and delegate to a private async freestanding function.

## How Has This Been Tested?
The code compiles and my project is happy. Not sure how to test this, but I'm open to suggestions!

## Types of changes

## Checklist:

### DIFF
--- a/src/sse/endpoint.rs
+++ b/src/sse/endpoint.rs
@@ -16,12 +16,11 @@ pub fn endpoint<F, Fut, State>(handler: F) -> SseEndpoint<F, Fut, State>
 where
     State: Send + Sync + 'static,
     F: Fn(Request<State>, Sender) -> Fut + Send + Sync + 'static,
-    Fut: Future<Output = Result<()>> + Send + Sync + 'static,
+    Fut: Future<Output = Result<()>> + Send + 'static,
 {
     SseEndpoint {
         handler: Arc::new(handler),
         __state: PhantomData,
-        __fut: PhantomData,
     }
 }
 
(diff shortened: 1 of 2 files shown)


rubendura:
I pushed an extra change making `sse::endpoint` accept a handler returning a non-`Sync` `Future`. This is mostly the same change I applied to `sse::upgrade`.

In order to do so I had to remove an apparently unnecessary `PhantomData` in `SseEndpoint` capturing the `Fut` type being made non-`Sync` in this last commit, such that `SseEndpoint` stays `Sync`, while the handler can return a non-`Sync` `Future` (since that'll only be `spawn`ed and that doesn't require the `Sync` bound.

There's still a `PhantomData<State>` in the `SseEndpoint` type which I don't fully understand why it's needed. Removing it makes the compiler complain that the `State` generic parameter is unused. However, that parameter is in fact used within the `F` type which is stored in the `SseEndpoint` type. Is this a compiler bug? (latest stable 1.45.0)
Also curiously removing the `PhantomData<State>` from the type makes the compiler _also_ complain about the `Fut` type being unused. Can someone shed some light on what's going on there?

PR #648, opened 2020-07-15
meta: fix the DCO by adding the text of DCO-1.1

Fishrock123:
The text was missing before. The DCO helps overcome some legal controversy regarding contribution authority for MIT in particular and perhaps also Apache-2.0. The Node.js Foundation (at the time) determined it was best to just stick the whole thing in the contributing document, which is what I am replicating here.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -55,9 +55,29 @@ that you *might* be wrong is critical for any successful open collaboration.
 
 Don't be a bad actor.
 
-## Developer Certificate of Origin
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
We should probably get rid of the DCO and ensure our Contributing.md file correctly references incoming license agreements. We already have this text as part of our [README](https://github.com/http-rs/tide#contribution), but repeating it in our contributing is probably good.

> Contribution

> Unless you explicitly state otherwise, any contribution intentionally submitted for inclusion in the work by you, as defined in the Apache-2.0 license, shall be dual licensed as above, without any additional terms or conditions.

PR #653, opened 2020-07-20
src, utils: add TideState alias

Fishrock123:
This replaces `State: Clone + Send + Sync + 'static` uses with
`State: TideState`, a trait alias for the same thing.

The goal is to make it so that any future changes to these bounds
needn't be directly modified in every instance of the `State` generic,
both in Tide and also for external middleware.

I suppose this maybe could be called `ThreadSafeClone` but that isn't as flexible for the future.
I do wonder if a `ThreadSafe` wouldn't be nice for all the other internal `Send + Sync + 'static` though.

### DIFF
--- a/examples/middleware.rs
+++ b/examples/middleware.rs
@@ -4,7 +4,7 @@ use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
 
 use tide::http::mime;
-use tide::utils::{After, Before};
+use tide::utils::{After, Before, TideState};
 use tide::{Middleware, Next, Request, Response, Result, StatusCode};
 
 #[derive(Debug)]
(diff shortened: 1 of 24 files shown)


yoshuawuyts:
It seems this PR proposed trading off ease of refactoring for readability. Seeing "TideState" as a trait bound by itself is pretty hard to follow, and I know it would certainly confuse me.

Instead I think if we want to make changes to the trade bounds we can either use refactor tools (which I believe are underway in rust-analyzer), or manually run a replacement query for all instances of `State: <traits>`.

In the further future the lang team is working to address this problem by making it so when a set of traits is declared as a bound on a struct, those traits don't need to be repeated in the `impl` blocks params. In our case that would remove most of the roughness.

PR #658, opened 2020-07-23
Reorder deps in cargo.toml

yoshuawuyts:
Moves `logtest` to dev deps and alphabetizes our dependencies. Thanks!


jbr:
Is there a tool that does this automatically, like yarn does? I'm surprised cargo-edit doesn't alphabetize

Issue #476, opened 2020-04-26
not possible to remove standard middleware when using Server::with_state()


leonbreedt:
In my case, I use `Server::with_state()`, which means I get the standard `LogMiddleware`.

However, I have a custom logging middleware to add more context for request logging (method, path, etc as structured arguments instead of a formatted string).

So I get doubled up log entries.

It feels like deciding to have state (DB pool in my case) and getting default middleware should be separate decisions, like `with_logging()`, `with_cookies()`?

vladan:
Hi @leonbreedt, I've submitted a PR for this here https://github.com/http-rs/tide/pull/468, but like your proposition for `with_logging()` and `with_cookies()` better.
IMHO it would be good to have them completely out of the server creation and use them as a builder `Server::with_state(...).with_logger().with_cookies()`, or maybe even:
```
    let app = tide::new()
	.with_state(...)
	.with_logging()
	.with_cookies()
```

vladan:
Hi @leonbreedt, I've submitted a PR for another issue regarding the default logging / cookies in nested servers here https://github.com/http-rs/tide/pull/468, but like your proposition for `with_logging()` and `with_cookies()` better :)

IMHO it would be good to have them completely out of the server creation and use them as a builder `Server::with_state(...).with_logger().with_cookies()`, or maybe even:
```
    let app = tide::new()
	.with_state(...)
	.with_logging()
	.with_cookies()
```
which would solve both of our problems :)

Issue #548, opened 2020-05-26
Consider having feature flags for all built-in middleware.


Fishrock123:
It's not inconceivable that people may want to disable any particularly built in middleware. [Particularly the logger middleware](https://github.com/http-rs/tide/issues/476), but it's not inconceivable that someone may want to disable even the cookies one for some reason.

These would still be on by default.

jbr:
It seems like this could be achieved without compiler feature flags, in code:
```
let mut app = tide::configure().skip_logger().with_state(state).build()
```
since it doesn't seem terribly unlikely tide will need some general way to optionally configure other aspects of the Server in code at some point

bluk:
I would prefer an explicit way to specify that no middleware should be added (versus having to discover any middleware added in every release and finding the flags/code to disable). The defaults can have appropriate middleware, but the idea is that it is easier to upgrade applications with new Tide versions if I can say I always want a Server without any default middleware and then I add any middleware which "works for me".

jbr:
Another relevant thing to this conversation: Would anyone actually want to disable cookies? When we talk about "disable all middleware," that currently would include breaking cookie parsing behavior

bluk:
If I have an API server which does not use cookies, I may not want to have any parsing done for cookies.

Also, if I have a custom session middleware, my concern would be with other middleware potentially interfering with the cookies set. For instance, I want to have private/signed cookies and while there's a PR for that today (which is awesome), there may be requirements which are not met by the API. Then, I would have to always be careful about the cookies or other middleware changing and conflicting with my middleware.

I think having cookie functionality and other default behavior like logging by default is great. Still, as much as I dislike having to choose middleware features (like Express) and just want all the features known to work together (like Rails), I think there are valid use cases where optional functionality is not desired, especially as the framework matures and apps know what they really want.

tirr-c:
I'd also prefer having none of the middleware applied by default, especially in the nested app scenario, where default middleware would be applied multiple times. Tide could provide some convenience methods to apply middleware that is used often, or let users apply what they need using `Server::middleware`.

Fishrock123:
That's a good point, mounting apps within apps having default middleware issues...

jbr:
I think the middlewares-running-multiple-times issue could be solved by allowing a middleware to specify that it should only be run once per request and tracking that info on the request. That would also allow for re-entrant routing

Tonyce:
I think we could remove the build-in middleware,
```rust
...
[...]

PR #660, opened 2020-07-25
lib, Server: move defaults to default()

Fishrock123:
This moves the default enabled built-in middleware to `tide::default()`,
and makes `tide::new()` return a plain `Server`.

Refs: https://github.com/http-rs/tide/issues/548
Fixes: https://github.com/http-rs/tide/issues/476

Or at least, this is one possible solution...

Alternatively: `new()` and `new_blank()` or something?

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ async-std = { version = "1.6.0", features = ["attributes"] }
 #[async_std::main]
 async fn main() -> Result<(), std::io::Error> {
     tide::log::start();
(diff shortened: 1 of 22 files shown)


Fishrock123:
Apparently, according to the rust discord, folks feel this is not a very good way of naming things. I'm not really sure what to do with that because this makes perfect sense to me.

Fishrock123:
Prefer https://github.com/http-rs/tide/pull/661

Issue #551, opened 2020-05-27
Tide logging issue with nested routes — needs investigation


jbr:

eribol:
This is my nested route:
`app.at("/school").nest({
            use crate::school::school::{school, school_detail, timetable, teachers};
            let mut api = tide::with_state(state.clone());
            api.at("").all(school);
            api.at("/:school").all(school_detail);
            api.at("/:school/timetable").all(timetable);
            api.at("/:school/teachers").all(teachers);
            api
        });
`
and i am requesting to "school/720917/teachers" and logging is:

> {"level":30,"time":1590551990049,"msg":"<-- Request received","method":"GET","path":"/school/720917/teachers"}
{"level":30,"time":1590551990049,"msg":"<-- Request received","method":"GET","path":"/720917/teachers"}
bura7
bura1
bura5
{"level":30,"time":1590551990056,"msg":"--> Response sent","method":"GET","path":"/720917/teachers","status":200,"duration":"6.599095ms"}
{"level":30,"time":1590551990056,"msg":"--> Response sent","method":"GET","path":"/school/720917/teachers","status":200,"duration":"6.682995ms"}
>
but its saying that response of requesting to "/720917/teachers" is okey. And when i tried requesting same route log says:
>{"level":30,"time":1590552199144,"msg":"<-- Request received","method":"GET","path":"/720917/teachers"}
{"level":40,"time":1590552199144,"msg":"--> Response sent","method":"GET","path":"/720917/teachers","status":404,"duration":"10.105µs"}
>
Which its true because there is no route with "/720917/teachers". But logging says its okey. It should not requesting but if it was about requesting, it should response 404. 

Sorry for bad language, not native speaker here.

yoshuawuyts:
Related: https://github.com/http-rs/tide/issues/396

PR #662, opened 2020-07-26
Only run the log middleware once per request

jbr:
Currently, nested applications like
```rust
    tide::log::start();
    let mut app = tide::new();
    app.at("/").nest({
        let mut api = tide::new();
        api.at("/").get(|_| async { Ok("Hello, world") });
        api
    });
    Ok(())
}
```
will log twice. This provides a simple mechanism to ensure that a particular middleware is only run once per request. This approach also supports reentrant routing. This approach also requires no sweeping changes to the tide middleware system to ensure only-once middleware behavior, and also allows the middleware to define the desired behavior if it has already run once for the request.  For example, a more complex logging middleware might record the time spent in each nested application but only emit that aggregated information once per request

## Description
Introduces a private unit struct which is added to request extensions.

## How Has This Been Tested?
added a test in tests/log.rs that uses logtest to assert that only one request received log entry is emitted, and only one response sent message is emitted per request/response cycle, even with a nested app. this test previously failed

## Types of changes

## Checklist:

closes: #551

This same pattern should be applied to the cookie middleware as well if this is merged

### DIFF
--- a/src/log/middleware.rs
+++ b/src/log/middleware.rs
@@ -3,7 +3,9 @@ use crate::{Middleware, Next, Request};
 
 /// Log all incoming requests and responses.
 ///
-/// This middleware is enabled by default in Tide.
+/// This middleware is enabled by default in Tide. In the case of
+/// nested applications, this middleware will only run once for each
+/// request.
 ///
 /// # Examples
 ///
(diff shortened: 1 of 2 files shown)


jbr:
@Fishrock123 I started to do this for the cookies middleware but couldn't figure out how to write a failing test. I think the issue is solved already for the cookies middleware because it reuses CookieData on request.ext and drains the cookie events on the response, so it can be run any number of times

PR #677, opened 2020-08-07
Fix --no-default-features

c410-f3r:
`cargo build/check/test --no-default-features` is broken and a CI step was included to avoid the same problem in the future

### DIFF
--- a/src/listener/tcp_listener.rs
+++ b/src/listener/tcp_listener.rs
@@ -1,11 +1,11 @@
 use super::is_transient_error;
 
 use crate::listener::Listener;
-use crate::{log, Server};
+use crate::Server;
 
 use std::fmt::{self, Display, Formatter};
 
-use async_std::net::{self, SocketAddr, TcpStream};
+use async_std::net::{self, SocketAddr};
 use async_std::prelude::*;
 use async_std::{io, task};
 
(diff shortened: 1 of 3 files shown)


src/listener/tcp_listener.rs
```diff
@@ -87,8 +91,9 @@ impl<State: Clone + Send + Sync + 'static> Listener<State> for TcpListener {
                     continue;
                 }
 
-                Ok(stream) => {
-                    handle_tcp(app.clone(), stream);
+                Ok(_stream) => {
+                    #[cfg(feature = "h1-server")]
+                    handle_tcp(_app.clone(), _stream);
```
c410-f3r, line 96: This is probably not the right solution. Any better ideas?

jbr: I think there's no reason to include any of the Listener implementations unless h1-server is provided

jbr: Addressed this in #678

jbr:
Closing in preference to #678, which makes more drastic removals for --no-default-features.  Thanks for raising this issue!

c410-f3r:
This closure was unexpected. The work was indented to fix the problem and if the the current approach isn't desired, then you guys could at least hint the appropriated direction.

jbr:
I'm sorry! I didn't want to make more work for you by saying "please re-do this pr." Fixing --no-default-features required making a lot of listener code conditional (everything except the traits)

c410-f3r:
@jbr Dont worry. Thanks for the fix 👍

PR #682, opened 2020-08-14
add more example application

azzamsa:
## Motivation and Context

To make easy for tide users adopt the framework, by using examples

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,7 @@ team. Use at your own risk.</sup>
 * [tide-example](https://github.com/jbr/tide-example) (sqlx + askama)
 * [playground-tide-mongodb](https://github.com/yoshuawuyts/playground-tide-mongodb)
 * [tide-morth-example](https://github.com/No9/tide-morth-example/)
+* [realworld-tide](https://github.com/colinbankier/realworld-tide). Rust / Tide example app following realworld.io. (v0.6)
 
 # Contributing
 Want to join us? Check out our [The "Contributing" section of the


jbr:
Thanks for adding this! Is it worth mentioning that this is built on tide from quite a while ago and may not represent current best practices or apis?

azzamsa:
> represent current best practices or apis?

if that the requirements of tide project to be posted in "example" section.  Then I have no authoritative answer, since I'm not the author.


@colinbankier is.

jbr:
Fair point. Maybe it's worth keeping a tide version number next to each of these

azzamsa:
Oh, I see your point. Since tide doesn't reach `1.0` yet. So the API will likely change.

Okay, let's add the version number next to the project examples.

Fishrock123:
I don't think we should include out-of-date examples?

azzamsa:
related: https://github.com/http-rs/tide/pull/684

jbr:
I think @Fishrock123's point is worth discussing. What is the purpose of this list, and does adding an outdated example help us meet that goal?

My answer to the former is "to help people structure their tide applications," and I don't see how outdated examples help us meet that goal.  In a mature project I can imagine someone saying "we have to use the old version for various reasons, so having old examples are useful," but I can't think of any reason to build a new tide application with anything older than the most recent release.

@azzamsa What do you think? Who does this help?

azzamsa:
Agree with that. those old version should be in the community maintained list.

jbr:
@azzamsa Why? What's the motivation behind including it in any list, if it doesn't help people use tide

azzamsa:
I learn approach, capabilities, best practices, API from popular *-boilerplates. (* = flask, and other web frameworks).

I want to know how people structure tide in a big project (real-world app), see what it capable of, etc.

It's okay if it's not relevant here.

jbr:
I think it _is_ relevant here, it's just that tide is rapidly changing, and the structure of applications changes as a result of those changes.  In particular, the 0.6 example uses a totally different approach to error handling than current tide apps would use, and that has somewhat far-reaching effects on the structure of an application. I worry that if someone were new to tide, they'd look at this example and not notice that tide's error handling story has improved since 0.6

jbr:
We very much _need_ more examples with different approaches, and I think it would be valuable to the community to bring the realworld.io example up to date so that it can be featured in this list

azzamsa:
Yes, I see your point since the beginning. Tide doesn't reach 1.0 yet. So the API will likely to change fast. We don't want to put any *old* project representing old tide API. We don't want to make people assume that bad old API is the current tide API and make people upset.

azzamsa:
> and I think it would be valuable to the community to bring the realworld.io example up to date so that it can be featured in this list

sure. It will auto-magically happened when tide had a bigger community, some days later, I hope.

colinbankier:
Thanks for considering the example. We would of course like to keep it up to date - I've opened https://github.com/colinbankier/realworld-tide/issues/29.
Also in terms of tide best practices etc, always happy to have issues or PRs opened to help that repo serve as a good example of building a non-trivial tide app.

PR #691, opened 2020-08-31
log: display client error messages when possible as warnings

Fishrock123:
This should help demystify some cases when client errors are determined by parsing and propagated via `?`.

## Description

Makes the logger display an associated error message when this warning occurs, if possible.

Also clarified that a client error occured.

## Motivation and Context

For never users some cases of client errors being caused by e.g. body parsing and then propgated via `?` may be unobvious. This helps demystify that.

## How Has This Been Tested?

Manual curl to the json example. Output:
```
tide::log::middleware Client error --> Response sent
    message missing field `name` at line 1 column 2
    method POST
    path /submit
    status 422
    duration 73.9µs
```

## Checklist:

### DIFF
--- a/src/log/middleware.rs
+++ b/src/log/middleware.rs
@@ -65,12 +65,22 @@ impl LogMiddleware {
                 });
             }
         } else if status.is_client_error() {
-            log::warn!("--> Response sent", {
-                method: method,
-                path: path,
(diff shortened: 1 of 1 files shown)


jbr (approved):
:thumbsup:

PR #693, opened 2020-09-07
Make `femme` dependency optional

rossmacarthur:
## Description

Make the `femme` dependency optional, enabled via the `logger` feature.

## Motivation and Context

Allows you to use a different log crate and `femme` doesn't have to be installed.

## Types of changes

      Technically speaking this is a breaking change in the case where someone removed the `logger` feature and still used `tide::log::LevelFilter`.

## Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ rustdoc-args = ["--cfg", "feature=\"docs\""]
 [features]
 default = ["h1-server", "logger", "sessions"]
 h1-server = ["async-h1"]
-logger = []
+logger = ["femme"]
 docs = ["unstable"]
 sessions = ["async-session"]
 unstable = []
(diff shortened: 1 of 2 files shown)


rossmacarthur:
Might be better with the following released though: https://github.com/yoshuawuyts/kv-log-macro/pull/6

Cargo.toml
```diff
@@ -39,8 +39,9 @@ async-session  = { version = "2.0.0", optional = true }
 async-sse = "4.0.0"
 async-std = { version = "1.6.0", features = ["unstable"] }
 async-trait = "0.1.36"
-femme = "2.0.1"
+femme = { version = "2.0.1", optional = true }
 futures-util = "0.3.5"
+log = { version = "0.4.8", features = ["std"] }
```
Fishrock123, line 44: What's this for?

rossmacarthur: Currently the dependency tree looks like this
```
tide
   -- kv-log-macro
      -- log (features = ["kv_unstable"])
   -- femme
      -- log (features = ["std"])
```

However, `tide` actually relies on the `std` feature in the `log` crate being turned on. Otherwise you get the following error

```
 the trait `log::kv::value::ToValue` is not implemented for `std::string::String`
```

Ideal way to solve this is actually like this: https://github.com/yoshuawuyts/kv-log-macro/pull/7

Fishrock123:
merging as is for now, yosh will have to address  the kv-log-macro thing some other time.

PR #701, opened 2020-09-21
Tide testing trait

jbr:
This approach is mutually exclusive with #697 (although it happens to be rebased on it) and http-rs/surf#229.
Instead of pulling in surf::Methods, a tide application would do the following:

```rust
// add a dev dep on tide with the feature "testing", then
let app = tide::new();
use tide::testing::TestingExt;
let body = app.get("/path").recv_string().await?;
```

In order to use this to test tide itself, we add a dev dep to `.` with the testing feature enabled. I was surprised that this works.

This includes a line that should be removed when https://github.com/http-rs/surf/pull/236 is merged and published: https://github.com/http-rs/tide/compare/main...jbr:tide-testing-trait?expand=1#diff-80398c5faae3c069e4e6aa2ed11b28c0R52 — There's no reason that enabling the tide testing trait should need to pull in a client backend for surf, but until http-rs/surf#236 is merged, it's mandatory to do so.

This also provides a starting place for future testing utilities to exist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ logger = []
 docs = ["unstable"]
 sessions = ["async-session"]
 unstable = []
+testing = ["surf"]
 # DO NOT USE. Only exists to expose internals so they can be benchmarked.
 __internal__bench = []
 
@@ -47,7 +48,7 @@ pin-project-lite = "0.1.7"
 route-recognizer = "0.2.0"
 serde = "1.0.102"
 serde_json = "1.0.41"
-http-client = {version = "4.0.0", default-features = false }
+surf = { version = "2.0.0-alpha.5", default-features = false, optional = true, features = ["h1-client"] } # TODO(jbr): remove this feature when surf can build without default features
 
 [dev-dependencies]
 async-std = { version = "1.6.0", features = ["unstable", "attributes"] }
@@ -59,6 +60,7 @@ portpicker = "0.1.0"
 serde = { version = "1.0.102", features = ["derive"] }
 surf = { version = "2.0.0-alpha.5", default-features = false, features = ["h1-client"] }
 tempfile = "3.1.0"
+tide = { path = ".", features = ["testing"] }
 
 [[test]]
 name = "nested"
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -30,6 +30,7 @@ logger = []
 docs = ["unstable"]
 sessions = ["async-session"]
 unstable = []
+testing = ["surf"]
```
yoshuawuyts, line 33: This is somewhat unfortunate; in order to write tests an opt-in feature needs to be enabled. Testing is a fairly common capability, and ideally we could have it always be on.

I understand that this is because we always compile `surf`; but for example #697 doesn't have this issue, and I think it's something we should do better.

Cargo.toml
```diff
@@ -47,6 +48,7 @@ pin-project-lite = "0.1.7"
 route-recognizer = "0.2.0"
 serde = "1.0.102"
 serde_json = "1.0.41"
+surf = { version = "2.0.0-alpha.5", default-features = false, optional = true, features = ["h1-client"] } # TODO(jbr): remove this feature when surf can build without default features
```
yoshuawuyts, line 51: Tying `tide` to a specific version of `surf` is likely not ideal. Other proposals have managed to avoid this, and I think if we can we should do so.

src/testing.rs
```diff
@@ -0,0 +1,71 @@
...
+    /// Builds a `TRACE` request.
+    fn trace(&self, path: &str) -> RequestBuilder {
+        self.client().trace(path)
+    }
+}
+
+impl<State: Clone + Send + Sync + Unpin + 'static> TestingExt for Server<State> {
+    fn client(&self) -> Client {
+        let mut client = Client::with_http_client(Arc::new(self.clone()));
+        client.set_base_url(Url::parse("http://example.com").unwrap());
```
yoshuawuyts, line 61: We should provide a way to configure the base URL; I don't think we can assume that `http://example.com` will be sufficient for all applications.

yoshuawuyts (commented):
In https://github.com/http-rs/surf/pull/229#issuecomment-695781065 I covered some of the issues that I see with a trait-based testing approach, and both this PR and https://github.com/http-rs/surf/pull/229 run into those.

Perhaps we should chat about this during this week's meeting?

PR #700, opened 2020-09-20
Refactor duplicate code

mendelt:
## Description
I refactored away some duplicate code in how middleware is added to endpoints. There is still some duplicate code left between the `Route::method` and `'Route::all` but factoring this out will probably require some lambda magic and will not improve readablilty.

## Motivation and Context
A check to see if there was middleware present was done in the same way in several places. This small refactoring moves
that check into the code that adds middleware onto an endpoint. This reduces code duplication and improves readability

## How Has This Been Tested?
Middleware has decent test coverage. The tests that touch this piece of code still succeed. This code does not add any functionality so no additional tests should be neccesary.

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [X] I have read the [CONTRIBUTING](https://github.com/rustasync/tide/blob/master/.github/CONTRIBUTING.md) document.
- [ ] I have added tests to cover my changes.
- [X] All new and existing tests passed.

PR #702, opened 2020-09-21
Bugfix, route prefix was always set to false after calling nest

mendelt:
## Description
I found this method in route.rs:
```rust
    pub fn nest<InnerState>(&mut self, service: crate::Server<InnerState>) -> &mut Self
    where
        State: Clone + Send + Sync + 'static,
        InnerState: Clone + Send + Sync + 'static,
    {
        self.prefix = true;
        self.all(service);
        self.prefix = false;
        self
    }
```
I'm not sure if this is expected behavior but I assume that if self.prefix was set to true before calling `nest` it should stay true.

## How Has This Been Tested?
This is a trivial code change, all tests still pass, however I could try to add a regression test for this. Let me know if this is required

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)
- [X] Breaking change (fix or feature that would cause existing functionality to change)

This does introduce a small change in behavior but it only impacts code that uses the unstable `strip-prefix` method

## Checklist:
- [X] All new and existing tests passed.

### DIFF
--- a/src/route.rs
+++ b/src/route.rs
@@ -104,9 +104,12 @@ impl<'a, State: Clone + Send + Sync + 'static> Route<'a, State> {
         State: Clone + Send + Sync + 'static,
         InnerState: Clone + Send + Sync + 'static,
     {
+        let prefix = self.prefix;
+
         self.prefix = true;
         self.all(service);
-        self.prefix = false;
+        self.prefix = prefix;
+
         self
     }
 


yoshuawuyts:
A regression test would be great. Also if we could split https://github.com/http-rs/tide/pull/702/commits/3d41fb90679017fe8940cd57ddd395cf95c710d4 from https://github.com/http-rs/tide/pull/702/commits/2ad051a4ea718b77e0954a01b8ea53ccc17ff4ae into separate patches that'd be greatly appreciated. Also we've fixed the tests on main, which means https://github.com/http-rs/tide/pull/702/commits/3a25c8535d2d5399340ccbfeb9f45ab716f0dda9 is no longer needed.

Thanks heaps!

mendelt:
Thank you, the two branches for https://github.com/http-rs/tide/pull/700 and this pr are now both based on `main` so they don't depend on each other anymore. And I removed the test fixes from https://github.com/http-rs/tide/pull/700
I'll let you know when I get around to adding the regression test.

mendelt:
Hmm.. I looked at testing this, but don't see how to do that the way tests are set up. The prefix flags is not accessible from outside the Route struct and to test this we need to call the unstable `nest` function. Not sure how to proceed.

yoshuawuyts (approved):
It's okay if we can't test it; happy enough to merge this as-is. Thanks heaps!

PR #704, opened 2020-09-26
Lints

yoshuawuyts:
Adds various lints and fixes our implementations. This ensures items are always documented and we don't miss any in CI. Thanks!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -54,8 +54,8 @@
 
 #![cfg_attr(feature = "docs", feature(doc_cfg))]
 #![forbid(unsafe_code, rust_2018_idioms)]
-#![deny(future_incompatible, missing_debug_implementations, nonstandard_style)]
-#![warn(missing_docs, unreachable_pub)]
+#![deny(missing_debug_implementations, nonstandard_style)]
+#![warn(missing_docs, unreachable_pub, future_incompatible)]
 #![doc(html_favicon_url = "https://yoshuawuyts.com/assets/http-rs/favicon.ico")]
 #![doc(html_logo_url = "https://yoshuawuyts.com/assets/http-rs/logo-rounded.png")]
 
@@ -69,11 +69,8 @@ mod request;
 mod response;
 mod response_builder;
 mod route;
+mod router;
 mod server;
-
-// Expose the router for benchmarking only.
-#[doc(hidden)]
-pub mod router;
 
 pub mod convert;
 pub mod listener;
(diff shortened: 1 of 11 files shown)


yoshuawuyts:
I've disabled the router bench temporarily; we don't really use it much anymore, but can always spin it back up if we want to.

yoshuawuyts:
And I've restored the router bench. All `Server::respond` does is clone various things and then call the router. This should be close enough to the original bench.

Fishrock123 (approved):

src/lib.rs
```diff
@@ -194,10 +194,9 @@
 //! for Async Rust. We have a long journey ahead of us. But we're excited you're here with us!
 
 #![cfg_attr(feature = "docs", feature(doc_cfg))]
-// #![warn(missing_docs)]
-#![warn(missing_debug_implementations, rust_2018_idioms)]
-#![doc(test(attr(deny(rust_2018_idioms, warnings))))]
-#![doc(test(attr(allow(unused_extern_crates, unused_variables))))]
+#![forbid(unsafe_code, rust_2018_idioms)]
+#![deny(missing_debug_implementations, nonstandard_style)]
+#![warn(missing_docs, unreachable_pub, future_incompatible)]
```
taiki-e, line 199:
```suggestion
#![forbid(unsafe_code)]
#![deny(missing_debug_implementations, nonstandard_style)]
#![warn(missing_docs, unreachable_pub, future_incompatible, rust_2018_idioms)]
```

As said in https://github.com/taiki-e/pin-project/issues/281#issuecomment-699543730, I do not recommend using `forbid` on the lints other than `unsafe_code` lint.

src/lib.rs
```diff
@@ -194,10 +194,9 @@
 //! for Async Rust. We have a long journey ahead of us. But we're excited you're here with us!
 
 #![cfg_attr(feature = "docs", feature(doc_cfg))]
-// #![warn(missing_docs)]
-#![warn(missing_debug_implementations, rust_2018_idioms)]
-#![doc(test(attr(deny(rust_2018_idioms, warnings))))]
-#![doc(test(attr(allow(unused_extern_crates, unused_variables))))]
```
taiki-e, line 200: These two lines are needed to [catch the doc test warnings](https://doc.rust-lang.org/nightly/rustdoc/the-doc-attribute.html#testattr). (Normal crate-level lints cannot catch doc test warnings.) So you probably don't need to remove them.

Fishrock123:
See also https://github.com/http-rs/async-sse/pull/11

PR #709, opened 2020-09-27
Remove parsing from `Request::param`

yoshuawuyts:
This PR removes the `parse` portion of `Request::param`, making that an explicit step that needs to be called by end-users. This was inspired by my analysis in https://github.com/rust-lang/rust/pull/75435#issuecomment-695805123 on similar ergonomics issues for a stdlib API. As a result many of our tests that want to extract a `&str` or `String` from the params are more intuitive.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ async-std = { version = "1.6.0", features = ["unstable"] }
 async-trait = "0.1.36"
 femme = "2.0.1"
 futures-util = "0.3.5"
-http-types = "2.2.1"
+http-types = "2.4.0"
 kv-log-macro = "1.0.4"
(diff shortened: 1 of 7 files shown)


jbr (commented):
This &str behavior makes sense, but might we also want a parse_param() with the current behavior?

yoshuawuyts:
@jbr I think it should be easy enough to call `.parse()` on the output of `.param()` to convert it into another type. I don't love the way we require a custom error enum just for that right now, and would prefer if we could remove that entirely in favor of letting people handle the errors individually.

jbr (approved):
On further review, :thumbsup: Would merge if it were fast-forward, but looks like it needs a rebase

PR #710, opened 2020-09-28
implement Endpoint trait for Box<dyn Endpoint>

mendelt:
Right now adding endpoints dynamically is difficult because only types that implement the Endpoint trait directly can be added. When you don't know in advance which endpoint you want to add in a certain situation it can be useful to be able to add boxed endpoints. This implementation makes this possible.

I did not add any tests for this. It seems like a trivial piece of code where the test will probably be more fragile than the code itself.

### DIFF
--- a/src/endpoint.rs
+++ b/src/endpoint.rs
@@ -118,3 +118,10 @@ where
         Ok(next.run(req).await)
     }
 }
+
+#[async_trait]
+impl<State: Clone + Send + Sync + 'static> Endpoint<State> for Box<dyn Endpoint<State>> {
+    async fn call(&self, request: Request<State>) -> crate::Result {
+        self.call(request).await
+    }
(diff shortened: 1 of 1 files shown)


yoshuawuyts (approved):

hasali19:
I just noticed this while going through the code for an unrelated reason. I'm still a bit new to rust but won't this cause an infinite recursion.

mendelt:
Hmm yeah, good catch! I'll see if I can fix that!
thanks!

mendelt:
Ok, this was stupid of me. I forgot to add as_ref in the method where the boxed endpoint should call the endpoint inside the box. So it called itself. This should be fixed in this pr https://github.com/http-rs/tide/pull/711

Thanks again @hasali19 for noticing this.

PR #716, opened 2020-10-07
Add built-in validation for input types

yoshuawuyts:
This adds the `Validate` trait as part of the `convert` and `prelude` submodules. This enables manually calling `.validate` on types, and integrates it directly into `Request::body_json` and `Request::body_form`. This makes it easy to define validation bounds beyond just types on structs.

Overall this should make it easier to define input types for REST APIs in Tide, reducing boilerplate needed. In particular the automatic validation in `Request::body_json` I think will come in useful; converting data into a struct will always check all rules it can, and prevents easy mistakes such as forgetting to call `.validate`.

This is powered by the `validator` crate. While we may want to replace it eventually; it does a really good job at providing the basics, and certainly a good starting point to better understand what else we may want from validation in Tide.

I hope this makes sense. Thanks!

## Examples

```rust
use tide::Request;
use tide::prelude::*;

#[derive(Debug, Deserialize, Validate)]
struct Animal {
    #[validate(length(min = 4))]
    name: String,
    #[validate(range(max = 12))]
    legs: u8,
}

#[async_std::main]
    let mut app = tide::new();
    app.at("/orders/shoes").post(order_shoes);
    Ok(())
}

async fn order_shoes(mut req: Request<()>) -> tide::Result {
    let Animal { name, legs } = req.body_json().await?;
    Ok(format!("Hello, {}! I've put in an order for {} shoes", name, legs).into())
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ pin-project-lite = "0.1.7"
 route-recognizer = "0.2.0"
 serde = "1.0.102"
 serde_json = "1.0.41"
+validator = { version = "0.11", features = ["derive"] }
 
 [dev-dependencies]
 async-std = { version = "1.6.0", features = ["unstable", "attributes"] }
--- a/examples/graphql.rs
+++ b/examples/graphql.rs
@@ -75,7 +75,8 @@ lazy_static! {
 }
 
 async fn handle_graphql(mut request: Request<State>) -> tide::Result {
-    let query: GraphQLRequest = request.body_json().await?;
+    let body = request.take_body();
+    let query: GraphQLRequest = body.into_json().await?;
     let response = query.execute(&SCHEMA, request.state());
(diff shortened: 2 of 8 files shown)


jbr:
Does this mean that someone can't opt out of implementing validate? I'm not sure that's bad, but it's not clear to me that we'd want to universally require that behavior. This seems a bit like the opposite direction of moving string parsing out of param. This might be better as a tide app pattern that we encourage instead of requiring it

Fishrock123:
I really like the idea here but it needs to be optional if it isn't already (and if it is, we may want to have it behind an on-by-default feature).

yoshuawuyts:
> This might be better as a tide app pattern that we encourage instead of requiring it

Fair enough; think we may have to rethink this.

yoshuawuyts:
Closing out for now; let's revisit this again later.

PR #717, opened 2020-10-09
Add an optional cookies feature

belak:
This seems related to #548. It was mostly done as an experiment to get my feet wet in the codebase, but I figured it could be useful for other people as well.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -3,6 +3,7 @@
 use async_std::io;
 use async_std::sync::Arc;
 
+#[cfg(feature = "cookies")]
 use crate::cookies;
 use crate::listener::{Listener, ToListener};
 use crate::log;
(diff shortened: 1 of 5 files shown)


src/server.rs
```diff
@@ -100,15 +101,28 @@ impl<State: Clone + Send + Sync + 'static> Server<State> {
...
-        let mut server = Self {
-            router: Arc::new(Router::new()),
-            middleware: Arc::new(vec![]),
-            state,
-        };
-        server.with(cookies::CookiesMiddleware::new());
-        #[cfg(feature = "logger")]
-        server.with(log::LogMiddleware::new());
-        server
+        #[cfg(any(feature = "cookies", feature = "logger"))]
```
Fishrock123, line 104: This seems unnecessary. Why not just add the one `#[cfg()]` without this outside wrapping? It'l compile to the same thing.

Is it just because something may warn about `mut`?

Fishrock123 (commented):
This seems mostly good, although I'd prefer to clean up what I commented on.

Also, I'm not super familiar with this stuff:
```toml
[[test]]
name = "cookies"
path = "tests/cookies.rs"
required-features = ["cookies"]
```

belak:
```
[[test]]
name = "cookies"
path = "tests/cookies.rs"
required-features = ["cookies"]
```

This stuff was done so these files will only be run if the required feature is enabled. I think you could also use `#![cfg(feature="cookies")]` at the top of the file but they have slightly different behaviors.

src/server.rs
belak: It's because the warning about `mut` is considered an error in CI, so the pipeline fails.

Fishrock123: I think that @jbr will probably strongly dislike this suggestion but, an alternate and perhaps better (if to avoid warnings) way to write this is to just move the `#[cfg()]`'s into the `vec![]`:

```rust
vec![
   #[cfg(feature = "cookies")]
   cookies::CookiesMiddleware::new(),
   #[cfg(feature = "logger")]
   log::LogMiddleware::new(),
]
```

belak: Ah, so just instantiate the server with the given middleware rather than calling `.with`?

Fishrock123: Correct. Again, if we are avoiding a warning for that `mut`, then this seems to be the best way to do so.

Fishrock123: @belak Were you able to make this change? I'd like to ensure this gets in for tide 0.14

belak: Sorry about the delay - give me a few min and I'll have that in

Issue #690, opened 2020-08-31
Add support for routing on subdomains


AlecDivito:
## Feature Request

Not all of the url routing is complete. I'm just wondering if  support for routing on subdomains could be added into the framework.

## Detailed Description

Add new methods to the `router.rs` and `server.rs` to support users to route on subdomains as well. An example of the resulting api I am thinking of would look like so:

```rust
let mut app = tide::new();
app.on_subdomain(":sub1.:sub2.:sub3").at("/").with(validate_subdomain).all(handle_user);
```

## Context

I have an application that can have multiple tenets that can upload content to the server and would like to configure my webserver to be able to route on subdomains. This will benefit other users of tide who want to add the same functionality to their own application.  

## Possible Implementation

I haven't dove into the code base that much but i would assume that I would need to add functionality to `server.rs` and `router.rs`. I know this maybe a big addition so I wouldn't mind taking a chance and implementing it and putting in a PR.

Just wondering if this project is also interested in adding support for it and want to start a discussion.

yoshuawuyts:
We had a bit of a chat about this earlier today, I think this is a fairly promising direction. Earlier art for subdomain routing includes:

- [koa-sub-domain](https://www.npmjs.com/package/koa-sub-domain)
- [Rails subdomain routing](https://gist.github.com/indiesquidge/b836647f851179589765) ([official docs](https://guides.rubyonrails.org/routing.html#request-based-constraints))

The Rails subrouting API seems quite interesting; if we were to translate that to Tide it could look like:

```rust
let mut app = tide::new();
app.subdomain("blog").at("/").get(|_| async { Ok("welcome to my blog") });
app.listen("localhost:8080").await?;
```

I'm leaning towards `app.subdomain` rather than abbreviating it as `app.sub`  to prevent confusion with `app.nest`. Also we should only add `subdomain` to `Server` but not `Route`. Because subdomains feel like they should be declared at the root of the app, not nested within routes. As such we should perhaps also guard that when nesting apps we disallow subdomains on the app we're nesting.

---

That's my opinion, @jbr had some different ideas that I'm hoping he'll share!

jbr:
## proposal in brief
```rust
app.at("/path/still/works").get(…);
app.at("http://*/only-http").get(…);
app.at("//example.com/any-scheme/but-only-example.com").get(…);
```

## why don't other servers do this?
[...]

PR #720, opened 2020-10-09
Add subdomain routing

AlecDivito:
Adds support for routing on subdomains.

## Description

Look at issue #690.

## Motivation and Context

I needed support for routing on subdomains inside of my project.

## How Has This Been Tested?

New tests have been added into the `/tests/namespace.rs` file.

## Types of changes

- [] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] I have read the CONTRIBUTING document.
- [x] Added example of new feature


Fishrock123:
@jbr Had a direction idea on this that wasn't posted to the original issue which I personally think is much better and more powerful than `.subdomain()`.

It's possible that idea may not work out or may not materialize, but I think we should either consider waiting or consider that this api may later be dropped.

AlecDivito:
Yes, he actually posted his idea inside of issue #720 but then soon deleted it. Though i still have his response. He proposed doing routing using the entire url like so:

```rust
app.at("/path/still/works").get(…);
app.at("http://*/only-http").get(…);
app.at("//example.com/any-scheme/but-only-example.com").get(…);
```

I agree that this would be a lot more powerful but I also think that it would also be a lot more work. I just assumed that because he deleted his comment it wasn't being considered anymore.

joshtriplett:
I would love to have support for this, as well.

Issue #703, opened 2020-09-22
tide::log::middleware: line number/backtrace for errors


bbigras:
## Detailed Description

Right now we can see errors in the log but we don't have the line number or a backtrace.

## Context

Easier to track down a bug.

jbr:
I believe this might be blocked on an http-types release, but then will absolutely be possible

bbigras:
Awesome. Thank you very much.

bbigras:
On a related note, when using `anyhow`, could we get the error and the cause in the logger?

In this example, the case with `tide::log::middleware` doesn't show `No such file or directory (os error 2)`.

Should I open a new issue/feature request?

**code**
```rust
use anyhow::{Error, Context};
use tide::Request;

use std::fs::File;

#[async_std::main]
async fn main() -> tide::Result<()> {
    if let Err(e) = example() {
        println!("err: {:?}", e);
    }

    tide::log::with_level(tide::log::LevelFilter::Info);

    let mut app = tide::new();
    app.at("/").get(index);
    app.with(tide::log::LogMiddleware::new());
    app.listen("127.0.0.1:8080").await?;
    Ok(())
}

fn example() -> Result<(), Error> {
    let _file = File::open("non-existing-file1.txt")
        .context("can't open non-existing-file1.txt")?;
    unimplemented!();
}

async fn index(_req: Request<()>) -> tide::Result {
    let _file = File::open("non-existing-file2.txt")
        .context("can't open non-existing-file2.txt")?;
    unimplemented!();
}
```

**output**
```
err: can't open non-existing-file1.txt

Caused by:
    No such file or directory (os error 2)

tide::log::middleware <-- Request received
    method GET
    path /
tide::log::middleware Internal error --> Response sent
    message can't open non-existing-file2.txt
    method GET
    path /
    status 500
    duration 79.312µs
```

Fishrock123:
Fwiw I also came across this and I have a patch essentially ready from an internal logger.

bbigras:
Which case, the backtrace one or the anyhow one?

Fishrock123:
Backtrace, error type name. I think I have a hunch on how to get the description properly as well.

Fishrock123:
Related: https://github.com/http-rs/http-types/pull/258

bbigras:
It seems to work perfectly. Thank you very much!

PR #721, opened 2020-10-09
Logger: properly print debug from errors

Fishrock123:
Also adds the error type name.

Fixes: https://github.com/http-rs/tide/issues/703

### DIFF
--- a/src/log/middleware.rs
+++ b/src/log/middleware.rs
@@ -50,7 +50,8 @@ impl LogMiddleware {
         if status.is_server_error() {
             if let Some(error) = response.error() {
                 log::error!("Internal error --> Response sent", {
-                    message: format!("\"{}\"", error.to_string()),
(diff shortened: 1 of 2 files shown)


src/log/middleware.rs
```diff
@@ -50,7 +50,8 @@ impl LogMiddleware {
         if status.is_server_error() {
             if let Some(error) = response.error() {
                 log::error!("Internal error --> Response sent", {
-                    message: format!("\"{}\"", error.to_string()),
+                    message: format!("{:?}", error),
```
Fishrock123, line 53: When using the `Debug` formatter on Nightly with `RUST_BACKTRACE` enabled, this will also print out with a backtrace.

Fishrock123:
I now have two codebases using custom loggers awaiting this, haha.

PR #724, opened 2020-10-16
Add `Server::listen_with`

### LINKED ISSUE #712: Add `Server::listen_with` to invoke logic _after_ the server has started
At work we were having an ordering issue where we want to trigger something once the server has started listening. However currently there is no good way of defining that since the `listen` method just waits until the server shuts down.

So I propose we add an additional `Server::listen_with` method that takes a callback as the second argument. From there we should be able to share how and what the server is listening on.

## Example

This is the logic we use at work: the "server listening" message is emitted before we actually listen. This is problematic for UDS connections since trying to send a message on a non-existing file will fail, which is slightly different from TCP which _is_ able to retry.

```rust
    // NOTE: This println is essential for the correct working of the client.
    info!("Started http server");

    // Start the Tide server and log the server details.
    // TODO: Tide should have a panicking listen_unix impl.
    if let Some(path) = opts.unix_path() {
        app.listen(&*format!("http+unix://{}", path)).await?;
    } else {
        app.listen((&*opts.host, opts.port)).await?;
    }
```

yoshuawuyts:
Implements `Server::listen_with` which takes an async closure that can be used to report information about the connection. Closes #712.

The way this works is by introducing a new trait `tide::connection::Report` which is called once per listener and is passed an instance of `tide::connection::ListenInfo`. By default `tide::listen` invokes a default log-based reporter (`tide::log::Reporter`), but by using `tide::listen_with` this can be overwritten.

This can also serve as a foundation to gradually expand our logging capabilities. For now `ListenInfo` only logs the string passed, but it'd be nice if we could eventually also include things like a vec of HTTP protocol version supported, TLS settings, and perhaps even routes registered. Also since the callback returns a future, it can be used to delay accepting connections until some signal is passed -- setting us up to experiment with something like [Swift's service lifecycles](https://swift.org/blog/swift-service-lifecycle/). We can gradually expand on this over time, and enable other people to experiment with this just by exposing this hook. Thanks!

## Examples

```rust
use tide::listener::ListenInfo;

let mut app = tide::new();
app.listen_with("127.0.0.1:8080", |info: ListenInfo| async move {
    println!("started listening on {}!", info.connection());
    Ok(())
}).await?;
```


Fishrock123:
I am curious, would something like this work?
```rust
let conn: ListenerConnection = server.listen_conn(addr).await?;

println!("{}", conn.addr());

// What actually keeps calling poll for new connections
let conn.await?;
```

Or does that not really work because it isn't really "listening" without being polled through the await?

jbr:
I really like the thing that @Fishrock123 suggested and plan on experimenting with some variants like that before commenting more substantially. Could I ask that we hold off on merging this for a bit while I try some variations? I see the problem space and am not thrilled with the solution offered by this PR, but would like to see if I can come up with an alternative for discussion. Callbacks like this don't feel particularly rusty to me, and they don't seem particularly like the swift example either. I like the idea of registering hooks on the server, but think that we should tackle that more generic problem instead of changing the listen signature.  I also had a solution to this in the initial version of Listeners, and want to revisit that. A callback should be attached to the listener itself, which shouldn't change the signature for Server::listen

Fishrock123:
If we are considering something like "events" I think we would benefit from taking a look at how Node.js has integrated Promises into events with awaitable helpers:
- [`events.once()` -> `Promise`](https://nodejs.org/dist/latest-v14.x/docs/api/events.html#events_events_once_emitter_name)
- [`events.on()` -> `AsyncIterator`](https://nodejs.org/dist/latest-v14.x/docs/api/events.html#events_events_on_emitter_eventname)

We could map those to `Future` and `Stream`, the question becomes around executing a second context (like an `async_std::task`) in order to `poll` the listener.

yoshuawuyts:
@jbr I think you have the right idea with `bind` returning a listener; in particular:

```rust
let listener = server.bind("localhost:8000").await?;
println!("started listening on {}!", &listener);
listener.await?;
```

Seems like a good idea. This seems like the right place to start experimenting with.

---

@Fishrock123 that's an interesting idea. We should probably scope out what it is that we want from the service in order to see if if it would make sense to fit that in, but I think the API mapping makes sense.

mendelt:
Sorry for jumping on this so late. I like the feature but I'm not so fond of the api. `listen_with` is not very clear to me, especially since we already use `with` for adding middleware.
Maybe we could split adding the callback and starting the server into separate methods. I think this will make for a better api? Have an `on_start_listening` method to add the callback and not touch the `listen` method keeps things nice and focused. It also makes it easier if we want to add other server lifecycle callbacks later.

mendelt:
Sorry for jumping on this so late in the review process.
I like the feature but I think the API can be better.
The name `listen_with` is very unclear to me. With is already used for adding middleware so it's becoming an overloaded term. Also it conflates two actions, starting the server and adding the callback.
I would prefer to have a separate method. Something like `on_listener_started` on the server where you can add a callback. And then leave the `listen` method as-is.

ratmice:
Maybe a silly question, but would this `bind` function allow multiple listeners on a `Server`?
The only way I could figure out to get both a tide-rustls, and a http listener, was to create 2 servers, and join as below.

```
let app = route(tide::with_state(state.clone()));
let app_tls = route(tide::with_state(state));
let ((), ()) = try_join!(
              app_tls.listen(
                    TlsListener::build()
                        .addrs("localhost:4433")
                        .cert(std::env::var("TIDE_CERT_PATH").unwrap())
                        .key(std::env::var("TIDE_KEY_PATH").unwrap()),
                ),
              app.listen("localhost:8080"),
)?;
```

I didn't see any cases above of which involved multiple listeners (So i'm not sure if it would work), but it at least seems plausible to me, that people might assume by the name bind, that you can `server.bind` multiple listeners.

yoshuawuyts:
@ratmice it sounds like you may be looking for [`tide::Listener::ConcurrentListener`](https://docs.rs/tide/0.14.0/tide/listener/struct.ConcurrentListener.html). Using this will enable you to listen on both `http` and `https` with the same application.

ratmice:
Awesome, disregard that comment above then!

jbr:
@yoshuawuyts Let's chat about this in Zulip at some point?

yoshuawuyts:
Worked on this today; `Server::bind` now works though it requires a fair bit of cleanup. The patch is not done, but I'm rounding out for the day so I wanted to share the status update.

I attempted do a clean-slate impl of `Server::bind`, but got really deep into the weeds of separating "connect the tcp stream" from "accept connections" at the trait level, and realized that would become a nightmare to maintain. I wasn't even sure if we could get past the coherence issues. And finally, we'd lack a uniform interface to access information about the listener. 

So I've opted to keep "connect the tcp stream" and "accept connections" together in the same trait impl, but require implementers to send information data through a channel in between. This is currently implemented on top of the generic callback interface we had before ("Reporter" etc), but we can get rid of all the extra type params in favor of just using a dedicated type.

Next steps for this PR are:

- [ ] pass tests
- [ ] remove the callback interface
- [ ] rename types so that they actually make sense
- [ ] add docs & examples

yoshuawuyts:
Talked with @jbr, and wrote #740 which modifies the underlying trait to enable binding. This feels like a much more solid design overall.

PR #725, opened 2020-10-21
Add `serve_file` method to route

mendelt:
This pr adds a serve_file method alongside the already existing `serve_dir` method.

### DIFF
--- a/src/fs/serve_file.rs
+++ b/src/fs/serve_file.rs
@@ -7,15 +7,15 @@ use async_std::path::PathBuf as AsyncPathBuf;
 use async_trait::async_trait;
 
 pub struct ServeFile {
-    file: AsyncPathBuf,
+    path: AsyncPathBuf,
 }
 
 impl ServeFile {
     /// Create a new instance of `ServeFile`.
-    pub(crate) fn init(file: impl AsRef<Path>) -> io::Result<Self> {
-        let file = file.as_ref().to_owned().canonicalize()?;
+    pub(crate) fn init(path: impl AsRef<Path>) -> io::Result<Self> {
+        let file = path.as_ref().to_owned().canonicalize()?;
         Ok(Self {
-            file: AsyncPathBuf::from(file),
+            path: AsyncPathBuf::from(file),
         })
     }
 }
@@ -23,13 +23,12 @@ impl ServeFile {
 #[async_trait]
 impl<State: Clone + Send + Sync + 'static> Endpoint<State> for ServeFile {
     async fn call(&self, _: Request<State>) -> Result {
-        if !self.file.exists().await {
-            log::warn!("File not found: {:?}", &self.file);
-            println!("File not found: {:?}", &self.file);
+        if !self.path.exists().await {
+            log::warn!("File not found: {:?}", &self.path);
             Ok(Response::new(StatusCode::NotFound))
         } else {
             Ok(Response::builder(StatusCode::Ok)
-                .body(Body::from_file(&self.file).await?)
+                .body(Body::from_file(&self.path).await?)
                 .build())
         }
     }
@@ -74,7 +73,7 @@ mod test {
     #[async_std::test]
     async fn should_serve_404_when_file_missing() {
         let serve_file = ServeFile {
-            file: AsyncPathBuf::from("gone/file"),
+            path: AsyncPathBuf::from("gone/file"),
         };
 
         let res: Response = serve_file.call(request("static/foo")).await.unwrap().into();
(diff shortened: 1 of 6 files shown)


src/fs/serve_file.rs
```diff
@@ -0,0 +1,84 @@
...
+use async_std::path::PathBuf as AsyncPathBuf;
+use async_trait::async_trait;
+
+pub struct ServeFile {
+    file: AsyncPathBuf,
+}
+
+impl ServeFile {
+    /// Create a new instance of `ServeFile`.
+    pub(crate) fn init(file: impl AsRef<Path>) -> io::Result<Self> {
```
Fishrock123, line 15:
```suggestion
    pub(crate) fn init(path: impl AsRef<Path>) -> io::Result<Self> {
```

src/fs/serve_file.rs
```diff
@@ -0,0 +1,84 @@
...
+        })
+    }
+}
+
+#[async_trait]
+impl<State: Clone + Send + Sync + 'static> Endpoint<State> for ServeFile {
+    async fn call(&self, _: Request<State>) -> Result {
+        if !self.file.exists().await {
+            log::warn!("File not found: {:?}", &self.file);
+            println!("File not found: {:?}", &self.file);
```
Fishrock123, line 28:
```suggestion
```

Fishrock123 (commented):
seems reasonable

src/fs/serve_file.rs
mendelt: Thank you! Done, also in a couple of other places.

src/fs/serve_file.rs
mendelt: Thank you! removed this.

mendelt:
I added `serve_file` to Route to be consistent with how `serve_dir` works for now.
Personally I would like the Route struct to stay a bit cleaner. We could add specific helper methods like serve_dir and serve_file with extension traits, that way helper methods like this can be added to Route by importing the extension traits you need.
Does anyone have an opinion on this?

src/fs/serve_file.rs
```diff
@@ -0,0 +1,83 @@
...
+        Ok(Self {
+            path: AsyncPathBuf::from(file),
+        })
+    }
+}
+
+#[async_trait]
+impl<State: Clone + Send + Sync + 'static> Endpoint<State> for ServeFile {
+    async fn call(&self, _: Request<State>) -> Result {
+        if !self.path.exists().await {
```
goto-bus-stop, line 26: This looks like a potential [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) issue. I think we should just try to read the file immediately and return `StatusCode::NotFound` if that fails with `io::ErrorKind::NotFound`.

`serve_dir()` has the same issue. We could also land this as is and fix it in a separate PR, whichever you prefer

goto-bus-stop (commented):
I don't think an extension trait is necessary at this stage. There are only a few different methods on the trait, if you count the 10 HTTP method aliases as one :stuck_out_tongue_winking_eye:

src/fs/serve_file.rs
mendelt: Ah yes, I copied this from `serve_dir` I'd prefer to fix this as part of this PR, at least for `serve_file` I don't really want to add known bugs to the code-base if I can avoid it. I'll fix this later today.

mendelt:
> I don't think an extension trait is necessary at this stage. There are only a few different methods on the trait, if you count the 10 HTTP method aliases as one stuck_out_tongue_winking_eye
Funny you put it that way. Because in my fluent routes crate I put all of those aliases in their own extension trait too :-) https://github.com/mendelt/tide-fluent-routes/blob/c20244d6e2254f275ebc0ad46ebed2c621ef90a6/src/routebuilder.rs#L27

But for now this might be a bit premature, and I'd want to put serve_dir in an extension trait too. And that will break backward compatibiltiy.

src/fs/serve_file.rs
mendelt: Fixed

yoshuawuyts (approved):
Thanks so much for this PR @mendelt; this looks good!

In the past there's been talk about perhaps moving `serve` / `serve_file` into some sort of `FileServer` struct which would resemble much of how Tide's redirects work today. But we still need to finish up the design for that, and I don't think we should miss out on a chance to make serving single files easier until then.

Going to go ahead and merge this; thanks heaps!

yoshuawuyts:
Tested this locally, and clippy + rustfmt pass. We need to update the CI, but in the interim I think this is fine to merge as-is. Thanks!

PR #726, opened 2020-10-23
Request: make ::new() public

Fishrock123:
Usually tide does this for you once listening, and passes the request through middleware and to an endpoint.
However, if you wish to call an endpoint manually, then this is useful.

(We ran into this internally.)

I realize this may not be considered desirable, so... discuss, I suppose.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -33,7 +33,28 @@ pin_project_lite::pin_project! {
 
 impl<State> Request<State> {
     /// Create a new `Request`.
-    pub(crate) fn new(state: State, req: http_types::Request, route_params: Vec<Params>) -> Self {
+    ///
+    /// Usually tide does this for you once listening, and passes the request through middleware and to an endpoint.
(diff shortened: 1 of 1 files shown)


jbr:
Could you describe the use case for calling an endpoint manually better?

Fishrock123:
> I'm trying to understand why `server.respond(http_types::Request)` _is_ insufficient

Hmm, I hadn't thought of that. Seems like unnecessary routing though.

jbr:
does it accomplish the goal though? increasing surface area with an api that might seem like a good place to start seems like something to avoid if at all possible

jbr:
Alternatively, since this is advanced usage,
```rust
let req: tide::Request<()> = tide::http::Request::new(
    tide::http::Method::Get,
    tide::http::Url::parse("http://_").unwrap(),
)
.into();
```

Fishrock123:
Ah. The `Into` is smart. I'm surprised I forgot about that. Yeah that means we don't need this.

PR #730, opened 2020-10-30
Add with_cookie_domain method to SessionMiddleware

rhuffus:
We are using tide as a session/authentication context and we need to set the domain of the cookie as the root domain on the SessionMiddleware.

This PR adds the hability to set de cookie domain through the SessionMiddleware builder.


jbr (approved):
Thanks!

PR #732, opened 2020-10-30
Fix port in README

jtdowney:
The readme example listens on port 8080 but the curl examples following it were hitting 8000.


jbr (approved):
Thanks!

PR #734, opened 2020-11-02
In `serve_dir`, only strip the prefix from the start of the path once.

lumi-me-not:
Currently, `serve_dir` is using `str::trim_start_matches`, which may strip the prefix multiple times.

I've changed this to use `strip_prefix`, which will only strip it once.

I think this unwrap is okay, because the prefix is guaranteed to be there. Do tell me if I'm wrong.

### DIFF
--- a/src/fs/serve_dir.rs
+++ b/src/fs/serve_dir.rs
@@ -25,7 +25,7 @@ where
 {
     async fn call(&self, req: Request<State>) -> Result {
         let path = req.url().path();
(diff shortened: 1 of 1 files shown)


joshtriplett:
It looks like this failed spuriously due to a transient CI issue.

jbr (approved):
seems reasonable to me. I also believe the prefix is always expected to be present so the unwrap should be fine

PR #735, opened 2020-11-03
Logger improvements

Fishrock123:
Some minor follow-up for my previous change to the logger which made it use the debug formatter. Very helpful for debugging  but probably annoying in "production-like" logs.

This also does some code deduplication.

I'd like to clean this up more but that probably means investing time in either changing the log format entirely or separating out the key-value formatter from log so that we can call it separately before passing to a log level.

### DIFF
--- a/src/log/middleware.rs
+++ b/src/log/middleware.rs
@@ -49,8 +49,13 @@ impl LogMiddleware {
         let status = response.status();
         if status.is_server_error() {
             if let Some(error) = response.error() {
+                #[cfg(debug_assertions)]
+                let message = format!("{:?}", error);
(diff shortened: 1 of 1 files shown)


Fishrock123:
```
error: component 'rustfmt' for target 'x86_64-unknown-linux-gnu' is unavailable for download for channel nightly
Sometimes not all components are available in any given nightly.
Error: Process completed with exit code 1.
```

Well that's fun. Maybe we should run fmt and clippy on beta and not nightly.

tirr-c:
Well rustfmt is broken for over a week according to rust-toolstate, I haven't seen this before. Maybe we really need beta?

PR #737, opened 2020-11-07
Fix param documentation

joshtriplett:
param doesn't infer its return type, it always returns a string.
Document it accordingly.


jbr:
Thanks!

PR #714, opened 2020-10-07
dev-deps: surf @ 2.0.0

Fishrock123:
Might as well pick up any patches automatically.

PR #740, opened 2020-11-11
Add `Server::bind`

### LINKED ISSUE #712: Add `Server::listen_with` to invoke logic _after_ the server has started

yoshuawuyts:
Alternative to #724's `Server::listen_with` implementation, as suggested by @jbr.

## `Server::bind`

This patch introduces a mechanism to "start listening" but don't "start accepting requests" quite yet. This enables logging in between the two actions, which fixes #712  An example:

```rust
use tide::prelude::*;

let mut app = tide::new();
let mut listener = app.bind("127.0.0.1:8080").await?;
for info in listener.info().iter() {
    println!("Server listening on {}", info);
}
listener.accept().await?;
```

In order for this to work we had to also adjust `Listener::listen`, breaking it up into `Listener::bind` which opens the connection, and `Listener::accept` which starts accepting the connections. Terminology-wise "listening" is the action of "binding" + "accepting".

## `listener::ConnectionInfo`

In addition to introducing `Server::bind` this patch also introduces the `ListenInfo` struct. This contains various information about the connection: the url in string form, whether the connection is encrypted, and what the underlying transform is. This information is returned through the `Listener::info` method, and always returns a vector because listeners such as `ConcurrentListener` allow the server to listen on multiple connections at once.

The `Server::listen` method uses `log::info` to log out the various connections we're listening on, but using `Server::bind` + `Listener::accept` end-users can choose to author their own log output. This first implementation is mostly intended to provide access to some basics; I hope over time we can polish it as we learn about which information is useful to surface from the listeners.

### DIFF
--- a/src/listener/failover_listener.rs
+++ b/src/listener/failover_listener.rs
@@ -31,9 +31,11 @@ use async_std::io;
 ///    })
 ///}
 ///```
-
 #[derive(Default)]
-pub struct FailoverListener<State>(Vec<Box<dyn Listener<State>>>);
+pub struct FailoverListener<State> {
+    listeners: Vec<Option<Box<dyn Listener<State>>>>,
+    index: Option<usize>,
+}
 
 impl<State> FailoverListener<State>
 where
@@ -41,7 +43,10 @@ where
 {
     /// creates a new FailoverListener
     pub fn new() -> Self {
-        Self(vec![])
+        Self {
+            listeners: vec![],
+            index: None,
+        }
     }
 
     /// Adds any [`ToListener`](crate::listener::ToListener) to this
@@ -64,7 +69,7 @@ where
     where
         L: ToListener<State>,
     {
-        self.0.push(Box::new(listener.to_listener()?));
+        self.listeners.push(Some(Box::new(listener.to_listener()?)));
         Ok(())
     }
 
@@ -93,13 +98,16 @@ impl<State> Listener<State> for FailoverListener<State>
 where
     State: Clone + Send + Sync + 'static,
 {
-    async fn listen(&mut self, app: Server<State>) -> io::Result<()> {
-        for listener in self.0.iter_mut() {
-            let app = app.clone();
-            match listener.listen(app).await {
-                Ok(_) => return Ok(()),
+    async fn bind(&mut self) -> io::Result<()> {
+        for (index, listener) in self.listeners.iter_mut().enumerate() {
+            let listener = listener.as_deref_mut().expect("bind called twice");
+            match listener.bind().await {
+                Ok(_) => {
+                    self.index = Some(index);
+                    return Ok(());
+                }
                 Err(e) => {
-                    crate::log::info!("unable to listen", {
+                    crate::log::info!("unable to bind", {
                         listener: listener.to_string(),
                         error: e.to_string()
                     });
@@ -112,20 +120,37 @@ where
             "unable to bind to any supplied listener spec",
         ))
     }
+
+    async fn accept(&mut self, app: Server<State>) -> io::Result<()> {
+        match self.index {
+            Some(index) => {
+                let mut listener = self.listeners[index].take().expect("accept called twice");
+                listener.accept(app).await?;
+                Ok(())
+            }
+            None => Err(io::Error::new(
+                io::ErrorKind::AddrNotAvailable,
+                "unable to listen to any supplied listener spec",
+            )),
+        }
+    }
 }
 
 impl<State> Debug for FailoverListener<State> {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        write!(f, "{:?}", self.0)
+        write!(f, "{:?}", self.listeners)
     }
 }
 
 impl<State> Display for FailoverListener<State> {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         let string = self
-            .0
+            .listeners
             .iter()
-            .map(|l| l.to_string())
+            .map(|l| match l {
+                Some(l) => l.to_string(),
+                None => String::new(),
+            })
             .collect::<Vec<_>>()
             .join(", ");
 
--- a/src/listener/failover_listener.rs
+++ b/src/listener/failover_listener.rs
@@ -35,7 +35,10 @@ use async_std::io;
 #[derive(Default)]
 pub struct FailoverListener<State>(Vec<Box<dyn Listener<State>>>);
 
-impl<State: Clone + Send + Sync + 'static> FailoverListener<State> {
+impl<State> FailoverListener<State>
+where
+    State: Clone + Send + Sync + 'static,
+{
     /// creates a new FailoverListener
     pub fn new() -> Self {
         Self(vec![])
@@ -57,7 +60,10 @@ impl<State: Clone + Send + Sync + 'static> FailoverListener<State> {
     /// # std::mem::drop(tide::new().listen(listener)); // for the State generic
     /// # Ok(()) }
     /// ```
-    pub fn add<TL: ToListener<State>>(&mut self, listener: TL) -> io::Result<()> {
+    pub fn add<L>(&mut self, listener: L) -> io::Result<()>
+    where
+        L: ToListener<State>,
+    {
         self.0.push(Box::new(listener.to_listener()?));
         Ok(())
     }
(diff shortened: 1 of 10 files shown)


src/listener/to_listener_impls.rs
```diff
@@ -188,55 +238,31 @@ impl<TL: ToListener<State>, State: Clone + Send + Sync + 'static> ToListener<Sta
...
         listener.to_listener()
     }
 
     #[test]
     fn url_to_tcp_listener() {
         let listener = listen(Url::parse("http://localhost:8000").unwrap()).unwrap();
-        assert!(matches!(
-            listener,
-            ParsedListener::Tcp(TcpListener::FromAddrs(_, None))
-        ));
```
yoshuawuyts, line 201: `TcpListener` and `UnixListener` have been reworked to no longer use an enum, which means the internal state machine can no longer be inspected. However we maintain the tests that validate the output is correct.

src/listener/unix_listener.rs
```diff
@@ -7,63 +7,44 @@ use std::fmt::{self, Display, Formatter};
...
 use async_std::prelude::*;
-use async_std::{io, path::PathBuf, task};
+use async_std::{io, task};
+use async_std::path::PathBuf;
 
 /// This represents a tide [Listener](crate::listener::Listener) that
 /// wraps an [async_std::os::unix::net::UnixListener]. It is implemented as an
 /// enum in order to allow creation of a tide::listener::UnixListener
-/// from a [PathBuf] that has not yet been opened/bound OR from a bound
+/// from a [`PathBuf`] spec that has not yet been bound OR from a bound
```
yoshuawuyts, line 16: oops; I copy pasted the `TcpListener` impl and changed it. Looks like some of the wording came over. That shouldn't be too big a deal I hope.

tests/log.rs
```diff
@@ -26,7 +26,7 @@ async fn test_server_listen(logger: &mut logtest::Logger) {
         .unwrap();
     assert_eq!(
         record.args(),
-        format!("Server listening on http://[::1]:{}", port)
+        format!("Server listening on http://127.0.0.1:{}", port)
```
yoshuawuyts, line 29: I changed these tests because they kept failing on WSL. "localhost" parse down to "127.0.0.1" instead of the Ipv6 equivalent. Didn't know that was a thing. Changed them so they parse the same on all platforms.

ratmice:
> This enables logging in between the two actions, which fixes #714

Should be #712 I believe (in the PR initial comment).

yoshuawuyts:
@ratmice good catch; updated!

src/listener/failover_listener.rs
```diff
@@ -73,21 +86,30 @@ impl<State: Clone + Send + Sync + 'static> FailoverListener<State> {
...
+        for (index, listener) in self.listeners.iter_mut().enumerate() {
+            let listener = listener.as_deref_mut().expect("bind called twice");
+            match listener.bind(app.clone()).await {
+                Ok(_) => {
+                    self.index = Some(index);
+                    return Ok(());
+                }
                 Err(e) => {
-                    crate::log::info!("unable to listen", {
+                    crate::log::info!("unable to bind", {
```
Fishrock123, line 112:
```suggestion
                    crate::log::warn!("unable to bind", {
```

src/listener/failover_listener.rs
```diff
@@ -100,20 +122,47 @@ impl<State: Clone + Send + Sync + 'static> Listener<State> for FailoverListener<
             "unable to bind to any supplied listener spec",
         ))
     }
+
+    async fn accept(&mut self) -> io::Result<()> {
+        match self.index {
+            Some(index) => {
+                let mut listener = self.listeners[index].take().expect("accept called twice");
```
Fishrock123, line 129: also regarding `expect`

src/listener/mod.rs
```diff
@@ -26,18 +30,46 @@ pub(crate) use tcp_listener::TcpListener;
...
+    State: Send + Sync + 'static,
+{
+    /// Bind the listener. This starts the listening process by opening the
+    /// necessary network ports, but not yet accepting incoming connections. This
+    /// method must be called before `accept`.
+    async fn bind(&mut self, app: Server<State>) -> io::Result<()>;
+
+    /// Start accepting incoming connections. This method must be called only
+    /// after `bind` has succeeded.
+    async fn accept(&mut self) -> io::Result<()>;
```
Fishrock123, line 49: What if we put this function on something that `bind` returns? Or would that need to consume the app defeating the point?

src/listener/mod.rs
```diff
@@ -52,3 +84,50 @@ pub(crate) fn is_transient_error(e: &io::Error) -> bool {
...
 }
+
+/// Information about the `Listener`.
+///
+/// See [`Report`](../listener/trait.Report.html) for more.
+#[derive(Debug, Clone)]
+pub struct ListenInfo {
+    conn_string: String,
+    transport: String,
+    tls: bool,
```
Fishrock123, line 95: Maybe this should be an option over some tls version Enum?

src/listener/tcp_listener.rs
```diff
@@ -69,11 +62,41 @@ fn handle_tcp<State: Clone + Send + Sync + 'static>(app: Server<State>, stream:
...
-    async fn listen(&mut self, app: Server<State>) -> io::Result<()> {
-        self.connect().await?;
-        let listener = self.listener()?;
-        crate::log::info!("Server listening on {}", self);
+impl<State> Listener<State> for TcpListener<State>
+where
+    State: Clone + Send + Sync + 'static,
+{
+    async fn bind(&mut self, server: Server<State>) -> io::Result<()> {
+        assert!(self.server.is_none(), "`bind` should only be called once");
```
Fishrock123, line 70: An Err might be nicer here.

Fishrock123 (commented):
Overall super happy this is nearing completion. Have a few comments tho.

src/listener/mod.rs
Fishrock123: Actually I think `Server::bind()` already does this.

src/listener/failover_listener.rs
```diff
@@ -73,21 +86,30 @@ impl<State: Clone + Send + Sync + 'static> FailoverListener<State> {
...
-            let app = app.clone();
-            match listener.listen(app).await {
-                Ok(_) => return Ok(()),
+impl<State> Listener<State> for FailoverListener<State>
+where
+    State: Clone + Send + Sync + 'static,
+{
+    async fn bind(&mut self, app: Server<State>) -> io::Result<()> {
+        for (index, listener) in self.listeners.iter_mut().enumerate() {
+            let listener = listener.as_deref_mut().expect("bind called twice");
```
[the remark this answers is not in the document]
yoshuawuyts: `expect` here is intentional: encoding this in the type system gets messy; but we do want this to fail quickly and clearly when the constraints are violated. That's why this is a static assertion with no way to recover. Perhaps in the future we can replace this with a token that needs to be threaded through in the type system, or separate traits or something.

src/listener/failover_listener.rs
yoshuawuyts: oh yeah, good call! -- I think this may even be log level error.

src/listener/failover_listener.rs
yoshuawuyts: Same as above, `expect` here is intended.

src/listener/tcp_listener.rs
yoshuawuyts: Same as above: failing early is very much the intended behavior. Perhaps we can encode it in the type system in the future, but for now this is intended to fail before booting and signal the programmer needs to change their usage patterns -- rather than signaling an error that originates from the the operating system (e.g. "port already in use" or similar).

src/listener/mod.rs
yoshuawuyts: Oh, that's a fun one! -- It'd be nice if we could get information about the TLS version / features used -- but we probably should design that further when we introduce TLS into Tide itself. I'm thinking until then just having a bool may be good enough.

yoshuawuyts:
Talked about this yesterday during the core meeting; @jbr seemed to be positive about the PR, so together with the existing review I think this is ready to merge!

jbr:
I didn't fully read it, but the general direction seemed good :thumbsup:

PR #742, opened 2020-11-12
Add simple state example

joshtriplett:
I found this example useful in understanding how a tide-based server can
handle and maintain state.

### DIFF
--- a/examples/state.rs
+++ b/examples/state.rs
@@ -1,0 +1,34 @@
+use std::sync::atomic::{AtomicU32, Ordering};
+use std::sync::Arc;
+
+struct State {
+    value: AtomicU32,
+}
+
+impl State {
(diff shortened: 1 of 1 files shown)


joshtriplett:
Re-pushed with formatting fixed.

jbr:
Thanks! I think we'd probably want to encourage people to put the Arc inside of their State (`struct State { value: Arc<AtomicU32> }` and `Server<State>` instead of `struct State { value: AtomicU32 }` and `Server<Arc<State>>`), since often State will contain types that are already Clone, like database connection pools. Happy to be convinced that this is a better exemplar, though

joshtriplett:
@jbr Perfectly happy to do it that way instead.

jbr:
Thanks!

Issue #528, opened 2020-05-22
Graceful shutdown


ivan770:
Tide could store `JoinHandle` of every request it handles, and `await` them when server is shutting down.

https://github.com/http-rs/tide/blob/7d0b8484cafbfacf4c3c6859caa83a8c9756202e/src/server.rs#L300-L311
[async-std book](https://book.async.rs/tutorial/clean_shutdown.html)

Fishrock123:
I think the server only shuts down in a `panic!` or due to an external signal, such `SIGINT`, or whatever the Windows equivalents are.

I suppose this could possibly be desirable for `SIGTERM` which allows for such a thing. If you panic I'm not sure it is ever really safe to continue.

Fishrock123:
Oh I think I see what you are saying - essentially we should have a way to await all tasks which are still alive, I think?

ivan770:
> 
> 
> Oh I think I see what you are saying - essentially we should have a way to await all tasks which are still alive, I think?

Yes. I mean, new requests are not handled, but at least those that are still awaiting response should be finished

mehcode:
Was just coming to make an issue for this. It's critical from a production standpoint that a load balancer or swarm manager can initiate a controlled or graceful shutdown.

Any requests that came in before the "shut down please" request should still finish. Icing would be a way to shutdown with a timeout but I think we can compose that if there is an API for shutdown.

The way I envision usage of such an API

```rust

```

repnop:
Definitely agree with @mehcode. Having the ability to gracefully shutdown the server is pretty important for me as well. :+1:

ririsoft:
FWIW here is how I do it using the [async-ctrlc](https://crates.io/crates/async-ctrlc) crate:

```rust
async fn run() -> Result<()> {
    let ctrlc = async {
        CtrlC::new().expect("Cannot use CTRL-C handler").await;
        println!("termination signal received, stopping server...");
        Ok(())
    };

    let app = async {
        let app = tide::new();
        app.listen("localhost:8080").await
    };

    app.race(ctrlc).await?;
    println!("server stopped.");
    Ok(())
}
```

Please note that this will not wait for all running requests to complete, this is a "brutal graceful shutdown", if I dare. But at least it handles SIGTERM and CTRL-C shutdown (cross platform) which is better than nothing.

repnop:
[...]

PR #746, opened 2020-11-18
Graceful shutdown2

GWBasic:
Fixes https://github.com/http-rs/tide/issues/528

It works by awaiting on two futures and returns when the first completes: https://github.com/http-rs/tide/compare/main...GWBasic:GracefulShutdown2?w=1#diff-1f14b8b6953e2388b19efaf814862a89c0b57c45a6814f79ed373fde05d864d0R82

I then created a "CancelationToken" future that returns once it is canceled.

(For context, see https://github.com/http-rs/tide/issues/528#issuecomment-723601893)


GWBasic:
Note: The syntax might be cleaner using the select macro: https://rust-lang.github.io/async-book/06_multiple_futures/03_select.html

GWBasic:
Just an FYI that I created a crate with CancelationToken. Again, looking for some feedback!

Cargo.toml
```diff
@@ -41,6 +41,7 @@ async-sse = "4.0.1"
 async-std = { version = "1.6.5", features = ["unstable"] }
 async-trait = "0.1.41"
 femme = { version = "2.1.1", optional = true }
+futures = "0.3.7"
```
yoshuawuyts, line 44: We probably don't want to depend on `futures` directly since it adds about a minute of compile time in certain setups. Instead it's preferable to use smaller crates and/or async-std directly.

src/cancelation_token.rs
```diff
@@ -0,0 +1,60 @@
...
+	}
+}
+
+impl Clone for CancelationToken {
+	fn clone(&self) -> Self {
+		CancelationToken {
+			shared_state: self.shared_state.clone()
+		}
+	}
+}
```
yoshuawuyts, line 60: The [stop-token](https://docs.rs/crate/stop-token/0.1.2) does much the same, and we're looking to integrate it into async-std in the near future. If possible it'd be great to use that crate directly instead of defining our own here.

yoshuawuyts (changes requested):
Hey @GWBasic, thanks so much for kicking this off!

In broad strokes it seems what you're implementing is in the right direction; but we could probably use a round of design feedback on this. Usually I'd suggest opening an issue first, but given we're already discussing this in a PR, doing that here seems fine.

Specifically I want to get a better idea of how we expect end-users to register cancellation in Tide, and how that will work with the various listeners. cc/ @jbr I think you may have had a few ideas here?

src/cancelation_token.rs
GWBasic: One thing I like about stop-token is that it supports streams.

However, the authors describe it as "experimental." Upon closer inspection:

- The documentation doesn't build: https://docs.rs/crate/stop-token/0.1.2
- The tests are limited: https://github.com/async-rs/stop-token/blob/master/tests/tests.rs
- Based on how I interpret https://github.com/async-rs/stop-token/blob/master/src/lib.rs#L170, canceling via stop token cancels the next read from the stream. (IE, if I cancel NOW, the background task doesn't cancel until there's another item in the stream.)

I should point out that the library that I made with cancelation token has full documentation and tests: https://crates.io/crates/sync-tokens

Cargo.toml
GWBasic: Futures is required for "select": https://rust-lang.github.io/async-book/06_multiple_futures/03_select.html

Honestly, I don't see a way to do this without "select," otherwise you end up waiting for a final incoming http request before ending the loop accepting incoming requests. (See my comment about when I reviewed stop-token.)

Maybe it's worth trying to implement something like select manually, but, IMO, that might be a fools' errand.

GWBasic:
@yoshuawuyts : Regarding how end users can register for cancelation, take a look at the example code I published here: https://docs.rs/sync-tokens/0.1.0/sync_tokens/

Pay attention to main:

1. The caller knows when the server is listening
2. The caller explicitly calls cancel
3. The caller can await for the server to complete

Cargo.toml
Fishrock123: I think what we would prefer is [`futures_lite::future::race`](https://docs.rs/futures-lite/1.11.2/futures_lite/future/fn.race.html).

GWBasic: Great advice!

Let me see what I can do. It'll probably be a few days before I can update this PR.

GWBasic:
Replaced with https://github.com/http-rs/tide/pull/766 due to excessive merge conflicts

PR #749, opened 2020-11-23
Add more `ToListener` implementations

### LINKED ISSUE #744: Implement `ToListener` for `(&String, u16)`
When passing an `&String` to `Server::listen` the following error occurs:

```txt
the trait bound `(&std::string::String, u16): tide::listener::ToListener<server::State>` is not satisfied
the following implementations were found:
  <(&str, u16) as tide::listener::ToListener<State>>
```

We should fix this by adding an additional `ToListener` impl. This matches the socket constructor impl added in Rust 1.46.

rossmacarthur:
- Implements `ToListener` for  `&String`.
- Implements `ToListener` for `(String, u16)` and `(&String, u16)`.

Resolves #744

### DIFF
--- a/src/listener/to_listener_impls.rs
+++ b/src/listener/to_listener_impls.rs
@@ -61,6 +61,16 @@ where
     }
 }
 
+impl<State> ToListener<State> for &String
+where
+    State: Clone + Send + Sync + 'static,
+{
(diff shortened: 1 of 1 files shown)


yoshuawuyts (approved):
This is great; thanks so much!

PR #753, opened 2020-11-24
docs: do not inline docs from deps that drift

Fishrock123:
Inlining these docs causes them to be perpetually out-of-date, since Docs.rs only builds on the actual containing crate publish.


Fishrock123:
Similar PR to Surf: https://github.com/http-rs/surf/pull/269

PR #755, opened 2020-11-27
remove the logger from the default middleware

jbr:
Not a large code change, but probably still worth discussing. Here's a PR for that discussion

### DIFF
--- a/examples/catflap.rs
+++ b/examples/catflap.rs
@@ -4,6 +4,7 @@ async fn main() -> Result<(), std::io::Error> {
     use std::{env, net::TcpListener, os::unix::io::FromRawFd};
     tide::log::start();
     let mut app = tide::new();
+    app.with(tide::log::LogMiddleware::new());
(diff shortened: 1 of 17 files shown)


yoshuawuyts (approved):
I'm a bit torn on this PR, but I think it may be a bullet we have to bite.

I've always wanted logging to be somewhat turn-key, but we may have factored it wrong. I wonder if the better solution long-term may be:

```rust
// current
tide::log::start();
let mut app = tide::new();
app.with(tide::log::LogMiddleware::new());

// proposed
let mut app = tide::new();
app.with(tide::log::LogMiddleware::new());
```

Where `LogMiddleware` has the ability to start the logger, only `log` without starting a logger, take different log producers, and other configurations. Logging would then always become opt-in, easily to replace with custom solutions at various layers, but still be easy to configure. Possibly based on `tracing` as well.

But all in all this patch is required to bring us there, so no objections if we want to merge this in the interim.

PR #756, opened 2020-12-04
ci: run clippy, docs and fmt checks on stable

mendelt:
The build was failing for a while because clippy, doc-tests and cargo fmt are run on rust nightly. Unfortunately the nightly build does not always include all these components. In this case cargo fmt was failing.
It used to be necessary to run clippy on nightly so I assume this is why these steps were run on nightly for the Tide build. This is not necessary anymore. This pr changes the build script to run these steps on rust stable.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -76,11 +76,6 @@ jobs:
     steps:
     - uses: actions/checkout@master
 
-    - uses: actions-rs/toolchain@v1
-      with:
-        toolchain: nightly
-        override: true
-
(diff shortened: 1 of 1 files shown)


mendelt:
I just noticed this has 100% overlap with https://github.com/http-rs/tide/pull/736 which is a bit more extensive

yoshuawuyts (approved):
#736 is currently failing; I believe it can be rebased on this PR if required, and we do need to fix some of these things. With that in mind I'm going to go ahead and merge this

Issue #728, opened 2020-10-27
serve_dir() does not automatically decode URL-encoded file paths


zelda-at-tempus:
I have an app that serves static files that have spaces in their names i.e. the file path would look like this on linux:
`/app/static/media/some name with spaces.txt`

When my website requests the file from `tide`, it URL-encodes the path, changing spaces to `%20` like so:
`http://localhost/static/media/some%20name%20with%20spaces.txt`

I'm currently working around this by renaming files, but I wanted to know if this was expected behavior or not.

I imagine that this would cause filenames with certain symbols to also be unservable in the same way

jbr:
Thanks for filing this. Seems like a bug to me!

mendelt:
I was looking at this issue but I don't think we url-decode anything in Tide anywhere right now? Not route parameters, not query parameters. Is this something we want to build in?

Anglebrackets:
I see this has been open since 2020.  I don't know if anyone will read this comment, but it is still a problem. It means I have to implement serve_dir() myself. Is that the approach that others have taken?

PR #757, opened 2020-12-04
Percent-decode url parameters

mendelt:
This pr fixes a bug where the `param` method on a `Request` would supply the raw value of the parameter instead of the url-decoded or percent-decoded value.
This does change the Tide public API because the `param` method now returns a result of 'String' instead of a '&str'. Returning a reference after conversion would not make much sense anymore.
It also adds a crate-dependency to `percent-decode` version 2.1.0. This was already an indirect dependency, this crate is used via (among others) http-types -> serde_urlencoded -> form_urlencoded to decode query parameters.

decoding is done on-demand when a user of Tide requests a parameter. The `Params` struct still stores parameters encoded. We could also encode eagerly when storing parameters. This could also allow us to change the API back to returning &str. The reason I did things this way is to avoid decoding for unused parameters. (don't pay for what you don't use) the downside is that if a user calls .param multiple times the parameter will be decoded multiple times.

I tested this for parameters from : wildcards and * wildcards that match multiple path segments. I can imagine there might be some edge-cases that were missed. So let me know if you can think of any.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -1,5 +1,6 @@
 use async_std::io::{self, prelude::*};
 use async_std::task::{Context, Poll};
+use percent_encoding::percent_decode_str;
 use route_recognizer::Params;
 
 use std::ops::Index;
@@ -282,7 +283,7 @@ impl<State> Request<State> {
     /// use tide::{Request, Result};
     ///
     /// async fn greet(req: Request<()>) -> Result<String> {
-    ///     let name = req.param("name").unwrap_or("world");
+    ///     let name = req.param("name").unwrap_or("world".to_string());
     ///     Ok(format!("Hello, {}!", name))
     /// }
     ///
@@ -293,11 +294,12 @@ impl<State> Request<State> {
     /// #
     /// # Ok(()) })}
     /// ```
-    pub fn param(&self, key: &str) -> crate::Result<&str> {
+    pub fn param(&self, key: &str) -> crate::Result<String> {
         self.route_params
             .iter()
             .rev()
             .find_map(|params| params.find(key))
+            .map(|param| percent_decode_str(param).decode_utf8_lossy().into())
             .ok_or_else(|| format_err!("Param \"{}\" not found", key.to_string()))
     }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ http-client = { version = "6.1.0", default-features = false }
 http-types = { version = "2.10.0", default-features = false, features = ["fs"] }
 kv-log-macro = "1.0.7"
 log = { version = "0.4.13", features = ["kv_unstable_std"] }
+percent-encoding = "2.1.0"
 pin-project-lite = "0.2.0"
 route-recognizer = "0.2.0"
 serde = "1.0.117"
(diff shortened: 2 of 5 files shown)


mendelt:
I also added decoding to `ServeDir` so this pr now fixes https://github.com/http-rs/tide/issues/728

src/request.rs
```diff
@@ -293,11 +294,12 @@ impl<State> Request<State> {
     /// #
     /// # Ok(()) })}
     /// ```
-    pub fn param(&self, key: &str) -> crate::Result<&str> {
+    pub fn param(&self, key: &str) -> crate::Result<String> {
         self.route_params
             .iter()
             .rev()
             .find_map(|params| params.find(key))
+            .map(|param| percent_decode_str(param).decode_utf8_lossy().into())
```
yoshuawuyts, line 302: If a malformed URL is passed, perhaps we should return a 400 error rather than apply lossy encoding? I would assume that's a case of an invalid URL being passed?

Or do we at this point already know this is valid ASCII so we can continue without a problem?

yoshuawuyts (approved):
One question, but this looks good otherwise. Feel free to merge once that's been addressed.

src/request.rs
mendelt: Good point, I'll think a bit more about error handling here and maybe document what we expect to get as input/output.

If we want to return an error here I'm not sure if we should return a 400, or maybe a more specific error that a user can handle?

yoshuawuyts: > If we want to return an error here I'm not sure if we should return a 400, or maybe a more specific error that a user can handle?

By user, you mean a user of Tide or a user making a request? In this case I think it's fair to say that the URL sent in the request is invalid, and the server in fact cannot parse it. I'm not sure how else we could handle this on the Tide side.

mendelt: I meant a Tide user. Maybe someone wants some more flexibility around url-parsing-error-handling than just shortcutting to a 400-response. But after looking at the error type we use I don't think we're really set up for that? I think our error type is designed more for reporting errors  to end-users than API-users. So I think that will be a discussion for another time.

I'll fix this to return a 400.

Thanks for the feedback!

mendelt:
I might need some feedback here;

After looking at this again decided it might be better to do percent decoding eagerly instead of on-demand like this pull request. This has a couple of benefits;
- Errors with malformed URLs can be returned immediately with an HTTP 401 before the users endpoint code is called. In this pr we would return an error to the calling endpoint which would need to be handled.
- No Tide API change will be neccesary. In this pr we return a decoded String instead of an &str reference to the stored parameter. We need this because we do decoding on-the-fly. If we decode whenever a request comes in and store the decoded parameters we can return a &str reference like before.

I'm still unsure if we should do the decoding in Tide, I have a branch where this is implemented here: https://github.com/mendelt/tide/tree/percent_decode_params and I can create a PR for that or if we should push this into `route_recognizer`. Which would leave `Tide` unchanged. This would add a dependency from `route_recognizer` to the percent-encoding crate with all the serde stuff that comes with it.

mendelt:
I solved this problem in a different way in https://github.com/http-rs/tide/pull/799.

mendelt:
I think https://github.com/http-rs/tide/pull/799 is a better option for percent decoding. Closing this one.

PR #764, opened 2020-12-12
Fix port in `curl` invocation in example.

jswrenn:
The example server binds port `8080`:
https://github.com/http-rs/tide/blob/46bf28a13be9951e7a61f647f98d5232106d61fb/src/lib.rs#L36

...but the example `curl` invocations post to port `8000`:
https://github.com/http-rs/tide/blob/46bf28a13be9951e7a61f647f98d5232106d61fb/src/lib.rs#L47-L51

This patch modifies the `curl` invocations to post to port `8080`, instead.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -44,10 +44,10 @@
 //! ````
 //!
 //! ```sh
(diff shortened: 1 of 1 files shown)


yoshuawuyts (approved):
Oh, great catch -- thanks so much!

PR #765, opened 2020-12-14
Update version numbers of deps for README example

jmn:
The example code doesn't build with the old versions anymore.


yoshuawuyts (approved):
Ah yes, great catch. Thanks so much!

PR #769, opened 2020-12-26
Fix #762

### LINKED ISSUE #762: Session regeneration is ignored
SessionMiddleware seems to ignore session regeneration. The middleware should reset the session cookie with a new id after `session.regenerate()`, but instead it continues to use the old cookie. This can be a security vulnerability because it allows session fixation attacks.

Here is a simple test case. The session id should change after logging in, but it doesn't.

```rust
use tide::sessions::{MemoryStore, SessionMiddleware};
use tide::{Redirect, Request, Response};

#[async_std::main]
    let mut app = tide::new();

    app.with(SessionMiddleware::new(
        MemoryStore::new(),
        b"01234567890123456789012345678901",
    ));

    app.at("/").get(index);
    app.at("/login").get(login);

    app.listen("127.0.0.1:8081").await?;
    Ok(())
}

async fn index(req: Request<()>) -> tide::Result {
    let session = req.session();
    let user: String = session.get("user").unwrap_or("guest".to_string());

    let mut resp: Response = format!(
        r#"
        Logged in as {}<br>
        Session id: {}<br>
        <a href="/login">Login</a>"#,
        user,
        session.id()
    )
    .into();

    resp.set_content_type("text/html");
    Ok(resp)
}

async fn login(mut req: Request<()>) -> tide::Result {
    let session = req.session_mut();
    session.regenerate();
    session.insert("user", "admin")?;
    Ok(Redirect::see_other("/").into())
}
```

pepoviola:
Hi @yoshuawuyts  / @No9,
                                              I make this changes trying to fix the bug reported in #762, where the `cookie id` isn't updated ( resent ) when the session is regenerated. I put the `session` behind an `Arc::RwLock` ( std ) when we set the `ext` and then in the request return the `guard` to don't change the `req` api. ( e.g req.session() /  req.session_mut() ).  It's ok to use the RwLock from `std::sync` ? 

Also, I wonder if we still need the `RwLock` and the `Arc` in the `Session` struct  ( https://github.com/http-rs/async-session/blob/main/src/session.rs#L60-L67 )  if we already are using a mutex ( from this pr )  on top to of that struct ?

Open here as draft since I'm not sure if my approach is ok, and the code have this `clippy` warnings that I'm not sure how to resolve.

```
519 |     pub fn session_mut(&mut self) -> std::sync::RwLockWriteGuard<crate::sessions::Session> {
    |                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: indicate the anonymous lifetime: `std::sync::RwLockWriteGuard<'_, crate::sessions::Session>`
```

Thanks! any feedback is welcome :)

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -505,9 +505,8 @@ impl<State> Request<State> {
         let lock = self.ext::<Arc<RwLock<crate::sessions::Session>>>().expect(
             "request session not initialized, did you enable tide::sessions::SessionMiddleware?",
         );
-        let guard = lock.read().unwrap();
 
-        guard
+        lock.read().unwrap()
     }
 
     /// Retrieves a mutable reference to the current session.
@@ -521,9 +520,8 @@ impl<State> Request<State> {
         let lock = self.ext::<Arc<RwLock<crate::sessions::Session>>>().expect(
             "request session not initialized, did you enable tide::sessions::SessionMiddleware?",
         );
-        let guard = lock.write().unwrap();
 
-        guard
+        lock.write().unwrap()
     }
 
     /// Get the length of the body stream, if it has been set.
--- a/src/request.rs
+++ b/src/request.rs
@@ -2,6 +2,7 @@ use async_std::io::{self, prelude::*};
 use async_std::task::{Context, Poll};
 use route_recognizer::Params;
 
+use std::sync::{Arc,RwLock};
 use std::ops::Index;
 use std::pin::Pin;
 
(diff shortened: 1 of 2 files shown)


jbr:
I'd be happy to talk through the design considerations that went into the current approach. It's an intentional aspect of the design that we drop the cookie value on clone

pepoviola:
Hi @jbr, thanks for the reply. Yes, that would we great! As I say I'm not sure if the fix is in the correct path, I also realize that the test in async-session expect that the cookie gets dropped.

Thanks for the feedback!🙌

pepoviola:
Hi @jbr,  If you have time can we make a quick talk to understand the design and how to resolve the `regenerate` issue. Maybe is better to have an struct Session that holds an inner struct with a Arc<RwLock> but i'm not sure.

Thanks in advance!

PR #776, opened 2021-01-11
Update log to use kv_unstable_std instead of std

KodrAus:
Part of https://github.com/rust-lang/log/issues/437

The `log` crate has an unstable structured logging API under the `kv_unstable` feature. In previous releases, if you specified both the `kv_unstable` and `std` features of `log` like so:

```toml
log = { features = ["std", "kv_unstable"]}
```

you'd get support for standard library types in `log`'s structured logging API.

Going forward, this functionality is now gated under `kv_unstable_std`:

```toml
log = { features = ["kv_unstable_std"]}
```

This change was made because we need to enable features in optional dependencies when both the `std` and `kv_unstable` features are enabled, which isn't currently supported by Cargo.

This PR updates this library to follow the new approach. It can be merged at any time and is currently non-blocking, but on 2020-01-18 the version of `log` requiring `kv_unstable_std` instead of `kv_unstable` and `std` will be published.

Thanks for trying out `log`'s structured logging API and sorry for any disruption! :bow:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ futures-util = "0.3.6"
 http-client = { version = "6.1.0", default-features = false }
 http-types = "2.5.0"
 kv-log-macro = "1.0.7"
-log = { version = "0.4.11", features = ["std"] }
+log = { version = "0.4.13", features = ["kv_unstable_std"] }
 pin-project-lite = "0.1.10"
 route-recognizer = "0.2.0"
 serde = "1.0.117"


jbr:
i believe this closes #787

KodrAus:
It should, yes 👍

It was unfortunate to have to break but I’m trying to avoid any more until we can stabilize the API proper.

pepoviola:
Hi @KodrAus, thanks for the pr! I just resolve the conflict. ping @yoshuawuyts  I think is ready to merge :)
Thx!

Fishrock123 (approved):

yoshuawuyts:
Fixed in `v0.15.1`. Release notes: https://github.com/http-rs/tide/releases/tag/v0.15.1

PR #779, opened 2021-01-19
Request: make missing param a 400 Bad Request error

Fishrock123:
Previously was a 500 Internal Server Error, which is incorrect since this is a client url error location.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -9,7 +9,7 @@ use std::pin::Pin;
 use crate::cookies::CookieData;
 #[cfg(feature = "cookies")]
 use crate::http::cookies::Cookie;
-use crate::http::format_err;
+use crate::http::format_err_status;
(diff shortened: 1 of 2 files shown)


pepoviola:
Hi @Fishrock123, I remember to ask @yoshuawuyts ( in discord https://discord.com/channels/598880689856970762/649056551835009097/761552491031822356 ) about this and the `500` is because is trying to extract a `param` from a route that don't have that param.

Thanks!

Fishrock123:
It occurred later to me that most of these should be 404 and not 400.

This site should probably be 400 but it should be paired with another that fixes the above mentioned issue to be 404.

mendelt:
Not sure if I understand this correctly but if you define a route "/some/:route/" and then try to extract the parameter "bla" that seems like a bug in your program, not an invalid url. So a server error seems to be right for that.

PR #780, opened 2021-01-22
Reduce dependencies when cookies not enabled

joshtriplett:
Upgrade to http-types 2.10 with optional support for cookies, and make
use of that support to reduce dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ rustdoc-args = ["--cfg", "feature=\"docs\""]
 
 [features]
 default = ["h1-server", "cookies", "logger", "sessions"]
-cookies = []
+cookies = ["http-types/cookies"]
 h1-server = ["async-h1"]
 logger = ["femme"]
 docs = ["unstable"]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -41,7 +41,7 @@ async-trait = "0.1.41"
 femme = { version = "2.1.1", optional = true }
 futures-util = "0.3.6"
 http-client = { version = "6.1.0", default-features = false }
-http-types = "2.5.0"
+http-types = { version = "2.10.0", default-features = false }
```
yoshuawuyts, line 44: This also disables the `async-std` feature, which we rely on for `Body::from_file`. We should probably re-enable this feature.

joshtriplett: Strange! I did test this with `cargo test --no-default-features --features h1-server,logger`, which build-tested examples that call `Body::from_file`. It looks like something *else* is pulling in the `http-types` "default" feature.

I'll fix `tide` to require the `fs` feature of `http-types`, and also look into what else in the dependency tree is still pulling in `http-types` with default features enabled.

joshtriplett:
Updated to ensure `Body::from_file` still works, which requires the `fs` feature of `http-types`. Right now, that feature is being enabled by other crates in the dependency tree, but since tide uses it directly, tide should depend on it.

Fishrock123 (approved):

PR #784, opened 2021-01-24
Add simple reverse proxy example

silvioprog:
I found this example useful in understanding how a tide-based server can handle reverse proxy.

### DIFF
--- a/examples/proxy.rs
+++ b/examples/proxy.rs
@@ -1,3 +1,6 @@
+// WARNING: Don't not use it in production!
+// This is just an example, so there is much more to a correct secure reverse proxy implementation.
+//
 // Example: HTTP GET to http://localhost:8080/http-rs/tide
 // $ curl "http://localhost:8080/http-rs/tide"
 // I'll show the Tide page at GitHub
--- a/examples/proxy.rs
+++ b/examples/proxy.rs
@@ -1,0 +1,28 @@
+// Example: HTTP GET to http://localhost:8080/http-rs/tide
+// $ curl "http://localhost:8080/http-rs/tide"
+// I'll show the Tide page at GitHub
+#[async_std::main]
+async fn main() -> Result<(), std::io::Error> {
+    tide::log::start();
+    let mut app = tide::new();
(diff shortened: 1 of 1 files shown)


darkdarkfruit:
Good one

jbr:
I think we probably don't want to include an example like this, as a properly functioning reverse proxy has security implications and also more surface area than it appears. Here's another example of a reverse proxy in tide, but I imagine there are probably flaws with that one too and would not consider it an official example https://github.com/jbr/tide-serve/blob/main/src/forward_middleware.rs

My concern is that an example indicates recommended / best practices, and writing a safe reverse proxy for tide is something that should be a well-tested crate of its own, not something we encourage people to do in their own code. If folks need an example for writing a reverse proxy in tide, they probably shouldn't be writing a reverse proxy in tide.

jbr:
I think before merging anything like this, the example should have a large and clear warning not to use this in production. There is more to a correct reverse proxy implementation than this example, and it should be a standalone crate. This is a security-sensitive feature

silvioprog:
> I think before merging anything like this, the example should have a large and clear warning not to use this in production.

Done!

DavidBM:
I was able to make it work with:

```rust
fn bypass_to<State: Send + 'static>(
    method: tide::http::Method,
    url: tide::http::Url,
) -> impl Fn(tide::Request<State>) -> Pin<Box<dyn Future<Output = tide::Result> + Send + 'static>> {
    move |request: tide::Request<State>| {
        let url = url.clone();
        Box::pin(async move {
            let request: &tide::http::Request = request.as_ref();

            let mut request = request.clone();

            *request.url_mut() = url;
            request.set_method(method);

            let request = surf::Request::from(request);

            let response = surf::Client::default().send(request).await?;

            Ok(tide::Response::from_res(response))
        })
    }
}
```

and then you can just

```rust
let mut app = tide::new();

app.at("/your route")
        .get(bypass_to(tide::http::Method::Get, tide::http::Url::parse("...").unwrap()));
```

Issue #777, opened 2021-01-15
serve_dir shadows existing path


mankinskin:
serve_dir seems to have precedence to other routes, even if it is specified after them.
So this server:
```rust
let mut server = tide::new();
server.with(TraceMiddleware::new());
server.at("api").get(async move |_| Ok("Api Endpoint"));
server.at("").serve_dir(format!("{}/pkg", CLIENT_PATH)).expect("Cannot serve directory");
server.listen(SocketAddr::from(([0, 0, 0, 0], 8000))).await
```
Shows this log output upon receiving a request to `http://0.0.0.0:8000/api`:
```rust
Jan 15 20:06:32.946  INFO tide::log::middleware: <-- Request received
Jan 15 20:06:32.947  INFO Request{path="/api/"}: tide_tracing: method="GET"
Jan 15 20:06:32.947  INFO Request{path="/api/"}: tide::fs::serve_dir: Requested file: "/home/user/git/web-app/client/pkg/api"
Jan 15 20:06:32.948  WARN Request{path="/api/"}: tide::fs::serve_dir: File not found: PathBuf { inner: "/home/user/git/web-app/client/pkg/api" }
Jan 15 20:06:32.948  INFO Request{path="/api/"}:Response: tide_tracing: status=404
Jan 15 20:06:32.948  INFO Request{path="/api/"}:Response: tide_tracing: duration="1.268409ms"
Jan 15 20:06:32.949  WARN Request{path="/api/"}:Response:Client error: tide_tracing: Response sent
Jan 15 20:06:32.949  WARN tide::log::middleware: Client error --> Response sent
```
And responds with 404.

The expected behavior would be that the endpoint defined first should have precedence and should therefore catch a request to `/api` before `serve_dir`.

u5surf:
I’ve written the same situation test, it failed indeed. 
```
diff --git a/tests/serve_dir.rs b/tests/serve_dir.rs
index 39cf7e2..cc7903d 100644
--- a/tests/serve_dir.rs
+++ b/tests/serve_dir.rs
@@ -3,6 +3,10 @@ use tide::{http, Result, Server};
 use std::fs::{self, File};
 use std::io::Write;
 
+fn api() -> Box<dyn tide::Endpoint<()>> {
+    Box::new(|_| async { Ok("api") })
+}
+
 fn app(tempdir: &tempfile::TempDir) -> Result<Server<()>> {
     let static_dir = tempdir.path().join("static");
[...]

PR #795, opened 2021-02-07
bugfix : serve_dir doesn't precede to other route

u5surf:
Fixes: #777 

I had fixed that `serve_dir` precede to other existing routes. 

It has a little breaking changes that specified routing path of `serve_dir` to add suffix wild card asterisk. 

But unit/integrate test has certainly all passed and it has changed examples as anyone can make sense.


yoshuawuyts (approved):
This looks great; thanks so much!

PR #799, opened 2021-02-14
Percent decode parameters

mendelt:
This pr solves the same problem as https://github.com/http-rs/tide/pull/757 but instead of percent-decoding lazilly (when a user queries a parameter) this does it eagerly, when a request comes in all parameters are decoded.
This has one downside, it's a bit slower. It iterates through all the `Params` in the request, decodes them and stores them in a new `BTreeMap`
But it has a couple of benefits;
- It can return a 401 for a malformed request immediately without involving the users code.
- The api stays the same, no need to return a param as a `String` instead of a `&str`

~~This pr also merges the Vec<Param> into a single BTreeMap<String, String> because it already has to iterate through and copy all the params. This makes other code a bit simpeler.~~

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ http-client = { version = "6.1.0", default-features = false }
 http-types = { version = "2.10.0", default-features = false, features = ["fs"] }
 kv-log-macro = "1.0.7"
 log = { version = "0.4.13", features = ["kv_unstable_std"] }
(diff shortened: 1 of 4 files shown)


mendelt:
Merging the vector of parameter maps did not play well with the strip_prefix feature. The fix for that would probably have been uglier than just keeping the vec.

PR #803, opened 2021-02-21
Set SameSite policy Lax as default

### LINKED ISSUE #801: Consider using SameSite=Lax as default for SessionMiddleware
I don't want to make it very personal but I wasted about 5 hours on a bug that was due to SameSite=Strict that is tide's default. Browsers treat SameSite as Lax if it's not specified, and I think that many developers expect the cookie behavior to be Lax intuitively. For example, when after successful payment on a different domain I'm redirected to my own site I want to send a cookie with my access_token but I can't because of Strict. I'm aware it is in the docs but I somehow managed to not notice it, because I didn't expect the bug to be caused by it.

jlkiri:
Closes #801 
Browsers treat SameSite policy as Lax if it is not specified and this behaviour is more intuitive.

### DIFF
--- a/src/sessions/middleware.rs
+++ b/src/sessions/middleware.rs
@@ -172,7 +172,7 @@ impl<Store: SessionStore> SessionMiddleware<Store> {
             cookie_path: "/".into(),
             cookie_name: "tide.sid".into(),
             cookie_domain: None,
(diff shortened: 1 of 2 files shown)


jbr (approved):
Thanks!

murphysean (approved):
Looks good!

PR #824, opened 2021-04-22
feat: add new lambda feature

Fishrock123:
For interacting with `lambda_http`, part of the official awslabs toolkit.

Needs more real testing. TBD.

This is easier to do in Tide directly because otherwise you have to make a wrapper, which seems unnecessary (though I could be convinced that perhaps this is still better in a separate crate).


ItsiAdam:
This looks good and works completely fine for me. Only gripe I had was that binary is returned for all content types, but I couldn't find a better solution than just switching it out for `lambda_http::Body::Text(body.into_string().await?)`. 

Maybe we could pass the body type as a generic or something? I don't like the idea of checking the headers to find out what type should be returned.

lambda-http 0.4.0 adds explicit lifetimes: 72111315fc58ae60e3d1ccd0e48fb64ab6e4e044

Fishrock123:
I ended up doing this as the `"lambda-http"` for the Tide-wrapping framework I maintain for my employer: https://github.com/eaze/preroll#list-of-optional-add-on-features

ItsiAdam:
I've just had a quick look over `tide-lambda-listener`; by comparison to this solution, it seems unnecessarily complicated? I don't see any difference functionality wise - how come you chose that solution over this?

Fishrock123:
Oh right. This is also available standalone as https://github.com/Fishrock123/tide-lambda-listener, which I completely forgot about.

Issue #825, opened 2021-04-23
The example code on the front page does not return errors


skovmand:
Hi! Thanks for your nice project!

I just tried running the basic example in the repository README.md. 
It compiles and runs well, but the error message given as example is not returned.

This example is given in README.md:

```
$ curl localhost:8080/orders/shoes -d '{ "name": "Chashu", "legs": 4 }'
Hello, Chashu! I've put in an order for 4 shoes

$ curl localhost:8080/orders/shoes -d '{ "name": "Mary Millipede", "legs": 750 }'
number too large to fit in target type
```

The first response works as expected, but I get no response from the second request. It's simply empty with a 422 HTTP Status Code. It seems related to https://github.com/http-rs/tide/issues/739

So perhaps it simply the README.md that needs to be updated (or perhaps a change to the code to return the error).

My curl output (with verbose mode):

```
$ curl localhost:8080/orders/shoes -d '{ "name": "Mary Millipede", "legs": 750 }' -v
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8080 (#0)
> POST /orders/shoes HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.64.1
> Accept: */*
> Content-Length: 41
> Content-Type: application/x-www-form-urlencoded
> 
* upload completely sent off: 41 out of 41 bytes
< HTTP/1.1 422 Unprocessable Entity
< content-length: 0
< date: Fri, 23 Apr 2021 18:52:16 GMT
< 
* Connection #0 to host localhost left intact
* Closing connection 0
```

yoshuawuyts:
Closed by https://github.com/http-rs/tide/pull/868

PR #828, opened 2021-05-03
Fix readme example

0xor1:
Return a useful error message to match the expected value. for issue #825

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ struct Animal {
 
 #[async_std::main]
 async fn main() -> tide::Result<()> {
+    tide::log::start();
     let mut app = tide::new();
     app.at("/orders/shoes").post(order_shoes);
     app.listen("127.0.0.1:8080").await?;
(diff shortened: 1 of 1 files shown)


jbr:
I think the way we want to address this is probably by creating an error handler middleware that ships with tide that we can use with this example, in order to demonstrate the use of `?` in an endpoint. @yoshuawuyts @Fishrock123 do you agree?

Fishrock123:
I'm biased, but I think we should either just ship https://github.com/eaze/preroll/blob/latest/src/middleware/json_error.rs or ship a version which changes it's output based on the `Accept` header (defaulting to json).

leemari (commented):
Shouldn't it return the correct status code though, i.e. BadRequest.

If you run the JSON through serde:

```rust
use serde::{Deserialize, Serialize};
use serde_json::{Result, Error};

#[derive(Deserialize)]
struct Animal {
    name: String,
    legs: u8,
}

fn main() {
    let json = "{\"name\": \"Richard\",\"legs\": 750}";
    let r1: Result<Animal> = serde_json::from_str(json);
    match r1 {
        Ok(t) => println!("Hello, {}! I've put in an order for {} shoes", t.name, t.legs),
        Err(e) => eprintln!("{}", e),
    };
}
```

It returns:

``` bash
invalid value: integer `750`, expected u8 at line 1 column 30
```

Which should, in turn, return BadRequest.

PR #833, opened 2021-06-01
update async-session to 3.0.0

jbr:
this is a patch level change for tide, as it introduces no changes to the public api

also fixes several unrelated clippies. chose to allow clippy::len_without_is_empty because that's part of the public api and any changes to that should be done in a more considered fashion than "trying to fix a clippy lint"

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -57,6 +57,7 @@
 #![forbid(unsafe_code)]
 #![deny(missing_debug_implementations, nonstandard_style)]
 #![warn(missing_docs, unreachable_pub, future_incompatible, rust_2018_idioms)]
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -57,6 +57,7 @@
 #![forbid(unsafe_code)]
 #![deny(missing_debug_implementations, nonstandard_style)]
 #![warn(missing_docs, unreachable_pub, future_incompatible, rust_2018_idioms)]
+#![allow(clippy::len_without_is_empty)]
```
Fishrock123, line 60: I told them about this and I swear it was fixed but :shrug:

Fishrock123 (approved):

PR #834, opened 2021-06-18
Fix response if cookies are not enabled

george-hopkins:
In case Tide is built without support for cookies (specifically `http-types/cookies`), some of the exports used in `Response` are not available. I guess the issue didn't pop up earlier because some other dependencies did enable `http-types/cookies` on their own.

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -4,11 +4,13 @@ use std::ops::Index;
 
 use serde::Serialize;
 
+#[cfg(feature = "cookies")]
 use crate::http::cookies::Cookie;
 use crate::http::headers::{self, HeaderName, HeaderValues, ToHeaderValues};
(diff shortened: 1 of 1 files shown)


src/response.rs
```diff
@@ -335,6 +338,7 @@ impl Response {
     }
 
     /// Insert cookie in the cookie jar.
+    #[cfg(feature = "cookies")]
```
Fishrock123, line 341: Can you please also add:

```rust
#[cfg_attr(feature = "docs", doc(cfg(feature = "cookies")))]
```

To both of the functions?

Fishrock123:
Oh, huh we don’t use that docs thing here at all yet. Guess that should be a separate whole PR.

Fishrock123 (approved):

PR #836, opened 2021-06-25
Proposal: Cancellation

pbzweihander:
Related issue: #528

Related pull request: #766

I made `StopSource` and `StopToken` types similar to the API described in https://github.com/yoshuawuyts/notes/blob/7ecb13a9999ae1fccd366ef0e246be5ca78867c0/rust/cancellation.md and used in `tcp_listener` and `unix_listener` modules.

But I'm not sure it actually 'gracefully' shuts down the server...


jbr:
Is there a reason we wouldn't use [stop-token](https://github.com/async-rs/stop-token) (or [stopper](https://github.com/stopper)])?

pbzweihander:
stopper seems great! I was worried that stop-token states in the documentation that it is 'Experimental'.

jbr:
Two things that we'll want to address in addition to stopping the tcp stream: we will need to keep track of how many outstanding requests there are so we know when to shut down the server, and we will need to cut off keep alive connections. Currently, async-h1 will hold onto the tcp stream and wait for a subsequent request, which would delay shutdown by the keepalive timeout

jbr:
As far as stop-token vs stopper, I don't have a great understanding of the current state of development of stop-token, which is part of why I wrote stopper. I also needed it to be runtime independent, but tide can depend on async-std

pbzweihander:
I made a draft PR in async-h1: https://github.com/http-rs/async-h1/pull/188

And it seems working in [this test](https://github.com/http-rs/tide/pull/836/commits/95ed038442a9f80a2f8bb48bd6753b94229cb1da#diff-0fa46cf13157406190efda6bdb7c4bfcc2ab5d95dff884eaaf42d73f1ea8d509). (It fails without [the last commit](https://github.com/http-rs/tide/pull/836/commits/aac7e98469eb420f4bb19546a4fed59ac1237337))

src/listener/tcp_listener.rs
```diff
@@ -111,10 +133,19 @@ where
...
                 }
             };
         }
+
+        let join_handles = std::mem::take(&mut self.join_handles);
+        join_handles
+            .into_iter()
+            .collect::<FuturesUnordered<task::JoinHandle<()>>>()
+            .collect::<()>()
+            .await;
```
pbzweihander, line 147: I don't think it is ideal... Any suggestions?

pbzweihander:
I replaced vector of JoinHandles with [waitgroup](https://crates.io/crates/waitgroup)

jbr:
Waitgroup looks a lot like how I did this in my other framework, with very minor differences (they're using an arc, I'm using an atomicusize which lets me keep track of how many outstanding requests there are): https://github.com/trillium-rs/trillium/blob/main/server-common/src/clone_counter.rs

I'd be totally fine copying mine over into tide, but waitgroup seems good and less code is certainly preferable

laizy:
@jbr waitgroup author here, just for technical discussion.  :)
> they're using an arc, I'm using an atomicusize which lets me keep track of how many outstanding requests there are

1. Actually, Arc's strong_count can track the number of outstanding tasks, so another atomicusize is not needed.
2. Using Arc's strong_count has another benefit: waitgroup will be waked up only once when all outstanding tasks are finished, instead of waked up every time some task is finished.

PR #840, opened 2021-08-09
Document security of serve_dir router

halvko:
As someone new to the library, I worry about the security builtin routes, and here I had to check the (well documented) source to find what i was looking for. This PR adds the information to the route documentation itself.

### DIFF
--- a/src/route.rs
+++ b/src/route.rs
@@ -118,6 +118,12 @@ impl<'a, State: Clone + Send + Sync + 'static> Route<'a, State> {
     /// Each file will be streamed from disk, and a mime type will be determined
     /// based on magic bytes.
     ///
+    /// # Security
+    /// 
+    /// This handler ensures no folders outside the requested folder can be served,
+    /// and attempts to access such files will return StatusCode::Forbidden to the
+    /// caller 
+    ///
     /// # Examples
     ///
     /// Serve the contents of the local directory `./public/images/*` from
--- a/src/route.rs
+++ b/src/route.rs
@@ -119,7 +119,7 @@ impl<'a, State: Clone + Send + Sync + 'static> Route<'a, State> {
     /// based on magic bytes.
     ///
     /// # Security
-    /// 
+    ///
     /// This handler ensures no folders outside the specified folder can be
     /// served, and attempts to access any path outside this folder (no matter
     /// if it exists or not) will return `StatusCode::Forbidden` to the caller.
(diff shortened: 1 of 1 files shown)


jbr (approved):
Seems reasonable to me. Do we 406 in response to paths outside of the served first whether or not there's a file at the path? If so, we might want to clarify that. (And if not, that's a bug)

src/route.rs
```diff
@@ -118,6 +118,12 @@ impl<'a, State: Clone + Send + Sync + 'static> Route<'a, State> {
     /// Each file will be streamed from disk, and a mime type will be determined
     /// based on magic bytes.
     ///
+    /// # Security
+    /// 
+    /// This handler ensures no folders outside the requested folder can be served,
+    /// and attempts to access such files will return StatusCode::Forbidden to the
+    /// caller
```
Fishrock123, line 125:
```suggestion
    /// caller.
```

src/route.rs
```diff
@@ -118,6 +118,12 @@ impl<'a, State: Clone + Send + Sync + 'static> Route<'a, State> {
     /// Each file will be streamed from disk, and a mime type will be determined
     /// based on magic bytes.
     ///
+    /// # Security
+    /// 
+    /// This handler ensures no folders outside the requested folder can be served,
+    /// and attempts to access such files will return StatusCode::Forbidden to the
```
Fishrock123, line 124:
```suggestion
    /// and attempts to access such files will return `StatusCode::Forbidden` to the
```

halvko:
> Seems reasonable to me. Do we 406 in response to paths outside of the served first whether or not there's a file at the path? If so, we might want to clarify that. (And if not, that's a bug)

Yes we do, I'll add it to the PR

src/route.rs
halvko: Oops, thanks ^.^

src/route.rs
halvko: Thanks

halvko:
@jbr I was a little quick there, we 403: Forbidden - which i believe is correct behavior. From http-types:
403 Forbidden: "The client does not have acces rights to the content [...]"
406 NotAcceptable: "[... the server] doesn't find any content that conforms to the criteria given by the user agent"

halvko:
> @halvko Could you rebase to pick up clippy lint fixes from `main`?

Is now done

Fishrock123:
@halvko Sorry, one more thing - could you please run `cargo fmt --all`?

halvko:
Should now be fixed

Fishrock123:
Thanks!

PR #841, opened 2021-08-10
Document behaviour of nested overlapping routes

halvko:
I was surprised by this behavior, so I thought it useful to add it to the documentation

### DIFF
--- a/src/route.rs
+++ b/src/route.rs
@@ -98,6 +98,31 @@ impl<'a, State: Clone + Send + Sync + 'static> Route<'a, State> {
 
     /// Nest a [`Server`] at the current path.
     ///
+    /// # Note
+    ///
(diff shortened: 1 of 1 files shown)


src/route.rs
```diff
@@ -98,6 +98,31 @@ impl<'a, State: Clone + Send + Sync + 'static> Route<'a, State> {
 
     /// Nest a [`Server`] at the current path.
     ///
+    /// # Note
+    ///
+    /// The outer server *always* have precedence when disambiguating
```
Fishrock123, line 103:
```suggestion
    /// The outer server *always* has precedence when disambiguating
```

halvko:
Rebased to version with clippy fixes

halvko:
Thanks a lot for fixing my grammar ^^

PR #851, opened 2021-10-01
Environment Variable to set logging level

modulated:
Most other logging frameworks allow you to set the log level using an env var.
This update preserves the original functionality if no env var is set.


modulated:
Hey thanks for taking the time to respond. That's a fair enough explanation. 

Really liking using this framework.

PR #864, opened 2021-12-31
further async-sqlx-session support

stappersg:
async-sqlx-session also supports mysql and ready for further support on database engines


jbr:
Thanks!

PR #867, opened 2022-01-15
docs(readme): correct surfer link

Haroenv:
This confused me initially, as I thought https://github.com/http-rs/surf was meant with the title. On another note, as surf is also related, maybe it should be linked too?


yoshuawuyts (approved):
bless, thank you!

PR #870, opened 2022-01-22
Fix the sample output in the README example

crodjer:
The code was fixed in https://github.com/http-rs/tide/pull/868/, but the sample output is still incorrect.


yoshuawuyts (approved):
Bless, thank you

PR #872, opened 2022-02-07
Add tide-diesel to middlewares

bluemoon:
Adding my package `tide-diesel` to middlewares for easier discoverability


jbr:
Does this address the the issue that diesel does blocking IO and is unsuitable for using directly inside of tide?

PR #874, opened 2022-02-22
Yew Server-side Rendering with tide & surf

zzy:
This example demonstrates yew server-side rendering with tide & surf, use the GitHub REST API.


jbr:
Does this example need to live within the tide repository? The link in the readme seems sufficient to serve as an example

Issue #863, opened 2021-12-30
dinamycally check allowed origins (or with a regex)


opensas:
I'd like to configure cors to allow every domain from `http://<what.ever.you.wish>.mydomain.com`. In that case I would need to echo back the `Origin` header into `ACCESS_CONTROL_ALLOW_ORIGIN`.

I had a look at the [source code](https://github.com/http-rs/tide/blob/main/src/security/cors.rs#L190) but it doesn't seem to be supported.

Is there a way, or any kind of workaround, to achieve such a thing? Is a feature like this planned for the future?

opensas:
:clap: :clap: :clap: :clap: :clap: :clap:

PR #879, opened 2022-03-16
Allow cors origin to be dynamically matched

marcoslopes:
Hi team, thanks for all the work you guys are doing towards tide.

This PR is for issue https://github.com/http-rs/tide/issues/863.

Our use case is that vercel dynamically generates domains for previewing the changes and our test environment is using tide.

There were a couple of decisions that I am happy to rectify if you see any issues.

1.) That it was okay to bring another dependency, in this case latest [regex](https://github.com/http-rs/tide/compare/main...marcoslopes:main#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R51)

2.) Implementing PartialEq and Hash manually is ok, and decided to allow the most permissive implementation, by that I mean, the contents of the Origin variants is what is actually checked, and not only the variants themselves.

Thank you very much in advance.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,7 @@ pin-project-lite = "0.2.0"
 serde = "1.0.117"
 serde_json = "1.0.59"
 routefinder = "0.5.0"
+regex = "1.5.5"
 
 [dev-dependencies]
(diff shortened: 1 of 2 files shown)


Fishrock123:
Looks good to me

Fishrock123 (approved):

PR #881, opened 2022-04-11
update femme logger

placrosse:


joshtriplett:
AFAICT we're not using any methods from the newer version, and the newer version is a compatible update, so we don't need to upgrade the version requirement in `Cargo.toml`.

PR #885, opened 2022-04-27
Make several features non-default

joshtriplett:
Since we're in the 0.17 beta, we can make breaking API changes, such as
making previously default features non-default.

Leave `h1-server` on by default, since tide isn't an HTTP server without
it. Disable all other optional features by default, and let the user opt
into them as needed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ features = ["docs"]
 rustdoc-args = ["--cfg", "feature=\"docs\""]
 
 [features]
-default = ["h1-server", "cookies", "logger", "sessions"]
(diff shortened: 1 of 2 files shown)


Fishrock123:
I suspect there must be more docs adjustments to be made?

joshtriplett:
On Mon, May 09, 2022 at 05:56:18PM -0700, Jeremiah Senkpiel wrote:
> I suspect there must be more docs adjustments to be made?

I dug through the documentation and didn't find any other cases that
depended on these features being enabled by default.

Fishrock123 (approved):

PR #887, opened 2022-04-27
Update version numbers in README and docs to 0.17

joshtriplett:


Fishrock123:
this will cause the need for a lot of crates to be updated and should be done with caution and probably only when necessary

joshtriplett:
@Fishrock123 The current version of the crate has already updated to 0.17 (not yet uploaded); this just changes the documentation to match.

PR #889, opened 2022-04-30
Stop re-exporting logging details from tide::log

joshtriplett:
Continue using logging internally, but don't re-export log macros or
mechanisms to start logging.


Fishrock123:
looks like a braking change to me, but I otherwise approve

joshtriplett:
git HEAD is currently 0.17, which is a breaking-change release already, so I think it's safe to do this.

Fishrock123 (approved):

Issue #854, opened 2021-10-17
Feature Request: Add `impl Middlware` to support simple stateless middlewares


NfNitLoop:
I read a [blog post] that included a simplified middleware so I wrote one of my own:

```rust
/// Middleware that disables browser caching by default.
async fn no_store<'a, Data>(req: tide::Request<Data>, next: tide::Next<'a, Data>) -> tide::Result<Response>
where Data: Clone + Send + Sync + 'static
{
    use tide::http::cache::{CacheControl, CacheDirective};
    let mut response = next.run(req).await;

    if let None = response.header("Cache-Control") {
        let mut header = CacheControl::new();
        header.push(CacheDirective::NoStore);
        header.push(CacheDirective::MaxAge(Duration::from_secs(0)));

        response.insert_header(header.name(), header.value());
    }
    Ok(response)
}
```

Unfortunately, that didn't work.  Off to [the API docs], and I find that Middleware only has an `impl` for: 

```
impl<State, F> Middleware<State> for F
where
    State: Clone + Send + Sync + 'static,
    F: Send + Sync + 'static + for<'a> Fn(Request<State>, Next<'a, State>) -> Pin<Box<dyn Future<Output = Result> + Send + 'a>>, 
```

At this point, as a merely intermediate Rust user, my eyes glazed over.  So I generated my own impl like this: (Though I had to just try out different lifetimes until the compiler was happy. Because I'm still a bit vague on the how async_trait transforms the code, and so what signature I need to implement to match the one in the Rust docs.)

```rust

struct NoStore {}

#[async_trait]
impl <State: Clone + Send + Sync + 'static> tide::Middleware<State> for NoStore {
    async fn handle<'a, 'b>(&'a self, req: tide::Request<State>, next: tide::Next<'b, State>) -> tide::Result<Response>
    {
        // same body.
    }
}
```

But, AFAICT, there's no reason that Tide couldn't just add an `impl Middleware` for these type of bare/stateless middlewares.  Then you could just:

```rust
app.with(no_store);

// instead of having to use a struct and manually impl the trait.

app.with(NoStore{});
```

[blog post]: https://blog.yoshuawuyts.com/tide/
[the API docs]: https://docs.rs/tide/0.16.0/tide/trait.Middleware.html#impl-Middleware%3CState%3E

nyxtom:
[...]

PR #895, opened 2022-07-15
Remove State and for<'a> as requirements from Middleware and Request

nyxtom:
> Note: this would be a breaking change to the ecosystem as it eliminates the generics on all the traits (`Server<State>`, `Listener<State>`, `Endpoint<State>`, `Request<State>`, `Next<'_>`, `Middleware<State>`) however I think we could agree this is a net gain in ergonomics and allows users to add multiple `.with_state` via a new `StateMiddleware` (placing them in extensions on *both* request and response)

There are a few existing #645 #715 and #643 related to improving ergonomics. As it currently stands the generic `State` is required on `Request<State>`, as well as `Middleware<State>`, `Next<'_, State>`, `Endpoint<State>`, `Listener<State>`, and `Server<State>` and other places. Since we already have the ability to use extensions for things like `req.ext::<User>()` I figured why not just merge the two and call it a day. 

I recognize that moving the *State* object out of the generics and into the Extensions may impact performance so this is something to consider - but I think this improvement in ergonomics and overall simplicity in navigating the code is worth it. The code now treats the state like any other middleware via the `StateMiddleware<S>`.

When you call `app.with_state(state)` this simply wraps the `<T: Clone + Send + Sync + 'static>` in a `StateMiddleware::new(state)`. The *StateMiddleware* will set the request extensions via `set_ext(data.clone())` and also on the response after calling `next.run`.

```rust
use tide::{Request};

/// The shared application state.
#[derive(Clone)]
struct State {
    name: String,
}

// Define a new instance of the state.
let state = State {
    name: "Nori".to_string()
};

// Initialize the application with state.
let mut app = tide::with_state(state);
app.at("/").get(|req: Request| async move {
    Ok(format!("Hello, {}!", &req.state::<State>().name))
});
```

The new `req.state::<State>()` has an `.expects` on it when you call it which will automatically unwrap for you (or error out when you pass it a type that wasn't set in the middleware. The same goes for `res.state::<State>()` which I like because now you can get the application state in the response as well.

The original `.with_state` on the `Server` was there for the generic type, but since state is now in middleware that function has been changed to require passing in the mutable server itself. The above example is the same as writing it below:

```rust
use tide::{Request};

/// The shared application state.
#[derive(Clone)]
struct State {
    name: String,
}

// Define a new instance of the state.
let state = State {
    name: "Nori".to_string()
};

// Initialize the application with state.
let mut app = tide::new();
app.with_state(state);
// app.with_state(state2);
app.at("/").get(|req: Request| async move {
    Ok(format!("Hello, {}!", &req.state::<State>().name))
});
```

The nice thing about this is now you can add multiple different state types rather than force everything into a single struct. You can also directly add the `StateMiddleware` by calling `app.with(StateMiddleware::new(s))` although in practice I think people will stick to `app.with_state`.

Overall I think this is a great gain in ergonomics! 

### Update on Next<'_>

> For this I may separate out into a different pull request but let me know

I've also now refactored the use of Next<'_> to no longer need the extra lifetimes in favor of Arc<T> for endpoints and middleware. Selection routing will return these as well rather than explicit borrows with lifetime requirements. This simplifies some of the code and opens up the possibility for passing along `impl Middleware` function closures.

```rust
// This is an example of a function middleware that uses the
// application state. Because it depends on a specific request state,
// it would likely be closely tied to a specific application
fn user_loader(mut request: Request, next: Next) -> Pin<Box<dyn Future<Output = Result> + Send>> {
    Box::pin(async {
        if let Some(user) = request.state().find_user().await {
            tide::log::trace!("user loaded", {user: user.name});
            request.set_ext(user);
            Ok(next.run(request).await)
        // this middleware only needs to run before the endpoint, so
        // it just passes through the result of Next
        } else {
            // do not run endpoints, we could not find a user
            Ok(Response::new(StatusCode::Unauthorized))
        }
    })
}
```

This is the new format as it no longer requires `<State>` or `Next<'_, State>` or `Next<'_>` or any `user_loader<'a>` as the requirement `for<'a>` is now gone. I will update this pull request to also support the use of a function closure. This will solve https://github.com/http-rs/tide/issues/854 as well through the use of `impl Middleware`

```rust
async fn auth_middleware(request: tide::Request, next: tide::Next) -> tide::Result {
    let authenticated = match request.header("X-Auth") {
        Some(header) => header == "secret_key",
        None => false,
    };

    if authenticated {
        Ok(next.run(request).await)
    } else {
        Ok(tide::Response::new(tide::StatusCode::Unauthorized))
    }
}
```

In the process I've simplified the use of the `Next` so that it can be used as an `Endpoint` (rather than have a separate MiddlewareEndpoint). This was useful for the `Server` to wrap the existing `Next` selection and add in its own middleware.


nyxtom:
The way I do this in my applications is I end up with a simple `request_ext.rs` where I want to add common extensions such as:

```rust
pub trait AuthRequest {
    fn is_authenticated(&self) -> bool;
}

impl<State> AuthRequest for Request<State> {
    fn is_authenticated(&self) -> bool {
        self.session().get::<Claims>("tide.uid").is_some()
    }
}
```

And then just have a prelude in `main.rs`

```rust
mod prelude {
    pub use crate::request_ext::*;
    pub use crate::route_ext::*;
    pub use tide_flash::ext::*;
}
```

The nice this is this would cleanup my extensions as well:

```rust
use tide::{Request, RequestState};

pub trait AuthRequest {
    fn is_authenticated(&self) -> bool;
}

impl AuthRequest for Request {
    fn is_authenticated(&self) -> bool {
        self.session().get::<Claims>("tide.uid").is_some()
    }
}

impl RequestState<State> for Request {
    fn state(&self) -> &State {
        self.ext::<State>().unwrap()    
    }
}
```

nyxtom:
I've also experimented with the use of macros to derive the `RequestState<State>` however I could never quite get auto complete to work at the moment. Looks like it may be related to this issue in [rust-analyzer #12759](https://github.com/rust-lang/rust-analyzer/issues/12759)

```rust
use proc_macro::{self, TokenStream};
use quote::quote;
use syn::{parse_macro_input, DeriveInput};

#[proc_macro_derive(RequestExt)]
pub fn derive_request_ext(input: TokenStream) -> TokenStream {
    let DeriveInput { ident, .. } = parse_macro_input!(input);
    let output = quote! {
        impl tide::RequestState<#ident> for tide::Request {
            fn state(&self) -> &#ident {
                self.ext::<#ident>().unwrap()
            }
        }
    };
    output.into()
}
```

examples/nested.rs
```diff
@@ -1,6 +1,6 @@
 #[async_std::main]
 async fn main() -> Result<(), std::io::Error> {
-    tide::log::start();
+    // tide::log::start();
```
jbr, line 3: Why this change?

nyxtom:
@Fishrock123 would you mind taking a look at this as well? Trying to get this one to move forward for 0.17

PR #898, opened 2022-07-19
Fix github ci clippy tests on nonminimal bool test

nyxtom:
The symmetrical test here fails a nonminimal bool test, but it looks fine to skip in this case


Fishrock123 (approved):

Fishrock123:
@nyxtom Just needs a rebase

PR #899, opened 2022-09-05
Expose the matched route as an extension on the request

kyrias:
I considered adding it directly to the `Request` struct, though now I'm not sure anymore why I didn't, haha.  I guess this PR is more of an RFC though.  I would be happy with either implementation, but I do have a need for getting the route.

---

Edit: I remembered now why I chose not to add it to the `Request` struct which is that it feels cleaner to have the `Request` struct only deal with the actual client-submitted request, and the concept of a matched route is purely a server-side abstraction.

But since the `Request` already contains the list of captured route parameters I guess this point is a bit moot.  Though I guess if this separation was wanted it wouldn't be unreasonable to remove the `route_params` struct member and turn that into an extension struct as well.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -2,7 +2,7 @@ use async_std::io::{self, prelude::*};
 use async_std::task::{Context, Poll};
 use routefinder::Captures;
 
-use std::ops::Index;
+use std::ops::{Deref, Index};
 use std::pin::Pin;
 
 #[cfg(feature = "cookies")]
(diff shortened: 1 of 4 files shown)


Fishrock123:
I think this is good, but it could just be a method on `tide::Request` IMO.

kyrias:
Reasonable.  I've made the struct internal to the crate and added a method to get the extension now.

PR #905, opened 2023-04-18
error msg when i run the example 

mohsen159:
Some errors have detailed explanations: E0432, E0433, E0670, E0752.
For more information about an error, try `rustc --explain E0432`.


joshtriplett:
This seems to be a pull request between two branches of the tide repository. Perhaps you meant to file an issue rather than a PR?

PR #906, opened 2023-04-27
Fix route in static_file example

vijfhoek:
According to the example, the route should be `/src/*` but this returns a 404 when going to e.g. http://localhost:8080/dir/Cargo.toml, and even a panic when going to http://localhost:8080/dir/dir/Cargo.toml:

```
   3: core::option::Option<T>::unwrap
             at /build/rustc-1.68.2-src/library/core/src/option.rs:823:21
   4: <tide::fs::serve_dir::ServeDir as tide::endpoint::Endpoint<State>>::call::{{closure}}
             at /home/sijmen/.cargo/registry/src/github.com-1ecc6299db9ec823/tide-0.16.0/src/fs/serve_dir.rs:30:20
```

Maybe replacing that `unwrap` with something more graceful is a good idea too?

### DIFF
--- a/examples/static_file.rs
+++ b/examples/static_file.rs
@@ -4,7 +4,7 @@ async fn main() -> Result<(), std::io::Error> {
     let mut app = tide::new();
     app.with(tide::log::LogMiddleware::new());
     app.at("/").get(|_| async { Ok("visit /src/*") });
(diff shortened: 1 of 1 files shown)


vijfhoek:
I'm dumb, was looking at the `main` branch while I should've looked at version 0.16.0

